PR #21, opened 2018-11-01
Add depth_cue and fake_tube to simulate tubes with lines

dmreagan:
This PR adds options to increase the rendering realism of `actor.line` by simulating the properties of tubes. These simulated tubes mimic the behavior of `actor.streamtube` by using a geometry shader to replace line segments with camera-aligned rectangles composed of triangle strips. This technique can give many of the benefits of `actor.streamtube` with little of its added computational cost. 

Additionally, this PR paves the way for future developments with custom GLSL shaders.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -13,7 +13,7 @@ from fury.utils import (lines_to_vtk_polydata, set_input, apply_affine,
 
 def slicer(data, affine=None, value_range=None, opacity=1.,
            lookup_colormap=None, interpolation='linear', picking_tol=0.025):
-    """ Cuts 3D scalar or rgb volumes into 2D images
+    """Cut 3D scalar or rgb volumes into 2D images.
 
     Parameters
     ----------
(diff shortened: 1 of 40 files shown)


fury/actor.py
```diff
@@ -516,6 +521,21 @@ def line(lines, colors=None, opacity=1, linewidth=1,
     poly_mapper.SelectColorArray("Colors")
     poly_mapper.Update()
 
+    if depth_cue:
+        geom_shader_file = open("fury/shaders/line.geom", "r")
```
MarcCote: This relative path breaks when the current working folder is not the root of `fury`. See PR https://github.com/dmreagan/fury/pull/1.

skoudoro:
LGTM, I just test it on my mac and it works like a charm. it just needs some tests to make codecov happy and we can merge it.

skoudoro:
Thank you @dmreagan! first shader code in!

PR #29, opened 2018-11-27
Added surface function

PDN96:
The surface function returns a surface_actor.


Garyfallidis:
Okay two quick comments. a) try to avoid using for loops if possible and b) add tests. See examples in the tests folders. Thx!

PR #39, opened 2018-12-10
removing widget module

skoudoro:
UI module replaces the widget module. This module is not used so it is unnecessary to keep it.


Garyfallidis:
Happy to see some great cleanups!

PR #40, opened 2018-12-12
Re-enable transparency test, change colors

dmreagan:
While investigating #6, I found that `test_order_transparent` had been mostly commented out. When I un-commented it, one of the assertions failed. When the green line was in front, the test expected the pixel's green component to be larger than the red component. In reality, it was smaller by 1, despite the image looking correct. I changed the green line's color from [0, 0.5, 0] to [0, 1., 0] and now the tests pass, both in offscreen and onscreen rendering.


skoudoro:
Thank you @dmreagan! LGTM, I just need to test it on my mac and then, I can merge it.

PR #43, opened 2019-01-07
BUG: Fix `peaks_slicer` actor properties not being enforced.

### LINKED ISSUE #22: Peak slicer doesn't honor linewidth parameter
Changing the `linewidth` doesn't change the width of the peak lines. You can change the width manually with 

```python
actor.GetProperty().SetLineWidth(5)
```

jhlegarreta:
The actor created in `peak_slicer` was not having its properties
enforced because the `PeakSlicerActor` instance was not explicitly
setting its `vtk::Property` properties. These properties were being set
in the actor of the `self.line` object, but needed to be set for the
`PeakSlicerActor` class itself.

Fixes #22.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1147,11 +1147,13 @@ def peak_slicer(peaks_dirs, peaks_values=None, mask=None, affine=None,
                                       peaks_dirs[tuple(center)][i] * pv + xyz))
                     list_dirs.append(symm)
 
-            self.mapper = line(list_dirs, colors=colors,
-                               opacity=opacity, linewidth=linewidth,
-                               lod=lod, lod_points=lod_points,
-                               lod_points_size=lod_points_size).GetMapper()
-            self.SetMapper(self.mapper)
+            self.line = line(list_dirs, colors=colors,
+                             opacity=opacity, linewidth=linewidth,
+                             lod=lod, lod_points=lod_points,
+                             lod_points_size=lod_points_size)
+
+            self.SetProperty(self.line.GetProperty())
+            self.SetMapper(self.line.GetMapper())
 
         def display(self, x=None, y=None, z=None):
             if x is None and y is None and z is None:


jhlegarreta:
Same as with #44, Travis errors come from the `bz2file` module not being found.

skoudoro:
Nibabel team plans to fix this and @dmreagan  will review your PR tomorrow so let's wait until Thursday for this PR.

fury/actor.py
```diff
@@ -1147,10 +1147,14 @@ def display_extent(self, x1, x2, y1, y2, z1, z2):
...
-            self.mapper = line(list_dirs, colors=colors,
-                               opacity=opacity, linewidth=linewidth,
-                               lod=lod, lod_points=lod_points,
-                               lod_points_size=lod_points_size).GetMapper()
+            self.line = line(list_dirs, colors=colors,
+                             opacity=opacity, linewidth=linewidth,
+                             lod=lod, lod_points=lod_points,
+                             lod_points_size=lod_points_size)
+
+            self.property = self.line.GetProperty()
```
dmreagan: Just for brevity, can we shorten it to: 

```python
self.SetProperty(self.line.GetProperty())
self.SetMapper(self.line.GetMapper())
```

Otherwise, this looks good to me.

jhlegarreta: +1 : cd005b1 applies the requested change.

jhlegarreta:
@dmreagan May be I should have also modified the test and add some opacity values to less than 1.0 to actually proof that the patch is working?

dmreagan:
More tests, or more explicit tests, are usually a good idea. If you have time, go for it. But since this PR fixes an existing feature rather than adding something new, I'm not going to be strict about it.

jhlegarreta:
@dmreagan looks reasonable. Thanks. Proposed PR #48 to effectively exercise the property.

PR #44, opened 2019-01-07
BUG: Fix elementwise comparison deprecation warning.

### LINKED ISSUE #37: Fix DeprecationWarning
We have to fix this 2 warnings:

```
/Users/skab/Software/fury/fury/window.py:942: DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
  flags[i] = np.any(np.all(im == col, axis=-1))

/Users/skab/Software/fury/fury/window.py:274: DeprecationWarning: Class 'fury.window.add' is deprecated, instead use class 'fury.window.Scene.add'.
  " use class 'fury.window.Scene.add'.", DeprecationWarning)
```

jhlegarreta:
Fix elementwise comparison deprecation warning:
```
/Users/travis/build/fury-gl/fury/fury/window.py:942: DeprecationWarning:
elementwise == comparison failed; this will raise an error in the
future.
```

The warning is thrown when the image and the color arrays differ in the
alpha channel being present/absent.

Fixes #37.

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -939,7 +939,7 @@ def analyze_snapshot(im, bg_color=(0, 0, 0), colors=None,
         flags = [False] * len(colors)
         for (i, col) in enumerate(colors):
             # find if the current color exist in the array
-            flags[i] = np.any(np.all(im == col, axis=-1))
+            flags[i] = np.any(np.any(np.all(im[..., :3] == col[:3], axis=-1)))
 
         report.colors_found = flags
 


fury/window.py
```diff
@@ -936,6 +936,9 @@ class ReportSnapshot(object):
     if colors is not None:
         if isinstance(colors, tuple):
             colors = [colors]
+        # Drop the alpha channel if the image does not have it
+        if np.asarray(colors).shape[1] != im.shape[-1]:
```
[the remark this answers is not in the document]
skoudoro: > I had to use asarray in case colors are passed as a list

Even if it is a list. you can do `len(colors[-1])` or `len(colors[0])` and avoid the `np.asarray`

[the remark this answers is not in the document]
skoudoro: I think you should manage the inverse case so use '>' operator. 

Can you add some check for the input image shape

skoudoro:
Thank you @jhlegarreta for this PR

fury/window.py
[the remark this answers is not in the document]
jhlegarreta: > Even if it is a list. you can do `len(colors[-1])` or `len(colors[0])` and avoid the `np.asarray`

Done in c1ca1e7.

> Can you add some check for the input image shape

Done in c1ca1e7.

fury/window.py
```diff
@@ -936,6 +936,11 @@ class ReportSnapshot(object):
     if colors is not None:
         if isinstance(colors, tuple):
             colors = [colors]
+        # Drop the alpha channel if the image/color objects do not have it
+        if len(colors[-1]) > im.shape[-1]:
+            colors = np.delete(colors, 3, axis=1)
+        if len(colors[-1]) < im.shape[-1]:
+            im = np.delete(im, 3, axis=1)
```
skoudoro: After looking at [line 953](https://github.com/fury-gl/fury/pull/44/files#diff-1205ce6da451e0358e8b1b545a6cf719R953), I just wonder whether it will be better to do the same [line 947](https://github.com/fury-gl/fury/pull/44/files#diff-1205ce6da451e0358e8b1b545a6cf719R947) and remove this block? (`np.any(np.all(im[..., :3] == col[:3], axis=-1))`).
it will avoid you to add tests for all this case. What do you think @jhlegarreta?

jhlegarreta: If this is an error, rather than the intended behavior, whether coming from miniconda or nibabel, we should may be check/report there? If the change in there was not intended, it may affect other projects, including DIPY/FURY themselves in some other parts (?).

In either case, and as for the aspect of adding more tests, I think the only case we are missing is the one the second `if` block addresses. I'd be happy to add a test if you spot some image dataset having RGBA.

If this is the new intended behavior, and as for the question itself, while writing the `if` blocks I too thought that it was too verbose, but could not think of a more elegant/compact way. Now you found it :thumbsup: . So using your proposed change seems reasonable, if, as said, we should not be making changes upstream.

skoudoro: Sorry, it was not clear, I was just saying that Travis is failing because of Nibabel or Miniconda. It is not related to your changes. But like you said, DIPY is not affected so I believe that is Miniconda. I will create a PR to fix this issue tomorrow and you can rebase on that.

Concerning your PR, great that you like the other option 😄

jhlegarreta: :thumbsup: OK, thanks for the explanation. I'll rebase 2761a29 once your PR gets merged then.

skoudoro:
LGTM, I will go ahead and merge it. Travis error is related to Nibabel.  I just created an issue on their repo

PR #47, opened 2019-01-10
BUG: Fix `TensorSlicerActor` actor opacity property not being enforced.

jhlegarreta:
Fix `TensorSlicerActor` actor opacity property not being enforced.

The method `tensor_slicer` returns an instance of a `TensorSlicerActor`
actor, and receives, among others the `opacity` that needs to be applied
to the actor as a parameter. However, the parameter was being passed to
the `_tensor_slicer_mapper` method, which returns a mapper, and thus was
not effectively using the `opacity` property.

Now the `opacity` parameter is no longer passed to the
`_tensor_slicer_mapper` and it is set as a `vtk::Property` to the
`TensorSlicerActor` instance.

The corresponding actor test is modified so that the `opacity` property
is effectively exercised.

### DIFF
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -585,13 +585,16 @@ def test_tensor_slicer(interactive=False):
     scene = window.Scene()
 
     tensor_actor = actor.tensor_slicer(mevals, mevecs, affine=affine,
-                                       sphere=sphere,  scale=.3)
+                                       sphere=sphere,  scale=.3, opacity=0.4)
     I, J, K = mevals.shape[:3]
     scene.add(tensor_actor)
     scene.reset_camera()
     scene.reset_clipping_range()
 
     tensor_actor.display_extent(0, 1, 0, J, 0, K)
+    if interactive:
+        window.show(scene, reset_camera=False)
+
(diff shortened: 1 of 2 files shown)


jhlegarreta:
A separate PR as discussed with @dmreagan in issue #22.

fury/tests/test_actors.py
```diff
@@ -566,7 +566,7 @@ def test_peak_slicer(interactive=False):
 
 @npt.dec.skipif(skip_it or not have_dipy)
 @xvfb_it
-def test_tensor_slicer(interactive=False):
+def test_tensor_slicer(interactive=True):
```
skoudoro: Can you put it back to `False`? The tests are not running because of that

jhlegarreta: Sorry, my bad. I changed it for test purposes and left it behind. 69b395f reverts that change.

jhlegarreta:
Just for info, still the `libbz2` library issues around:
```
    from bz2file import BZ2File
E   ImportError: No module named bz2file
```

skoudoro:
Thank you @jhlegarreta

PR #48, opened 2019-01-10
ENH: Exercise the `PeakSlicerActor` `opacity` explicitly.

jhlegarreta:
Exercise the `PeakSlicerActor` opacity explicitly by setting its value
to one different from the default value.


skoudoro:
Thank you @jhlegarreta, merging

PR #53, opened 2019-02-06
[Fix] resolution limit

### LINKED ISSUE #13: window.record() resolution limit
_From @dmreagan on February 9, 2018 18:43_

## Description
`window.record()` can be used to save a render to a PNG image. If you set the `size` to be higher than your monitor size, the resulting image will be capped to just under your monitor's resolution. For example, if you set `size=(5000, 5000)`, but your monitor's resolution is (1920, 1080), the image width will end up being something like 1916 pixels. 

## Way to reproduce
Run any viz example that saves a render with `window.record()`, but change the `size` to be very high. Tested on Windows and Mac with VTK 7.

_Copied from original issue: nipy/dipy#1420_

skoudoro:
the goal of this PR is to fix #13.  We can now save a screenshot bigger than our screen resolution

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -11,13 +11,14 @@ def load_image(file_name, as_vtktype=False):
     Parameters
     ----------
     file_name: string
-        should be *.png, *.bmp, *.jpeg, *.jpg
-    as_vtktype: bool
+        should be png, bmp, jpeg or jpg files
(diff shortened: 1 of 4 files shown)


Garyfallidis:
Try loading RGB images. And RGBA images.

PR #55, opened 2019-02-21
Add stereo rendering support

dmreagan:


fury/window.py
```diff
@@ -658,16 +685,17 @@ def show(scene, title='FURY', size=(300, 300),
 
     """
 
-    show_manager = ShowManager(scene, title, size,
-                               png_magnify, reset_camera, order_transparent)
+    show_manager = ShowManager(scene, title, size, png_magnify, reset_camera,
+                               order_transparent, stereo)
```
skoudoro: There is a problem here, the argument does not match with the `__init__` function so `stereo` value is going to `interactor_style`.

skoudoro:
perfect!!!  Thanks @dmreagan, merging as soon as Travis finish

PR #62, opened 2019-03-14
Directional color odfs

scott-trinkle:
This adds an option to color each point on each ODF by the direction on the sphere, so similarly oriented peaks have the same color. This is a common, and in my opinion extremely helpful, way to visualize multiple ODFs. 

Similar to how the same functionality is implemented in `peak_slicer`, this option is implemented by setting `colormap=None`. Currently, this throws an error if `colormap=None` and `global_cm=True`, but these are already sort of conflicting commands. 

Here are comparative visualizations, using the CSD example in the DIPY gallery: (http://nipy.org/dipy/examples_built/reconst_csd.html#example-reconst-csd)

`colormap='Plasma'`
csd_odfs

`colormap='None'`
csd_dec_odfs

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -915,7 +915,7 @@ def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
 
     if global_cm:
         if colormap is None:
-            raise IOError("if global_cm=True, colormap must be defined"
+            raise IOError("if global_cm=True, colormap must be defined")
         else:
             cols = create_colormap(all_ms.ravel(), colormap)
     else:
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -914,7 +914,7 @@ def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
     cells.SetCells(ncells, all_faces_vtk)
 
     if global_cm:
-        if colormap == None:
+        if colormap is None:
             raise IOError("if global_cm=True, colormap must be defined"
         else:
             cols = create_colormap(all_ms.ravel(), colormap)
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you for that @scott-trinkle! It would be good if you could add some test to exercise this new feature. 

Furthermore, I think you can manage the case that you are explaining above by raising an IOError like [here](https://github.com/fury-gl/fury/blob/master/fury/colormap.py#L236)

sitek:
> Thank you for that @scott-trinkle! It would be good if you could add some test to exercise this new feature.
> 
> Furthermore, I think you can manage the case that you are explaining above by raising an IOError like [here](https://github.com/fury-gl/fury/blob/master/fury/colormap.py#L236)

I added a test and raise this error for `colormap=None` and `global_cm=True`.

skoudoro:
@Garyfallidis, @StongeEtienne, Can you have a quick look at this PR before merging it? 

Thank you for your feedback

StongeEtienne:
@skoudoro It looks fine and works for me.

However, it might be preferable to "enforce" radial_scale=True (it was the same issue with white coloring), and maybe add a warning "RGB coloring can only be used with radial_scale".
(when colormap=None & radial_scale=False).

Especially if "RGB" (None) it's the default coloring.

skoudoro:
Thank for the review @StongeEtienne. I will go ahead and merge this nice feature.

Thank you @sitek and @scott-trinkle! 

> However, it might be preferable to "enforce" radial_scale=True (it was the same issue with white coloring), and maybe add a warning "RGB coloring can only be used with radial_scale".
(when colormap=None & radial_scale=False).

I will update this point on a new PR.

PR #86, opened 2019-07-16
Fixes a typo in the viz_ui tutorial.

arokem:
Also, adds six as a requirement.

See: https://github.com/fury-gl/fury/blob/master/fury/ui.py#L8


skoudoro:
Thanks for this @arokem. We will keep the compatibility with python 2.7 until the end of the year so good catch for `six` module.

Can you rebase and it will be ready to go?

arokem:
Rebased!

skoudoro:
Thank you @arokem, merging

PR #107, opened 2019-09-12
Increase coverage for test_actors

skoudoro:
This PR just adds some more global tests and fix some pep8.


Garyfallidis:
Good job. Thanks!

PR #110, opened 2019-09-19
Experimental work

guaje:


Garyfallidis:
Suggesting enabling automatic trailing space removal on save in your editor. Will save you precious time for manually removing these trailing spaces.

dmreagan:
I haven't looked closely at your texture experiment, but one thing to look at while you're trying things out is the difference between the GLSL `texture` and `texelFetch` functions. As I understand it, `texture` is used for standard graphics things, but `texelFetch` can be useful if you want to use a texture as a data structure. 

See [this Stack Exchange post](https://gamedev.stackexchange.com/a/66475) and [this VTK mailing list post](http://vtk.1045678.n5.nabble.com/Re-vtk-developers-Passing-arbitrary-float-data-to-a-shader-td5749994.html).

skoudoro:
Thank you for that @guaje!

Great information @dmreagan! In this case, we do not use the texture as a data so it is ok. But we plan to do it soon so we will keep your information preciously 😄. merging!

PR #124, opened 2019-11-21
Texture utilities and actors

Garyfallidis:
``scene.add(actor.texture(data))``

image

``actor.texture_on_sphere(data)``

image


skoudoro:
Thank you for these awesome features !!!! merging

PR #125, opened 2019-11-25
Add Deprecated decorator

skoudoro:
Highly inspired/copied from Nibabel, this is a utility function to prepare the deprecation of many functions/modules. 
Currently, our system is quite manual and not enough rigorous so it will bring more consistency.


skoudoro:
I think I encounter the same issue as https://github.com/pypa/setuptools/issues/1903, this is why python3.5 is failing on windows. Need to find a workaround.

PR #129, opened 2019-12-04
update loading and saving IO for polydata

StongeEtienne:
changed the if case for loading poly data, added vtp (XML vtk), 
changed the coloring default name from None to RGB

MNI obj is now detected if cannot read it as OBJ
to save MNI obj you need to use ".mni.obj" extension (safer for other tools)

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -562,7 +562,7 @@ def streamtube(lines, colors="RGB", opacity=1, linewidth=0.1, tube_sides=9,
     poly_mapper.Update()
 
     # Color Scale with a lookup table
-    if is_colormap:
+    if is_colormap is not None and is_colormap:
         if lookup_colormap is None:
             lookup_colormap = colormap_lookup_table()
         poly_mapper.SetLookupTable(lookup_colormap)
@@ -686,8 +686,7 @@ def line(lines, colors="RGB", opacity=1, linewidth=1,
                                 vtkShaderCallback)
 
     # Color Scale with a lookup table
-    if is_colormap:
-
+    if is_colormap is not None and is_colormap:
         if lookup_colormap is None:
             lookup_colormap = colormap_lookup_table()
 
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -523,7 +523,7 @@ def streamtube(lines, colors="RGB", opacity=1, linewidth=0.1, tube_sides=9,
 
     """
     # Poly data with lines and colors
-    poly_data, is_colormap = lines_to_vtk_polydata(lines, colors)
+    poly_data, color_is_scalar = lines_to_vtk_polydata(lines, colors)
     next_input = poly_data
 
     # Set Normals
@@ -562,7 +562,7 @@ def streamtube(lines, colors="RGB", opacity=1, linewidth=0.1, tube_sides=9,
     poly_mapper.Update()
 
     # Color Scale with a lookup table
-    if is_colormap is not None and is_colormap:
+    if color_is_scalar:
         if lookup_colormap is None:
             lookup_colormap = colormap_lookup_table()
         poly_mapper.SetLookupTable(lookup_colormap)
@@ -656,7 +656,7 @@ def line(lines, colors="RGB", opacity=1, linewidth=1,
     >>> #window.show(scene)
     """
     # Poly data with lines and colors
-    poly_data, is_colormap = lines_to_vtk_polydata(lines, colors)
+    poly_data, color_is_scalar = lines_to_vtk_polydata(lines, colors)
     next_input = poly_data
 
     # use spline interpolation
@@ -686,7 +686,7 @@ def line(lines, colors="RGB", opacity=1, linewidth=1,
                                 vtkShaderCallback)
 
     # Color Scale with a lookup table
-    if is_colormap is not None and is_colormap:
+    if color_is_scalar:
         if lookup_colormap is None:
             lookup_colormap = colormap_lookup_table()
 
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -441,7 +441,7 @@ def contour_from_roi(data, affine=None,
     return skin_actor
 
 
-def streamtube(lines, colors=None, opacity=1, linewidth=0.1, tube_sides=9,
+def streamtube(lines, colors="RGB", opacity=1, linewidth=0.1, tube_sides=9,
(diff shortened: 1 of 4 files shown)


StongeEtienne:
@skoudoro Now using a dictionary (exept for obj)
if you could check the new "RGB" for coloring doesnt screw dipy_horizon

There is change for dipy comming (io/vtk.py) simply to use
`
if have_fury:
    from fury.utils import lines_to_vtk_polydata, get_polydata_lines
    from fury.io import load_polydata, save_polydata
`
to simplify and remove duplicate

skoudoro:
Thank you for this @StongeEtienne! Can you fix CI's issue? I will review your code after that. Thanks

fury/actor.py
```diff
@@ -559,7 +562,7 @@ def streamtube(lines, colors=None, opacity=1, linewidth=0.1, tube_sides=9,
     poly_mapper.Update()
 
     # Color Scale with a lookup table
-    if is_colormap:
+    if is_colormap is not None and is_colormap:
```
skoudoro: I do not understand this change. Can you rename the variable or explain a bit more?

fury/actor.py
```diff
@@ -680,8 +686,7 @@ def vtkShaderCallback(_caller, _event, calldata=None):
                                 vtkShaderCallback)
 
     # Color Scale with a lookup table
-    if is_colormap:
-
+    if is_colormap is not None and is_colormap:
```
skoudoro: same as above

fury/io.py
```diff
@@ -6,6 +6,9 @@
 from fury.utils import set_input
 
 
+SURFACE_FORMAT = [".vtk", ".vtp", ".fib", ".ply", ".stl", ".xml", ".obj"]
```
skoudoro: What is the goal of this variable? I do not see any use

fury/io.py
```diff
@@ -186,35 +187,32 @@ def load_polydata(file_name, is_mni_obj=False):
...
-        except Exception:  # than try load a MNI obj format
+        # Special case, since there is two obj format
+        reader = vtk.vtkOBJReader()
+        reader.SetFileName(file_name)
+        reader.Update()
+        if reader.GetOutput().GetNumberOfCells() == 0:
             reader = vtk.vtkMNIObjectReader()
     else:
-        raise IOError("polydata " + file_extension + " is not suported")
+        raise IOError("." + file_extension + " is not supported by Fury-gl")
```
skoudoro: by "FURY" instead of "fury-gl"

fury/io.py
```diff
@@ -186,35 +187,32 @@ def load_polydata(file_name, is_mni_obj=False):
...
+        reader = poly_reader.get(file_extension)()
     elif file_extension == "obj":
-        try:  # try to read as a normal obj
-            reader = vtk.vtkOBJReader()
-        except Exception:  # than try load a MNI obj format
+        # Special case, since there is two obj format
+        reader = vtk.vtkOBJReader()
+        reader.SetFileName(file_name)
+        reader.Update()
+        if reader.GetOutput().GetNumberOfCells() == 0:
```
skoudoro: nice!

fury/io.py
```diff
@@ -223,32 +221,29 @@ def save_polydata(polydata, file_name, binary=False, color_array_name=None,
...
         else:
-            # vtkObjWriter not available on python
-            # vtk.vtkOBJWriter()
-            raise IOError("OBJ Writer not available. MNI obj is the only"
-                          " available writer so set mni_tag option to True")
+            raise IOError("Wavefront obj requires a scene \n"
+                          " for MNI obj, use '.mni.obj' extension")
     else:
-        raise IOError("Unknown extension ({})".format(file_extension))
+        raise IOError("." + file_extension + " is not supported by Fury-gl")
```
skoudoro: same comment as above: "FURY"

fury/utils.py
```diff
@@ -176,10 +178,19 @@ def lines_to_vtk_polydata(lines, colors=None):
...
+    # Create the poly_data
+    poly_data = vtk.vtkPolyData()
+    poly_data.SetPoints(vtk_points)
+    poly_data.SetLines(vtk_lines)
+
     # Get colors_array (reformat to have colors for each points)
     #           - if/else tested and work in normal simple case
-    if colors is None:  # set automatic rgb colors
+    is_colormap = False
+    if colors is False or colors is None:
```
skoudoro: `if not colors:` instead. it is shorter

fury/io.py
StongeEtienne: I feel this could be usefull later on for Dipy (or other outside package) to know what format is supported,
But for now it's not used, It's more a suggestion questionning.

Maybe there is also a better way to do it, with a function or ...

fury/actor.py
StongeEtienne: done (I always forget that None is a "false" statement)

fury/actor.py
StongeEtienne: renamed to "color_is_scalar" => can use colormap lut

StongeEtienne:
> Overall, looks good to me. I made a few comments below. I just need to test it. I wonder if the parameter "RGB" should be lower case or Uppercase... I prefer lower case but it needs some thinking
I'm never certain either, however I have tendency to simply do  *.lower()* == "rgb"  
so in the end, for this one it doesnt matter, however it depend of what is elsewhere.
I dont have strong oppinion on this either

fury/utils.py
StongeEtienne: I realized this one is necessary, 
since  **colors** can be an array or a list =>
for numpy array it crashes if not using "if colors is False or colors is None:"

For the others "is_colormap", they will be changed since they are only None,False or True

fury/io.py
skoudoro: Indeed, I prefer a function. It will avoid mistakes because we could get all extensions via the dictionary below and avoid duplication errors. I always prefer when we can change things in 1 place only.

Can you fix it or you prefer me to merge this PR and I will create a new PR?

StongeEtienne: I simply removed it for now, maybe will create a new issue / PR when needed

skoudoro: ok, sounds good. Waiting for the CI's

skoudoro:
Thank you @StongeEtienne!  Merging

Issue #81, opened 2019-06-17
Add superquadric function in actor.py


Garyfallidis:
Adding a supequadric function should be very similar to the cone function. See here
https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1453

All needs to change is this line
https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1510

To use this class and of course update the parameters
https://vtk.org/doc/nightly/html/classvtkSuperquadricSource.html
Alternatively you can derive the parameters from the superquadric equation directly.

skoudoro:
New feature added by #131, closing

PR #131, opened 2019-12-06
Add Superquadric primitives and actors

skoudoro:
Adding the superquadrics actor and primitive. Below a small example to use it

```
scene = window.Scene()
centers = np.random.rand(3, 3) * 10
directions = np.random.rand(3, 3)
colors = np.array([[255, 0, 0], [0, 255, 0], [0, 0, 255]])
scale = np.random.rand(5)
roundness = np.array([[1, 1], [1, 2], [2, 1]])

sq_actor = actor.superquadric(centers, roundness=roundness,
                              directions=directions,
                              colors=colors, scale=scale)
scene.add(sq_actor)
window.show(scene)
```

Capture d’écran 2019-12-06 à 06 47 32

### DIFF
--- a/fury/primitive.py
+++ b/fury/primitive.py
@@ -1,6 +1,7 @@
 from os.path import join as pjoin
 import numpy as np
 from fury.data import DATA_DIR
+from fury.transform import cart2sphere
 from scipy.spatial import ConvexHull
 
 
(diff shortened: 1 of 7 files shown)


skoudoro:
should fix #81

fury/primitive.py
```diff
@@ -134,3 +274,47 @@ def prim_sphere(name='symmetric362', gen_faces=False):
...
+    True
+    >>> faces.shape == (720, 3)
+    True
+
+    """
+    def _fexp(x, p):
+        """Return a different kind of exponentiation."""
+        return np.sign(x) * (np.abs(x) ** p)
+
+    sphere_verts, sphere_triangles = prim_sphere('symmetric362')
```
Garyfallidis: other spheres...

skoudoro: updated!

Garyfallidis:
Cool!

PR #132, opened 2019-12-09
Add Billboard actor 

skoudoro:
A simple actor that permits you to draw the shader that you want on one or many Canvas. it uses the billboard technique for always facing the camera. This offers us a powerful tool to draw whatever we want really fast. Below an example of 5 Million funky spheres.

This is close to being done, Maybe today.

Capture d’écran 2019-12-09 à 19 23 52


Garyfallidis:
It's canvas not canva. Please correct everywhere.

Garyfallidis:
Also please correct pep8 issues.

Garyfallidis:
Also in the images you do not have spheres but raging suns ;)

skoudoro:
> It's canvas not canva. Please correct everywhere

All our actors are not in plural, so it is Canva not canvas

> Also please correct pep8 issues.

They are in the experimental folder and all related to glsl code. There is no pep8 otherwise in the main code 

> Also in the images you do not have spheres but raging suns ;)

I know ;-)

PS : This is PR is still in WIP due to a small issue with billboarded scale/direction

guaje:
According to this [dictionary](https://en.wiktionary.org/wiki/canvas), canvas is the singular form.

fury/tests/test_actors.py
```diff
@@ -1208,5 +1208,58 @@ def test_superquadric_actor(interactive=False):
...
+    float df = max(0, dot(direction, normalizedPoint));
+    float sf = pow(df, 24);
+    fragOutput0 = vec4(max(df * color, sf * vec3(1)), 1);
+    """
+
+    canva_actor = actor.canvas(centers,
+                               colors=colors.astype(np.uint8),
+                               scale=scale,
+                               fs_impl=fake_sphere)
+    scene.add(canva_actor)
```
dmreagan: Change to `canvas_actor`

dmreagan: Native English-speaker here. Can confirm "canvas" is singular and "canvases" is plural.

skoudoro: typo fixed!

skoudoro:
Thank you all for the feedback. I addressed all of the issues. this is ready to go. You can have a last look at this PR.

skoudoro:
Renaming: from canvas to billboard actor

guaje:
After running the 2 test cases I found:

 -  If 2 or more billboards are sharing the same space, the fragment shader will color pixels randomly using each one of the billboard's coloring properties (test_square_actor).
 - We need to test how this new actor behaves when more actors are in the scene. There is an overlapping between the billboards and the axes (test_billboard_actor).

|test_square_actor |test_billboard_actor |
|---|---|
|Issue1|Issue2|

I point this here, but I think this must be an issue once this PR gets merged.

guaje:
In the shaders files, there is no clear naming convention. Some variables use the camel case, e.g., centeredVertexMC and others use PEP8, e.g., billboard_size.

dmreagan:
@guaje 

The issue with the squares is a common one, often called [z-fighting](https://en.wikipedia.org/wiki/Z-fighting). It happens anytime you have co-planar polygons.

I don't think I understand the problem with the spheres and the axes.

Regarding shader variable naming style, I suggest camelCase. GLSL uses C-like syntax, so the code should follow C conventions.

guaje:
> The issue with the squares is a common one, often called [z-fighting](https://en.wikipedia.org/wiki/Z-fighting). It happens anytime you have co-planar polygons.

Thanks for the info! I didn't know that.

> Regarding shader variable naming style, I suggest camelCase. GLSL uses C-like syntax, so the code should follow C conventions.

+1

> I don't think I understand the problem with the spheres and the axes.

More than a problem is something we need to be aware of. If you see the following gifs, you will notice how the axes are rendered as an object inside the sphere, but in the billboard example, parts of the axes appear either in front of or behind the billboard.

|Sphere|Billboard|
|---|---|
|Issue1|Issue2|

skoudoro:
Thank you all!  interesting! I did not know the z-fighting too. 

ok for the camelCase, I will update that.

concerning your last example @guaje, it seems there is a different opacity between sphere and billboard. Are you sure about this point?

dmreagan:
Right, the axes appear either in front of or behind because there is no "inside" of a fake sphere. That's also why the billboard spheres quickly pop in front of each other instead of smoothly moving through each other. That's just what you get when you have fake geometry.

skoudoro:
> In the shaders files, there is no clear naming convention. Some variables use the camel case, e.g., centeredVertexMC and others use PEP8, e.g., billboard_size

Done @guaje!

guaje:
> concerning your last example @guaje, it seems there is a different opacity between sphere and billboard. Are you sure about this point?

Yes, I created both examples with different opacities. It was just to illustrate how the traditional sphere interacts with other actors.

> Right, the axes appear either in front of or behind because there is no "inside" of a fake sphere. That's also why the billboard spheres quickly pop in front of each other instead of smoothly moving through each other. That's just what you get when you have fake geometry.

Yes, my point is, we have not explored how these fake geometry actors interact with other actors.

skoudoro:
Any news concerning this PR?

guaje:
@skoudoro 

Everything is working fine, therefore I'm going to merge this PR.

PR #134, opened 2019-12-10
Add event parameter on add_window_callback method in ShowManager class.

guaje:
This parameter will allow different event types. Currently, the only supported event is ModifiedEvent.


skoudoro:
LGTM, Thank you for this @guaje! 

CI's are happy so I will go ahead and merge it. Thanks!

PR #151, opened 2020-02-18
Prim shapes

lidonohu:
Primitive methods for star and rhombicuboctahedron are in primitive.py as well as the winding order algorithm which is separate (not in primitive and not refined using numpy yet).


fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
     vertices = np.ascontiguousarray(xyz)
 
     return vertices, sphere_triangles
+
+def primRhombi():
```
skoudoro: Can you rename it `prim_rhombicubohectadron` ?

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+    '''Return vertices and triangle for rhombicuboctahedron geometry
```
skoudoro: - Can you use double quotes instead of single quotes?
- `triangles` plural

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+
+    :return:
```
skoudoro: Returns instead of :return:

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+                             [2, 3, 14],
+                             [3, 14, 15],
+                             [0, 7, 12],
+                             [7, 12, 19],
+                             [3, 15, 16],
+                             [3, 4, 16],
+                             ], dtype='i8')
+    return my_vertices, my_triangles
+
+def primStar(dim=2):
```
skoudoro: `prim_star` instead of `primStar`

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+    '''Return vertices and triangle for star geometry
```
skoudoro: same comment as above

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+
+
```
skoudoro: You should add `Parameters` title

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+    :param dim: default 2
```
skoudoro: `dim : int`instead

fury/primitive.py
```diff
@@ -328,3 +328,163 @@ def _fexp(x, p):
...
+        triangles = np.array([[1, 9, 0],
+                          [1, 2, 3],
+                          [3, 4, 5],
+                          [5, 6, 7],
+                          [7, 8, 9],
+                          [1, 9, 3],
+                          [3, 7, 9],
+                          [3, 5, 7]], dtype='i8')
+
+    if(dim == 3):
```
skoudoro: `elif` instead of `if`

lidonohu:
Reopening later

PR #152, opened 2020-02-19
Improve checkbox options cases

### LINKED ISSUE #76: Improve Checkbox options access
## Description

Currently, it is quite hard to get access to a specific options since it is a list. 2 strategies to make it simpler:
1. transform `options` attribute (from list to dict)
2. create a method on `CheckBoxUI` to get access to a specific option

subhashsuman01:
fixed #76.
I implemented the suggested changes. I am new to open source so please review my and I would love to work on the feedback I receive.

### DIFF
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -3124,8 +3124,8 @@ class Checkbox(UI):
     ----------
     labels : list(string)
         List of labels of each option.
-    options : dict{label:Option}
-        Dictionary of all the (label,option) pairs in the checkbox set.
+    options : list(Option)
+        List of all the options in the checkbox set.
     padding : float
         Distance between two adjacent options
     """
@@ -3247,8 +3247,8 @@ class RadioButton(Checkbox):
     ----------
     labels : list(string)
         List of labels of each option.
-    options : dict{label:Option}
-        Dictionary of all the (label,option) pairs in the checkbox set.
+    options : list(Option)
+        List of all the options in the checkbox set.
     padding : float
         Distance between two adjacent options
     """
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -3172,9 +3172,9 @@ class Checkbox(UI):
             # Set callback
             option.on_change = self._handle_option_change
 
-    def _get_option(self,label):
+    def _get_option(self, label):
         """ Get the option corresponding to the given label.
-        
+
         Parameters
         ----------
         label : string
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you @subhashsuman01! I will review it as soon as I can.

Can you please add some tests. there are already some tests in `fury/tests/test_ui.py`.

subhashsuman01:
Ok I will add some tests shortly.

fury/ui.py
```diff
@@ -3167,16 +3167,25 @@ def _setup(self):
             line_spacing = option.text.actor.GetTextProperty().GetLineSpacing()
             button_y = button_y + self.font_size * \
                 (label.count('\n') + 1) * (line_spacing + 0.1) + self.padding
-            self.options.append(option)
+            self.options[label] = option
 
             # Set callback
             option.on_change = self._handle_option_change
 
+    def _get_option(self, label):
```
skoudoro: I do not think that this function is necessary. I think you can remove it

fury/ui.py
```diff
@@ -3124,8 +3124,8 @@ class Checkbox(UI):
     ----------
     labels : list(string)
         List of labels of each option.
-    options : list(Option)
-        List of all the options in the checkbox set.
+    options : dict{label:Option}
```
skoudoro: here, we want to keep a list or list of tuple. Then we convert these input to a dictionary.

fury/ui.py
```diff
@@ -3247,8 +3256,8 @@ class RadioButton(Checkbox):
...
```
skoudoro: same comment as above

subhashsuman01:
@skoudoro I don't know why it is failing some tests in for python 3.5. could you please guide me?

fury/ui.py
```diff
@@ -3176,7 +3176,7 @@ def _get_actors(self):
         """ Get the actors composing this UI component.
         """
         actors = []
-        for option in self.options:
+        for option in self.options.values():
```
MarcCote: You can't rely on the ordering of the elements in a Python dictionary. This iteration will depend on the PYTHONSEED and the machine. This means the actors will be added in arbitrary order.

MarcCote:
You can check the error here
https://travis-ci.org/fury-gl/fury/jobs/653760608#L550

The error is probably caused by the arbitrariness of Python's dicts ordering.

skoudoro:
superseded by #218, closing

PR #158, opened 2020-02-23
nose is missing

amrahsmaytas:
ImportError: Need nose >= 1.0.0 for tests - see https://nose.readthedocs.io
can be fixed by adding nose>=1.0.0 to the default.txt ,please accept my commit for smooth testing!
tested on my pc, by adding the above change, showing no error , working fine


skoudoro:
Closing since we do not want to depends on `nose`

Issue #108, opened 2019-09-16
Slider vertical layout


skoudoro:
## Description

Currently, the slider layout is only horizontal. It will be good to allow the vertical slider.

GautamGottipati:
I would like to work on this issue.

skoudoro:
ok @GautamGottipati, you can go ahead and tell us if you have any specific questions

GautamGottipati:
Sure

PR #165, opened 2020-02-25
Vertical Layout for LineSlider2D

Nibba2018:
Referring to #108 ,  my goal is to create a Vertical slider using LineSlider2D.
Changes added so far:
* Added a new parameter to LineSlider2D constructor called `axis` which determines whether the slider is horizontal or vertical.
* `axis=0` implies that the slider is horizontal and `axis=1` implies that it is vertical. (following NumPy's convention of rows and columns).
* Applied changes to the physical properties(**track**) of the slider.
* Modified slider handle to move vertically along the track.


GautamGottipati:
I think this only changes the alignment of the bar, we need to try to fix the movement of disk/square.

Nibba2018:
@GautamGottipati Yep I am working on that only

Nibba2018:
Vertical Slider gif

fury/ui.py
```diff
@@ -1973,11 +2014,21 @@ def format_text(self):
...
-            raise Exception("Disk position outside the slider line")
-        disk_position_x = self.handle.center[0]
-        self._ratio = (disk_position_x - self.left_x_position) / length
+        disk_position_x = None
+        disk_position_y = None
+
+        if self.axis == 0:
+            length = float(self.right_x_position - self.left_x_position)
+            if length != self.track.width:
+                raise Exception("Disk position outside the slider line")
```
skoudoro: `Exception` is too generic, you should pick another one

fury/ui.py
```diff
@@ -1973,11 +2014,21 @@ def format_text(self):
...
+            disk_position_x = self.handle.center[0]
+            self._ratio = (disk_position_x - self.left_x_position) / length
+        else:
+            length = float(self.top_y_position - self.bottom_y_position)
+            if length != self.track.height:
+                raise Exception("Disk position outside the slider line")
```
skoudoro: same as above

fury/ui.py
```diff
@@ -1792,6 +1792,7 @@ def __init__(self, center=(0, 0),
                  length=200, line_width=5,
                  inner_radius=0, outer_radius=10, handle_side=20,
                  font_size=16,
+                 axis=0,
```
skoudoro: Can you replace this variable name, please? use `orientation="horizontal"` instead of `axis=0`.  It is more explicit and not all user are familiar with Numpy convention

fury/ui.py
```diff
@@ -1816,6 +1817,8 @@ def __init__(self, center=(0, 0),
             Side length of the handles (if sqaure).
         font_size : int
             Size of the text to display alongside the slider (pt).
+        axis : int
```
skoudoro: See my comment above. 
```
orientation : str
     horizontal or vertical
```

fury/ui.py
Nibba2018: You mean the Type of Exception right?
for e.g:
`raise HandlePositionBeyondSlider("Exception Text")`

fury/ui.py
Nibba2018: Sure :+1:

fury/ui.py
skoudoro: yes, exactly. Not a custom one but just a standard one like `ValueError`

Nibba2018:
@skoudoro What do these `codecov` tests mean I don't understand...

skoudoro:
- percentage of your code tested for `codecov/patch`. Here, only 46.15% of your code tested, you need to add tests
- impact of your code on the whole project `codecov/project`. here -0.46% so your code reduced the amount of code tested in the whole project... you need to add tests 😄 @Nibba2018

Nibba2018:
@skoudoro Oh wow that's interesting, thanks for the information. I will get the tests done today.

Nibba2018:
@skoudoro I am not sure why the tests are failing :crying_cat_face:

skoudoro:
Wow, I'm impressed by your tests! Good job! UI tests are not obvious and you figure it out, I like it a lot! :+1:

Failing tests are not due to your changes, so no worries,  just some Timeout so I restarted them.

I still need to review your PR, but, finally,  it will be hard before Monday so no worries if you do not news.

Thanks @Nibba2018

Nibba2018:
Thank You so much for appreciating my work @skoudoro :smile: . 

To be honest, the test creation method is actually very smart. :1st_place_medal:

Initially, I was confused about how to create a new test so I kept digging around and eventually I found interesting files within the `data` folder. Then I noticed the `recording=False` argument and thus it was clear to me what I was supposed to do.  :computer_mouse:

fury/ui.py
```diff
@@ -1938,12 +1972,18 @@ def set_position(self, position):
         position : (float, float)
             The absolute position of the disk (x, y).
         """
-        x_position = position[0]
-        x_position = max(x_position, self.left_x_position)
-        x_position = min(x_position, self.right_x_position)
 
         # Move slider disk.
-        self.handle.center = (x_position, self.track.center[1])
+        if self.orientation == "horizontal":
```
skoudoro: same as above

fury/ui.py
```diff
@@ -1973,11 +2014,21 @@ def format_text(self):
...
         # Compute the ratio determined by the position of the slider disk.
-        length = float(self.right_x_position - self.left_x_position)
-        if length != self.track.width:
-            raise Exception("Disk position outside the slider line")
-        disk_position_x = self.handle.center[0]
-        self._ratio = (disk_position_x - self.left_x_position) / length
+        disk_position_x = None
+        disk_position_y = None
+
+        if self.orientation == "horizontal":
```
skoudoro: same as above

fury/ui.py
```diff
@@ -1988,7 +2039,10 @@ def update(self):
         self.text.message = text
 
         # Move the text below the slider's handle.
-        self.text.position = (disk_position_x, self.text.position[1])
+        if self.orientation == "horizontal":
```
skoudoro: same as above

fury/ui.py
```diff
@@ -1826,12 +1829,20 @@ def __init__(self, center=(0, 0),
...
         """
         self.shape = shape
+        self.orientation = orientation
         self.default_color = (1, 1, 1)
         self.active_color = (0, 0, 1)
         super(LineSlider2D, self).__init__()
 
-        self.track.width = length
-        self.track.height = line_width
+        if self.orientation == "horizontal":
```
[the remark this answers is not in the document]
Nibba2018: I did `self.orientation = orientation.lower()`

skoudoro:
good job @Nibba2018! merging

PR #169, opened 2020-02-27
Use primitive for box, cube, square, rectangle actors

### LINKED ISSUE #138: Box, cone etc. to work similarly to superquadric
Currently the superquadric function uses repeat_primitive_function. The rest of the actors need to change too so they are easier to use.
For example box has a single scale and heights parameter which does not allow to make boxes of different sizes with one call. We should be able to have boxes of different sizes with one call.

skoudoro:
The goal of this PR is to replace VTK*Source by our primitive.

This fix #138


fury/actor.py
```diff
@@ -1503,27 +1503,92 @@ def cylinder(centers, directions, colors, radius=0.05, heights=1,
...
+    >>> scene.add(rect_actor)
+    >>> # window.show(scene)
+
+    """
+    return square(centers=centers, directions=directions, colors=colors,
+                  scale=scale)
+
+
+def box(centers, directions=(1, 0, 0), colors=(1, 0, 0), scale=(1, 2, 3)):
+    """Visualize one or many Boxes with different features.
```
Garyfallidis: boxes

Garyfallidis:
Thx!

Issue #166, opened 2020-02-25
Build Native File Dialogs


Garyfallidis:
You can use FileMenu to build a dialogue and load, save files from the disk. One issue is that if a path is too long will not fit the window. Here is an example.

``` python
import numpy as np
from os.path import join as pjoin

from fury import actor, window, ui

scene = window.Scene()
scene.background((1, 1, 1))

open_static = ui.TextBlock2D()
open_static.message = "Select template"

use_dir = ui.TextBlock2D()
use_dir.message = "Use directory"

out_dir = ui.TextBlock2D()
out_dir.message = "Output directory"

from pathlib import Path
home = str(Path.home())

file_menu = ui.FileMenu2D(home, size=(1200, 400))

panel = ui.Panel2D(size=(1800, 420))


class ProcessingObject(object):

    ftemplate = None
    dname = None
    out_dname = None


po = ProcessingObject()


def open_static_callback(obj, event):
    po.ftemplate = pjoin(file_menu.current_directory,
                       file_menu.listbox.selected[0])
    open_static.message +=  ' ' + file_menu.listbox.selected[0]
    showm.render()

    print(po.ftemplate)


open_static.actor.AddObserver('LeftButtonPressEvent',
                              open_static_callback,
                              1.0)

def use_dir_callback(obj, event):
    po.dname = file_menu.current_directory
    use_dir.message += ' selected!'
    showm.render()
    print(po.dname)


use_dir.actor.AddObserver('LeftButtonPressEvent',
                          use_dir_callback,
                          1.0)


def out_dir_callback(obj, event):
    po.out_dname = file_menu.current_directory
    out_dir.message += ' selected!'
    showm.render()
    print(po.out_dname)


out_dir.actor.AddObserver('LeftButtonPressEvent',
                          out_dir_callback,
                          1.0)


showm = window.ShowManager(scene, size=(1980, 1080))

showm.initialize()
panel.add_element(file_menu, coords=(450, 10))
panel.add_element(open_static, coords=(20, 200))
panel.add_element(use_dir, coords=(20, 110))
panel.add_element(out_dir, coords=(20, 10))


scene.add(panel)
scene.add(actor.axes())

showm.render()
showm.start()

```

Nibba2018:
So, our objective is to implement text wrapping for Panel2D, right?

skoudoro:
related to #15

ibrahimAnis:
[...]

PR #173, opened 2020-03-01
Fixing Text Overflow of ListBox2D

Nibba2018:
Referencing #15 and #166 , my objective is to prevent text overflow in ListBox2D. As issue #15 is similar to #166 I am referencing both the issues within this Pull Request.

## Overview:
* Added Fix to prevent text-overflow in ListBox2D by replacing the **end of the string** exceeding the width of the item slot with `...`

## Code Snippet to reproduce UI
```python
import fury.ui as ui
import fury.window as window
values = ['abcdefghijklmnopqrstuvwxyz', 'ahdfjkhkashdfjkhdskfhiweuroqweqr', 'akdiuhdaiufhasdfusdffsdfusdyf', 'Soham']
listbox = ui.ListBox2D(values=values, size=(300, 300))
sm=window.ShowManager(size=(600,600))
sm.scene.add(listbox)
sm.start()
```

## Current Implementation
Previous Implementation

## Fixed Implementation 
Fixed Implementation

Thank You. :heart:

### DIFF
--- a/fury/tests/test_ui.py
+++ b/fury/tests/test_ui.py
@@ -725,7 +725,8 @@ def test_ui_listbox_2d(interactive=False):
     event_counter.check_counts(expected)
 
     # Check if the right values were selected.
-    expected = [[1], [1, 2], [1], ["A Very Very Long Item To Test Text Overflow of List Box 2D"], [1], values]
+    expected = [[1], [1, 2], [1], ["A Very Very Long Item To Test Text Overflow \
+of List Box 2D"], [1], values]
     npt.assert_equal(len(selected_values), len(expected))
     assert_arrays_equal(selected_values, expected)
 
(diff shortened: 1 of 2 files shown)


ibrahimAnis:
Hiii Soham,
I saw your solution.I want to tell you that you are missing one thing.You are losing the actual data.
This solution won't work if you want to create a native file dialog because there we want to the actual path not the wrapped one.

ibrahimAnis:
We want the actual path value not the wrapped path value

Nibba2018:
@ibrahimAnis The actual path is still stored in texblock.element only the textblock.message is modified with the wrapped value

ibrahimAnis:
Yes exactly the actual value is stored in textblock.element. Now suppose you are creating native file dialog and you have used  listBox. Now when you click on any value in list the wrapped value will be selected and their is a function which will search for the selected value but it could not find it because the function will compare it with actual values.

ibrahimAnis:
Bro! can you help me in creating a ui. I want to add a text block and set a callback function there. When the user click on list element it will be shown on text block.I dont know how to add callback function for list

Nibba2018:
@ibrahimAnis kindly show me an example, I might not be able to understand what you are trying to say.

skoudoro:
Hi @Nibba2018, 

Thank you for working on this issue. I just tried your fix and there is still a problem as you can see on the screenshot below. The code used to try your fix is #174 

Furthermore, Can you please add tests. Thank you

image

Nibba2018:
@skoudoro oops sorry, I will get them fixed.

Nibba2018:
@skoudoro let me know if you would like me to add something else.

PR #180, opened 2020-03-04
Native File Dialog Text Overflow Issue

ibrahimAnis:
I am referencing the issue #166  within this Pull Request.

### Overview:
- Added Fix to prevent text-overflow in ListBox2D by replacing the end of the string exceeding the the total no. of chars permissible in slot with '...'

-Now when we replace the overflow text with ... we lose the actual data so to save the actual name of item I used an array named as compress_values.This array will store the actual name of the content.

- I have also defined new functions like get_actual_name(self,compressed_values),set_values(self,values) and compress_values(self, values)

- get_actual_values:- As the name suggests this function will return the actual value of the content which is being replaced by wrapped value.The function will search the actual value in values array.
- compress_value:- This function compress the value by replacing the exceeding chars with ...
- set_values:- This function will assign the values to ListBox's object variable values and this will also call compress_values function to compress the values.

**Code Snippet to reproduce Native File Dialog UI**
```
from os.path import join as pjoin

from fury import ui;
scene = window.Scene()
scene.background((1, 1, 1))

open_static = ui.TextBlock2D()
open_static.message = "Select template"

use_dir = ui.TextBlock2D()
use_dir.message = "Use directory"

out_dir = ui.TextBlock2D()
out_dir.message = "Output directory"

from pathlib import Path
home = str(Path.home())

file_menu = ui.FileMenu2D(home, size=(450, 600))

panel = ui.Panel2D(size=(800,900))


class ProcessingObject(object):

    ftemplate = None
    dname = None
    out_dname = None


po = ProcessingObject()


def open_static_callback(obj, event):
    po.ftemplate = pjoin(file_menu.current_directory,
                       file_menu.listbox.selected[0])
    open_static.message +=  ' ' + file_menu.listbox.selected[0]
    showm.render()

    print(po.ftemplate)


open_static.actor.AddObserver('LeftButtonPressEvent',
                              open_static_callback,
                              1.0)

def use_dir_callback(obj, event):
    po.dname = file_menu.current_directory
    use_dir.message += ' selected!'
    showm.render()
    print(po.dname)


use_dir.actor.AddObserver('LeftButtonPressEvent',
                          use_dir_callback,
                          1.0)


def out_dir_callback(obj, event):
    po.out_dname = file_menu.current_directory
    out_dir.message += ' selected!'
    showm.render()
    print(po.out_dname)


out_dir.actor.AddObserver('LeftButtonPressEvent',
                          out_dir_callback,
                          1.0)


showm = window.ShowManager(scene, size=(1200, 1000))

showm.initialize()
panel.add_element(file_menu, coords=(250,20))
panel.add_element(open_static, coords=(20, 200))
panel.add_element(use_dir, coords=(20, 110))
panel.add_element(out_dir, coords=(20, 10))


scene.add(panel)
#scene.add(actor.axes())

showm.render()
showm.start()
```

The GIF of test is as follows:- 
 
file_dialog

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -22,8 +22,8 @@ coverage:
         # Having this threshold set a little higher (0.1%) than that makes it 
         # a little more tolerant to fluctuations
         target: auto
-        threshold: 0.1%
+        threshold: 100%
     patch:
       default:
         target: auto
-        threshold: 0.1%
+        threshold: 100%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -22,8 +22,8 @@ coverage:
         # Having this threshold set a little higher (0.1%) than that makes it 
         # a little more tolerant to fluctuations
         target: auto
-        threshold: 100%
+        threshold: 0.1%
     patch:
       default:
         target: auto
-        threshold: 100%
+        threshold: 0.1%
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -319,17 +319,12 @@ class UI(object, metaclass=abc.ABCMeta):
 
 class Button2D(UI):
     """A 2D overlay button and is of type vtkTexturedActor2D.
-
     Currently supports::
-
         - Multiple icons.
         - Switching between icons.
-
     """
-
     def __init__(self, icon_fnames, position=(0, 0), size=(30, 30)):
         """Init class instance.
-
         Parameters
         ----------
         icon_fnames : List(string, string)
@@ -338,10 +333,8 @@ class Button2D(UI):
             Absolute coordinates (x, y) of the lower-left corner of the button.
         size : (int, int), optional
             Width and height in pixels of the button.
-
         """
         super(Button2D, self).__init__(position)
-
         self.icon_extents = dict()
         self.icons = self._build_icons(icon_fnames)
         self.icon_names = [icon[0] for icon in self.icons]
@@ -358,32 +351,25 @@ class Button2D(UI):
 
     def _build_icons(self, icon_fnames):
         """Convert file names to vtkImageDataGeometryFilters.
-
         A pre-processing step to prevent re-read of file names during every
         state change.
-
         Parameters
         ----------
         icon_fnames : List(string, string)
             ((iconname, filename), (iconname, filename), ....)
-
         Returns
         -------
         icons : List
             A list of corresponding vtkImageDataGeometryFilters.
-
         """
         icons = []
         for icon_name, icon_fname in icon_fnames:
             icons.append((icon_name, load_image(icon_fname, as_vtktype=True)))
-
         return icons
 
     def _setup(self):
         """Set up this UI component.
-
         Creating the button actor used internally.
-
         """
         # This is highly inspired by
         # https://github.com/Kitware/VTK/blob/c3ec2495b183e3327820e927af7f8f90d34c3474/Interaction/Widgets/vtkBalloonRepresentation.cxx#L47
@@ -391,7 +377,6 @@ class Button2D(UI):
         self.texture_polydata = vtk.vtkPolyData()
         self.texture_points = vtk.vtkPoints()
         self.texture_points.SetNumberOfPoints(4)
-
         polys = vtk.vtkCellArray()
         polys.InsertNextCell(4)
         polys.InsertCellPoint(0)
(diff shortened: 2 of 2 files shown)


.codecov.yml
```diff
@@ -22,8 +22,8 @@ coverage:
         # Having this threshold set a little higher (0.1%) than that makes it 
         # a little more tolerant to fluctuations
         target: auto
-        threshold: 0.1%
+        threshold: 100%
```
skoudoro: Can you explain this change? Why it is needed?

.codecov.yml
```diff
@@ -22,8 +22,8 @@ coverage:
...
     patch:
       default:
         target: auto
-        threshold: 0.1%
+        threshold: 100%
```
skoudoro: same as above, Can you explain?

fury/ui.py
```diff
@@ -319,17 +319,12 @@ def key_press_callback(i_ren, obj, self):
 
 class Button2D(UI):
     """A 2D overlay button and is of type vtkTexturedActor2D.
-
```
skoudoro: This line should not be removed

fury/ui.py
```diff
@@ -319,17 +319,12 @@ def key_press_callback(i_ren, obj, self):
...
     Currently supports::
-
```
skoudoro: same here

fury/ui.py
```diff
@@ -319,17 +319,12 @@ def key_press_callback(i_ren, obj, self):
...
         - Multiple icons.
         - Switching between icons.
-
```
skoudoro: same in all below. Why do you remove all the empty line?

fury/ui.py
```diff
@@ -3707,41 +3649,75 @@ def select(self, item, multiselect=False, range_select=False):
...
+            total_chars = len(str(value))
+            if total_chars > permissible_chars:
+                excess_chars = total_chars - permissible_chars
+                wrapped_value = value[:int(-excess_chars) - 3] + "..."
+                compressed_names.append(wrapped_value)
+            else:
+                compressed_names.append(value)
+        return compressed_names
+
+    def set_values(self, values):
```
skoudoro: Why do we need this extra function? simpler better than complex. Can you explain me, please? Thank you
Can you add some docstring for this function

fury/ui.py
```diff
@@ -3707,41 +3649,75 @@ def select(self, item, multiselect=False, range_select=False):
...
+        self._values = values
+        self.compressed_values = self.compress_values(values)
+
+    def get_actual_value(self, compressed_value):
```
skoudoro: same question here

fury/ui.py
```diff
@@ -3707,41 +3649,75 @@ def select(self, item, multiselect=False, range_select=False):
...
+        """ Retrieving the actual value of element by compressed value.
+         Parameters
```
skoudoro: strange indentation, one space to remove

fury/ui.py
```diff
@@ -3707,41 +3649,75 @@ def select(self, item, multiselect=False, range_select=False):
...
+        ----------
+        compressed_value:the key for the actual value
```
skoudoro: missing space between before and after `:`

fury/ui.py
ibrahimAnis: We need this function because when someone assigns the value directly to ListBox object 'values' then we could not get the compressed_values.The compress_value function is called from the set_values function. Also I keep the values variable private in ListBox object so that no one can assign value directly to this object.The purpose of creating this function is to build an array of compressed_values.

fury/ui.py
ibrahimAnis: When someone clicks on ListBoxItem then the value showing at UI is assigned to item.element.
We need the actual value that's why I created this function.Suppose we have an item named as 
"Guardians Of The Galaxy (720p) Blue Ray" and because of text overflow we could see only 
" Guardians Of The Gal...".Now here when we select the element we will get the later string.
So to get the actual value I made this function.

fury/ui.py
ibrahimAnis: Hiii @skoudoro I have fixed the errors. Can you please check my latest PR? Please let me know if I need to improve something in my code.

PR #183, opened 2020-03-05
Added star2D and 3D, rhombicuboctahedron to tests_primitive

lidonohu:
The test passes with no issues.  If the coordinates are changed in primitive, then the test needs to be changed as well.  (sorry I accidentally added some changes to primitive while I was testing and they came up in the PR).


lidonohu:
needed to put the test with primitive PR

PR #185, opened 2020-03-05
NF: winding order in utils

lidonohu:
- [ ] Testing in test_utils for winding order and its 2 helper definitions
- [x]  Winding order and helpers


lidonohu:
redone in new PR

PR #178, opened 2020-03-03
changed text position

PaulNicolasHunter:
changed text position to - "above the slider" in the horizontal slider

PR #188, opened 2020-03-07
Added dynamic text positioning

PaulNicolasHunter:
I've added dynamic text positioning to the LineSlider2D continuing the reference of PR #178.
There is a new parameter "alignment" which puts the test position to the right/left (vertical) or top/bottom (horizontal).

I've also modified "viz_ui_slider.py" with tutorials respective to horizontal and left orientation and ran tests there.


PaulNicolasHunter:
I'll do tests locally

PR #190, opened 2020-03-08
Add initial state to checkbox

### LINKED ISSUE #75: Improve Checkbox initialisation
## Description

Currently, it is quite hard to initialize checkbox on a specific state. By default, they are unchecked, but we want them already checked sometimes.

ChenCheng0630:
Fixes #75 
- Added additional attributes to CheckBox
- Changed RadioButtons accordingly
- Added functional tests


requirements/test.txt
```diff
@@ -4,5 +4,6 @@ codecov
 coverage
 flake8
 pytest
+nose
```
skoudoro: `Nose` is deprecated and we do not depend on it. Can you remove it? thank you

fury/ui.py
```diff
@@ -3034,10 +3034,12 @@ def __init__(self, label, position=(0, 0), font_size=18):
             the button of the option.
         font_size : int
             Font size of the label.
+        checked : bool
```
skoudoro: `bool, optional` instead. You can update the previous parameter in the docstring too. Thank you

fury/ui.py
```diff
@@ -3146,24 +3152,32 @@ def __init__(self, labels, padding=1, font_size=18,
         position : (float, float)
             Absolute coordinates (x, y) of the lower-left corner of
             the button of the first option.
+        checked_labels: list(string)
```
skoudoro: `list(string), optional` instead

fury/ui.py
ChenCheng0630: "update the previous parameter in the docstring"
Do you mean adding "optional" whenever there are default values?  That means a lot of these attributes need to be changed, 
```python
def __init__(self, label, position=(0, 0), font_size=18, checked=False):
        """
        Parameters
        ----------
        label : str
            Text to be displayed next to the option's button.
        position : (float, float), optional
            Absolute coordinates (x, y) of the lower-left corner of
            the button of the option.
        font_size : int, optional
            Font size of the label.
        checked : bool, optional
            Boolean value indicates the initial state of the option
        """
```

skoudoro: Ok, you are right, we need a specific PR  for this. Just do it for you case

fury/ui.py
```diff
@@ -3146,24 +3152,32 @@ def __init__(self, labels, padding=1, font_size=18,
...
+            List of labels that are checked on setting up.
         """
+
         self.labels = list(reversed(labels))
         self._padding = padding
         self._font_size = font_size
         self.font_family = font_family
+        self.checked_labels = []
+        self.checked_labels.extend(checked_labels)
```
MarcCote: This will fail when `checked_labels` is `None` (i.e. the default value).
I suggest setting the default value to `[]`, then do a soft copy like so
`self.checked_labels = list(checked_labels)`

fury/ui.py
```diff
@@ -3146,24 +3152,32 @@ def __init__(self, labels, padding=1, font_size=18,
...
         self.on_change = lambda checkbox: None
-        self.checked = []
 
     def _setup(self):
         """ Setup this UI component.
         """
         self.options = []
         button_y = self.position[1]
         for label in self.labels:
+            checked = bool(label in self.checked_labels)
```
MarcCote: No need to cast to `bool` the result of the `in` operator is already a boolean value.

fury/ui.py
```diff
@@ -3146,24 +3152,32 @@ def __init__(self, labels, padding=1, font_size=18,
...
+
             option = Option(label=label,
                             font_size=self.font_size,
-                            position=(self.position[0], button_y))
+                            position=(self.position[0], button_y),
+                            checked=checked)
```
MarcCote: You could to `checked=(label in self.checked_labels))` (the inner parenthesis are optional.

fury/ui.py
```diff
@@ -3253,7 +3267,7 @@ class RadioButton(Checkbox):
         Distance between two adjacent options
     """
 
-    def __init__(self, labels, padding=1, font_size=18,
+    def __init__(self, labels, checked_labels, padding=1, font_size=18,
```
MarcCote: Docstring missing for this new parameter.

fury/ui.py
```diff
@@ -3270,17 +3284,22 @@ def __init__(self, labels, padding=1, font_size=18,
             Absolute coordinates (x, y) of the lower-left corner of
             the button of the first option.
         """
+        if len(checked_labels) > 1:
+            raise ValueError("Only one option \
```
MarcCote: I suggest using a temporary variable instead of breaking the text with a `\`. e.g.
```python
if len(checked_labels) > 1:
    msg = "Only one option can be pre-selected for radio buttons."
    raise ValueError(msg)
```
If needed you can break text string like so
```python
msg = ("Only one option can be pre-selected"
       " for radio buttons.")
```

fury/ui.py
```diff
@@ -3146,24 +3152,32 @@ def __init__(self, labels, padding=1, font_size=18,
...
         button_y = self.position[1]
         for label in self.labels:
+            checked = bool(label in self.checked_labels)
+
             option = Option(label=label,
                             font_size=self.font_size,
-                            position=(self.position[0], button_y))
+                            position=(self.position[0], button_y),
+                            checked=checked)
+
```
MarcCote: Instead of adding a `checked` parameter to `Option`, we could simply call `option.select()` here. Is there a good reason I fail to see that would require modifying the internal `Option` class?

skoudoro: I think both solutions are OK.  In one case you need to add an `if` with `option.select()` and with the other case, no `if`, no function call, just an easy parameter.

MarcCote: Fair enough.

skoudoro:
Hi @ChenCheng0630,

By curiosity, Do you plan to update this PR soon or not? I plan to release FURY next week so it will good if your code could be in. 
Thank you for your feedback

ChenCheng0630:
> Hi @ChenCheng0630,
> 
> By curiosity, Do you plan to update this PR soon or not? I plan to release FURY next week so it will good if your code could be in.
> Thank you for your feedback

Sorry I was a bit busy with work stuff and especially under the COVID-19 situation. But dont worry I will make amendments and other contributions throughout this weekend :)

fury/ui.py
ChenCheng0630: Hi @MarcCote, thank you very much for your review. I have changed most of the things according to your suggestion. However, setting default value to [] seems to be quite dangerous. I did some research on google, and found a better to this issue: https://docs.python-guide.org/writing/gotchas/

So I made the change according to the suggestions above. Pls let me know if there are any more issues :)

skoudoro: Interesting link, I did not know that and I need to test it. Thank you for this information.

BTW, instead of

```python
if checked_labels is None:
    self.checked_labels = []
```

you can do it a more pythonic way:

```python
self.checked_labels = checked_labels or []
```

MarcCote: Yes, you are right to be cautious of any mutable objects as an argument's default value. That's why the `list(...)` was important in my example.

On a similar note, @skoudoro what you proposed
> self.checked_labels = checked_labels or []
would modify the original checked_labels list since it is passed as a reference.

fury/ui.py
```diff
@@ -3349,24 +3355,30 @@ def __init__(self, labels, padding=1, font_size=18,
...
             the button of the first option.
+        checked_labels: list(string), optional
+            List of labels that are checked on setting up.
         """
+
         self.labels = list(reversed(labels))
         self._padding = padding
         self._font_size = font_size
         self.font_family = font_family
+        self.checked_labels = checked_labels or []
```
MarcCote: Any in-place changes to `self.checked_labels` will be reflected to the original `checked_labels` list.

fury/ui.py
```diff
@@ -3407,9 +3419,9 @@ def _handle_option_change(self, option):
         option : :class:`Option`
         """
         if option.checked:
-            self.checked.append(option.label)
+            self.checked_labels.append(option.label)
```
MarcCote: (Following up on my previous comment) In-place modification here.

fury/ui.py
skoudoro: Thank you for the feedback @MarcCote.  Indeed, you are right, I will update this on a new PR.

PR #197, opened 2020-03-12
Added Rhombocuboctahedron, 2D and 3D star to primitive

lidonohu:
Will close previous PR.  Had too many commits.

### DIFF
--- a/fury/tests/test_primitive.py
+++ b/fury/tests/test_primitive.py
@@ -7,8 +7,10 @@ def test_vertices_primitives():
     # Tests the default vertices of all the built in primitive shapes.
     l_primitives = [(fp.prim_square, (4, 3)),
                     (fp.prim_box, (8, 3)),
-                    (fp.prim_star, np.shape(fp.prim_star()), -3, 3, -0.666666666667),
-                    (fp.prim_rhombicuboctahedron, np.shape(fp.prim_rhombicuboctahedron()), -4, 4, 0)]
+                    (fp.prim_star, np.shape(fp.prim_star()),
+                     -3, 3, -0.666666666667),
+                    (fp.prim_rhombicuboctahedron,
+                     np.shape(fp.prim_rhombicuboctahedron()), -4, 4, 0)]
 
     for func, shape, min, max, mean in l_primitives:
         vertices, _ = func()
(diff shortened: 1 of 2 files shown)


fury/tests/test_primitive.py
```diff
@@ -1,19 +1,29 @@
 import numpy as np
 import numpy.testing as npt
-import fury.primitive as fp
+import fury.primitive as fpz
```
skoudoro: small mistake here, I suppose this is `fp`

fury/tests/test_primitive.py
```diff
@@ -1,19 +1,29 @@
...
     l_primitives = [(fp.prim_square, (4, 3)),
-                    (fp.prim_box, (8, 3))]
+                    (fp.prim_box, (8, 3)),
+                    (fp.prim_star, np.shape(fp.prim_star()),
+                     -3, 3, -0.666666666667),
+                    (fp.prim_rhombicuboctahedron,
+                     np.shape(fp.prim_rhombicuboctahedron()), -4, 4, 0)]
 
-    for func, shape in l_primitives:
+    for func, shape, min, max, mean in l_primitives:
```
skoudoro: `min, max, mean` are predefined by python so it is better to avoid this variable name. Can you change them? Just put `e_min, e_max, e_mean` the e is for expected

fury/tests/test_primitive.py
```diff
@@ -1,19 +1,29 @@
...
 import numpy.testing as npt
-import fury.primitive as fp
+import fury.primitive as fpz
 
 
 def test_vertices_primitives():
+    # Tests the default vertices of all the built in primitive shapes.
     l_primitives = [(fp.prim_square, (4, 3)),
-                    (fp.prim_box, (8, 3))]
+                    (fp.prim_box, (8, 3)),
```
skoudoro: the test will fail if you do not update this too. so you need to add the `-0.5, 0.5, etc...`

fury/tests/test_primitive.py
```diff
@@ -1,19 +1,29 @@
 import numpy as np
 import numpy.testing as npt
-import fury.primitive as fp
+import fury.primitive as fpz
 
 
 def test_vertices_primitives():
+    # Tests the default vertices of all the built in primitive shapes.
     l_primitives = [(fp.prim_square, (4, 3)),
```
skoudoro: the test will fail if you do not update this too. so you need to add the `-0.5, 0.5, etc...`

PR #200, opened 2020-03-13
updated start function to allow saving

### LINKED ISSUE #113: Antialiasing does not work with snapshot but works with Showmanager
I see this issue in a compiled version of VTK 9.0 (master) in Linux. Let me know if you see the same in other systems.

I suggest we make a save_as or save_fig method directly in ShowManager to do better what snapshot does now.

Saransh0905:
Fixes #113 
Changed `start(self)` to `start(self,fname = None)`  to save figure (using `vtkRenderLargeImage`)
only if fname is provided.

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -496,7 +496,7 @@ class ShowManager(object):
         """Render only once."""
         self.window.Render()
 
-    def start(self,fname = None):
+    def start(self, fname=None):
         """Start interaction."""
         try:
             self.iren.Start()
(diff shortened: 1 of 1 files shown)


Saransh0905:
@skoudoro Please check this whether it is right or not.

skoudoro:
I do not understand what is the goal of this PR @Saransh0905.
I do not understand what is the problem with #113. 
So, I can not review or merge this PR, I need more information.

Saransh0905:
Screenshot from 2020-03-15 03-02-15
The left one is saved by using `window.snapshot`
The right one is saved by `start(fname = 'Anyname')`.
Problem with snapshot seems to be that antialiasing does not work properly, so saved quality is not good.
So I used  `vtkRenderLargeImage` in `start` function to save the scene. 
@skoudoro Correct me if I am wrong

PR #203, opened 2020-03-14
Namanb009 windowtitlefix

namanb009:
The window.SetWindowName() should always be called after the window.Render(), as mentioned in the VTK documentation.

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -489,10 +489,9 @@ class ShowManager(object):
 
     def render(self):
         """Render only once."""
-        self.window.Render()
-        
+        self.window.Render() 
         if self.title=="FURY":
-            self.window.SetWindowName(self.title+" "+fury_version)
+            self.window.SetWindowName(self.title + " " + fury_version)
         else:
             self.window.SetWindowName(self.title)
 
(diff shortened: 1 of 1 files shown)


fury/window.py
```diff
@@ -494,7 +489,12 @@ def initialize(self):
 
     def render(self):
         """Render only once."""
-        self.window.Render()
+        self.window.Render() 
+        if self.title=="FURY":
```
skoudoro: `self.title.upper()` instead

namanb009:
Thank you @skoudoro 
I am definitely more than happy to contribute.

I've made the changes. Pushing them right away.

fury/window.py
```diff
@@ -494,7 +489,12 @@ def initialize(self):
...
+            self.window.SetWindowName(self.title + " " + fury_version)
+        else:
+            self.window.SetWindowName(self.title)
+
```
namanb009: Fixed PEP8 issues

Garyfallidis:
Good catch @namanb009. Merging...

namanb009:
Thank You @Garyfallidis Glad to make my first contribution.

PR #206, opened 2020-03-16
[FIX] avoid in-place replacements

skoudoro:
This is a follow of @MarcCote comments as you can see [here](https://github.com/fury-gl/fury/pull/190#discussion_r393198045).

Thank you for the feedback! Can you merge this PR when CI's are done and if you do not see anything else.

Thank you

### DIFF
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -3345,7 +3345,7 @@ class Checkbox(UI):
         Distance between two adjacent options
     """
 
-    def __init__(self, labels, checked_labels=None, padding=1, font_size=18,
+    def __init__(self, labels, checked_labels=(), padding=1, font_size=18,
                  font_family='Arial', position=(0, 0)):
         """
         Parameters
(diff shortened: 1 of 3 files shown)


MarcCote:
LGTM.

fury/ui.py
```diff
@@ -3349,26 +3350,26 @@ def __init__(self, labels, checked_labels=None, padding=1, font_size=18,
...
-        checked_labels: list(string), optional
-            List of labels that are checked on setting up.
         """
 
         self.labels = list(reversed(labels))
         self._padding = padding
         self._font_size = font_size
         self.font_family = font_family
-        self.checked_labels = checked_labels or []
+        self.checked_labels = list(checked_labels) or []
```
MarcCote: Oops, I spoke too fast. Just realized this `list(checked_labels)` will fail is it is `None` (i.e.  the default).

MarcCote: Better to just do `self.checked_labels = list(checked_labels)` and change the default to `[]` as I suggested. Calling `list` should do a soft copy of the list.

skoudoro: right! Why I did not see it too....  So maybe I should just replace `None` by `()`. What do you think?

MarcCote: That's even better ^.

MarcCote: and `self.checked_labels = list(checked_labels)` right?

MarcCote: No need for the `... or ...`

skoudoro: yes, I will update that now

PR #209, opened 2020-03-20
Namanb009 viz_ui.py does not show render window when run

namanb009:
There is no need for a bool variable (here: interactive) to call start().

Also, interactive was set as False and show_manager.start() is called only when it is set to True.

interactive is redundant here.


PaulNicolasHunter:
Hi @namanb009, the `interactive` variable is not redundant, it's just a flag that lets the person who is going through the tutorial to either enable the interactive mode or not.

namanb009:
@PaulNicolasHunter Thank you, Vivek for your comment.
Well, it just confused me when I was trying to play around FURY.

Since the file is meant for learning purposes, I do think it is redundant and does nothing but confuse beginners.

dmreagan:
I agree that it is confusing to have `interactive` false by default. But if I remember correctly, it is necessary for automated documentation generation. @skoudoro would know for sure.

skoudoro:
Yes, it is necessary for documentation generation. 

When we generate the documentation, we run all the examples. However, `show_manager.start()` is a blocking window so it blocks documentation generation.

Maybe, we need a smarter way to do that. It needs some thinking but currently, We can not accept this PR, it blocks the documentation generation (you can see the error [here](https://travis-ci.org/github/fury-gl/fury/builds/664913354?utm_source=github_status&utm_medium=notification)

Issue #77, opened 2019-06-07
contour_from_label


skoudoro:
## Description

like `contour_from_roi`, add the possibility to create the contours of a labeled array.

subhashsuman01:
Hi,
I would like to work on this issue. Could anyone help me get started?

skoudoro:
Hi @subhashsuman01 ,

Did you try `contour_from_roi` function and do you understand how it works? 

- If No,  I recommend you to start looking at the Tutorial and Examples on FURY website, look at [the contributing page](http://fury.gl/latest/contributing_link.html) and how VTK works.

- If Yes, you should be able to generate surface from a label array.

Hope it helps

subhashsuman01:
Hi @skoudoro ,
I tried out the contour_from_roi function but I don't know how to get 3D numpy array of data points from a labeled array.

grestonian:
Hi @skoudoro, I would like to work on this issue and go through the code and documentation about contour_from_roi function and will get back to you if I face any doubts. Is there any IRC where we can connect?

Nibba2018:
hello @skoudoro , what exactly do you mean by a labeled array? can you give an example?

skoudoro:
I think the tutorial below explains well what is a label array:

https://www.mathworks.com/help/images/label-and-measure-objects-in-a-binary-image.html

Nibba2018:
@skoudoro Thank you the information. Could you provide me an example so that I can test my implementation?

skoudoro:
fixed by #210, closing

PR #210, opened 2020-03-21
Added contour_from_label method

Nibba2018:
# Overview
In reference to issue #77 
* Takes `Labeled Array`, `affine`, `colors` and `opacity` as arguments for individual surfaces.
* Extracts unique _Regions of Interest_ from the labeled array.
* Generates unique **ROI** surfaces and returns them in a vtkAssembly instance in order of their label IDs.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -444,9 +444,8 @@ def contour_from_roi(data, affine=None,
     return skin_actor
 
 
-def contour_from_label(data, affine=None,
-                       color=None, opacity=None):
-    """Generate surface actor from a binary labeled Array.
+def contour_from_label(data, affine=None, color=None):
+    """Generate surface actor from a labeled Array.
 
     The color and opacity of individual surfaces can be customized.
 
@@ -457,13 +456,9 @@ def contour_from_label(data, affine=None,
     affine : array, shape (4, 4)
         Grid to space (usually RAS 1mm) transformation matrix. Default is None.
         If None then the identity matrix is used.
-    color : (N, 3) ndarray
-        RGB values in [0,1]. Default is None.
+    color : (N, 3) or (N, 4) ndarray
+        RGB/RGBA values in [0,1]. Default is None.
         If None then random colors are used.
-    opacity : float or (N,) ndarray
-        Opacity of surface between 0 and 1. Default is None
-        if opacity is numeric, same value is applied to all surfaces.
-        if opacity is None, then 1.0 is used by default for all surfaces.
 
     Returns
     -------
@@ -479,17 +474,16 @@ def contour_from_label(data, affine=None,
 
     unique_roi_surfaces = vtk.vtkAssembly()
 
-    if opacity is None:
-        opacity = np.ones((nb_surfaces, 1)).astype(np.float)
-    elif isinstance(opacity, (float, int)):
-        opacity = np.full((nb_surfaces), opacity)
-    elif opacity.shape != (nb_surfaces,):
-        raise ValueError("Incorrect opacity array shape")
-
     if color is None:
         color = np.random.rand(nb_surfaces, 3)
-    elif color.shape != (nb_surfaces, 3):
+    elif color.shape[0] != nb_surfaces or color.shape[1] < 3:
         raise ValueError("Incorrect color array shape")
+
+    if color.shape[1] == 4:
+        opacity = color[:, -1]
+        color = color[:, :-1]
+    else:
+        opacity = np.ones((nb_surfaces, 1)).astype(np.float)
 
     for roi_id in unique_roi_id:
         roi_data = np.isin(data, roi_id).astype(np.int)
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -445,7 +445,7 @@ def contour_from_roi(data, affine=None,
 
 
 def contour_from_label(data, affine=None,
-                       color=None, opacity=1):
+                       color=None, opacity=None):
     """Generate surface actor from a binary labeled Array.
 
     The color and opacity of the individual surfaces can be customized.
@@ -473,16 +473,25 @@ def contour_from_label(data, affine=None,
 
     unique_roi_id = np.delete(np.unique(data), 0)
 
+    nb_surfaces = len(unique_roi_id)
+
     unique_roi_surfaces = vtk.vtkAssembly()
 
+    if opacity is None:
+        opacity = np.ones((nb_surfaces, 1))
+    elif isinstance(opacity, (float, int)):
+        opacity = np.full((nb_surfaces, 1), opacity).flatten()
+    elif opacity.shape != (nb_surfaces, 1):
+        raise ValueError("Incorrect opacity array shape")
+
     if color is None:
-        color = np.random.rand(len(unique_roi_id), 3)
-    elif color.shape != (len(unique_roi_id), 3):
+        color = np.random.rand(nb_surfaces, 3)
+    elif color.shape != (nb_surfaces, 3):
         raise ValueError("Incorrect color array shape")
 
-    for i, roi_id in enumerate(unique_roi_id):
+    for roi_id in unique_roi_id:
         roi_data = np.isin(data, roi_id).astype(np.int)
-        roi_surface = contour_from_roi(roi_data, affine, color=color[i], opacity=opacity)
+        roi_surface = contour_from_roi(roi_data, affine, color=color[int(roi_id)-1], opacity=opacity[int(roi_id)-1])
         unique_roi_surfaces.AddPart(roi_surface)
 
     return unique_roi_surfaces
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -493,41 +493,41 @@ def contour_from_label(data, affine=None,
     skin_extractor = vtk.vtkContourFilter()
 
     for i, roi_id in enumerate(unique_roi_id):
-        if roi_id != 0:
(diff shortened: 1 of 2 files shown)


fury/actor.py
```diff
@@ -444,6 +444,45 @@ def contour_from_roi(data, affine=None,
...
+    unique_roi_id = np.unique(data)
+
+    for roi_id in unique_roi_id:
+        if roi_id != 0:
+            roi_surface = np.isin(data, roi_id)
+            roi_surface = roi_surface * 1
+            roi_surface = contour_from_roi(roi_surface, affine, color, opacity)
+            unique_roi_surfaces.append(roi_surface)
+
+    return tuple(unique_roi_surfaces)
```
skoudoro: This is a good idea!  unfortunately, we would like to return only 1 actor and not a list of actors. So your solution is not valid.
Look at the code of `contour_from_roi`and I think you should update this part: 

```python
skin_extractor = vtk.vtkContourFilter()
skin_extractor.SetInputData(image_resliced.GetOutput())

skin_extractor.SetValue(0, 1)
skin_extractor.SetValue(1, `your_label_value`)
skin_extractor.SetValue(2, `your_label_value_2`)
.....
```

Nibba2018: for the first argument of `SetValue()` I pass the ID of the contour, but what do I pass for the second argument? Do I need to take another parameter to store the labels?

Also, the vtk [documentation](https://vtk.org/doc/nightly/html/classvtkContourFilter.html#aa33aa2585c805a22bfa9b8454d69f9a1) says that the second argument should be a value.

Nibba2018:
@skoudoro kindly review my newly added changes

fury/actor.py
```diff
@@ -444,6 +444,109 @@ def contour_from_roi(data, affine=None,
...
+        affine[2][0], affine[2][1], affine[2][2], affine[2][3],
+        affine[3][0], affine[3][1], affine[3][2], affine[3][3]))
+    transform.SetMatrix(transform_matrix)
+    transform.Inverse()
+
+    skin_extractor = vtk.vtkContourFilter()
+
+    for i, roi_id in enumerate(unique_roi_id):
+        if roi_id != 0:
+            roi_data = np.isin(data, roi_id)
```
skoudoro: `roi_data = np.isin(data, roi_id).astype(np.int)`

fury/actor.py
```diff
@@ -444,6 +444,109 @@ def contour_from_roi(data, affine=None,
...
+    skin_mapper = vtk.vtkPolyDataMapper()
+    skin_mapper.SetInputConnection(skin_normals.GetOutputPort())
+    skin_mapper.ScalarVisibilityOff()
+
+    skin_actor = vtk.vtkActor()
+
+    skin_actor.SetMapper(skin_mapper)
+    skin_actor.GetProperty().SetOpacity(opacity)
+
+    skin_actor.GetProperty().SetColor(color[0], color[1], color[2])
```
skoudoro: Does this work? each label have different colors normally.

Nibba2018: You are correct, should I ask the pass a parameter or should I generate different colours on my own?

skoudoro: Both. If parameter is none generate one

Nibba2018: I think we need to maintain a lookup table for the different color values.

Nibba2018: @skoudoro I found a better way to implement this!!

Nibba2018:
@skoudoro Here's an implementation example

## Example
contour-labels-2020-03-27-010505

## Code Snippet for the above example
```python
import numpy as np

scene = window.Scene()
data = np.zeros((50, 50, 50))
data[15:20, 25, 25] = 1.
data[25, 20:30, 25] = 2.
data[25, 40:50, 30:50] = 3.

surface = actor.contour_from_label(data, color=np.random.rand(3, 3), opacity=0.6)

scene.add(surface)
scene.reset_camera()
scene.reset_clipping_range()
window.show(scene)
```

Nibba2018:
Next, I will be adding Tests for the same.

Nibba2018:
@skoudoro What kind of tests would you like me to perform for this method. Dipy examples were used for  `contour_from_roi`. Do you want something similar?

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
     return skin_actor
 
 
+def contour_from_label(data, affine=None,
+                       color=None, opacity=None):
```
skoudoro: Can you remove `opacity`? it is not needed because `color` should be RGB or RGBA (manage the opacity)

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
...
+    """Generate surface actor from a binary labeled Array.
```
skoudoro: word `binary` to remove

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
...
+    The color and opacity of individual surfaces can be customized.
+
+    Parameters
+    ----------
+    data : array, shape (X, Y, Z)
+        A labeled array file that will be binarized and displayed.
+    affine : array, shape (4, 4)
+        Grid to space (usually RAS 1mm) transformation matrix. Default is None.
+        If None then the identity matrix is used.
+    color : (N, 3) ndarray
```
skoudoro: (N, 3) or (N, 4)

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
...
+        RGB values in [0,1]. Default is None.
```
skoudoro: RGB/ RGBA

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
...
+        If None then random colors are used.
+    opacity : float or (N,) ndarray
```
skoudoro: to remove

fury/actor.py
```diff
@@ -444,6 +444,63 @@ def contour_from_roi(data, affine=None,
...
+    nb_surfaces = len(unique_roi_id)
+
+    unique_roi_surfaces = vtk.vtkAssembly()
+
+    if opacity is None:
+        opacity = np.ones((nb_surfaces, 1)).astype(np.float)
+    elif isinstance(opacity, (float, int)):
+        opacity = np.full((nb_surfaces), opacity)
+    elif opacity.shape != (nb_surfaces,):
+        raise ValueError("Incorrect opacity array shape")
```
skoudoro: this block is not needed normally

fury/tests/test_actors.py
```diff
@@ -310,6 +310,66 @@ def test_contour_from_roi():
...
+    # window.show(scene2)
+
+    arr = window.snapshot(scene, 'test_surface.png', offscreen=True)
+    arr2 = window.snapshot(scene2, 'test_surface2.png', offscreen=True)
+
+    report = window.analyze_snapshot(arr, find_objects=True)
+    report2 = window.analyze_snapshot(arr2, find_objects=True)
+
+    npt.assert_equal(report.objects, 3)
+    npt.assert_equal(report2.objects, 3)
```
skoudoro: Nice unittests!

fury/actor.py
Nibba2018: Hello @skoudoro , so basically we want to take 3 input arguments and calculate the opacity value internally via the color array?

fury/tests/test_actors.py
Nibba2018: @skoudoro Thank YOU!!

fury/actor.py
Nibba2018: Actually I kept this block in case the user wanted to have the same opacity for all the contour surfaces. So instead of passing an array of opacity values, one can simply pass a number as opacity and it would be converted to an array internally.

fury/actor.py
Nibba2018: I am actually a bit confused with what to do with the opacity parameter.

Nibba2018: I have implemented a solution hopefully it fulfils the requirement.

skoudoro:
Can you just push an empty commit to restart all CI's.

Thank you

Nibba2018:
@skoudoro Glad to help. :heart: 
Sure !

Nibba2018:
@skoudoro It still failed :(

fury/tests/test_actors.py
```diff
@@ -310,6 +310,59 @@ def test_contour_from_roi():
...
+         [1, 1, 0]])
+
+    surface2 = actor.contour_from_label(data2, color=color2)
+
+    scene2.add(surface2)
+    scene2.reset_camera()
+    scene2.reset_clipping_range()
+    # window.show(scene2)
+
+    arr = window.snapshot(scene, 'test_surface.png', offscreen=True)
```
skoudoro: since you are using alpha channel (opacity). try to add this option here: `order_transparent=True`. Be careful with pep8 when adding this option. I hope it will solve the masOS issue.

Nibba2018: Did the required changes.

Nibba2018:
@skoudoro codecov/patch seems to fail. What other tests could be added?

skoudoro:
check the diff tab on Codecov/patch link :smile: You will see what is red

Nibba2018:
@skoudoro Right, that was very stupid of me.

skoudoro:
Thank you @Nibba2018!

Issue #133, opened 2019-12-10
window.record always crashes the desktop environment


lemairecarl:
## Description
When trying to render to an image using `window.record`, the desktop environment ("cinnamon" or other) crashes and needs to be restarted.

## Way to reproduce

Output of `get_info()`:
```
{'fury_version': '0.4.0', 'pkg_path': '/home/carl/venv/lib/python3.6/site-packages/fury', 'commit_hash': 'd9bff9ebbb15fee7773845d0b0866870857b17bb', 'sys_version': '3.6.7 (default, Oct 22 2018, 11:32:17) \n[GCC 8.2.0]', 'sys_executable': '/home/carl/venv/bin/python', 'sys_platform': 'linux', 'numpy_version': '1.16.4', 'scipy_version': '1.3.0', 'vtk_version': '8.1.2', 'matplotlib_version': '3.1.0', 'dipy_version': '1.0.0'}
```

Snippet for reproducing:
```python
from fury import window
scene = window.Scene()
window.record(scene, out_path='test.png')
```

lemairecarl:
I just found out about `window.snapshot`... it works for my use case. However, `record` would be helpful, since I'd like to produce 2 images with different angles.

skoudoro:
Thank you for reporting @lemairecarl. 

I will try to reproduce but I think we need to find a workout for this point. It looks like the same problem I got with Window CI's so it is related to the `offscreen`. Not an easy fix but we will try our best 😄

skoudoro:
> I just found out about window.snapshot... it works for my use case. However, record would be helpful, since I'd like to produce 2 images with different angles.

ok, great! and yes I will still look into it

lemairecarl:
Seems to me that knowing that one function works but not the other will give you important hints (i.e. what's so different between the two?)
Thanks!

namanb009:
I am also taking a look at the issue. window.record() is clearly not working.

ckolluru:
I'm having a similar issue sometimes, when I run either window.record() or window.show().

I'm debugging python code in Visual Studio Code. The functions work well for a while, but they suddenly result in a crash. When it crashes, I see this line in the command window.

Backend QtAgg is interactive backend. Turning interactive mode on.

Does someone have an idea how I might be able to fix this?

irenepersoglia:
I'm having the same issue, tried also with window.snapshot but it won't work either. Any hints on how to resolve this?

kvttt:
[...]

PR #212, opened 2020-03-22
Namanb009 issue 133 fix

namanb009:
This PR is a fix for issue #133, window.record() not  working.

I have updated the code for window.record() based on the VTK documentation. 
It now works as expected.


skoudoro:
The previous approach was working too and the problem is somewhere else. Your approach is not valid to fix #133. Furthermore, The `Writer` object is already on the `save` function.

Closing

PR #215, opened 2020-03-24
Remove more than one actors at once

namanb009:
With this feature, users will now be able to remove more than one actors from the scene in one command


namanb009:
Happy to make a contribution @skoudoro

PR #216, opened 2020-03-24
Utiltiy function to use Hexadecimal color code

namanb009:
This function will allow users to use hexadecimal color throughout FURY instead of rgb.

Type: Enhancement
Improves: User usability.

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -964,3 +964,14 @@ def check_order(vert, triarr):
         else:
             correct_vert[nb] = i
     return correct_vert
+
+def hex_color(color):
+    if color[0] == "#":
+        color[1:]
+
+    r = int("0x" + color[0: 2], 0) / 256
+    g = int("0x" + color[2: 4], 0) / 256
+    b = int("0x" + color[4: 6], 0) / 256
+
+    return((r, g, b))
+    
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -1,0 +1,966 @@
+import sys
+import numpy as np
+import vtk
+from vtk.util import numpy_support
+from scipy.ndimage import map_coordinates
+from fury.colormap import line_colors
+
+
+def set_input(vtk_object, inp):
+    """Set Generic input function which takes into account VTK 5 or 6.
+
+    Parameters
+    ----------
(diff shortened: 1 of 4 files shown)


fury/utils.py
```diff
@@ -964,3 +964,14 @@ def check_order(vert, triarr):
         else:
             correct_vert[nb] = i
     return correct_vert
+
+def hex_color(color):
```
skoudoro: Can you rename this function `hex2rgb` or `hex_to_rgb`?
Can you add docstring (documentation for this function? Look at the functions above

fury/utils.py
```diff
@@ -964,3 +964,14 @@ def check_order(vert, triarr):
...
+    if color[0] == "#":
+        color = color[1:]
+
+    r = int("0x" + color[0: 2], 0) / 256
+    g = int("0x" + color[2: 4], 0) / 256
+    b = int("0x" + color[4: 6], 0) / 256
+
+    return((r, g, b))
```
skoudoro: `return np.array([r, g, b])` instead. we prefer numpy array instead of tuple

fury/utils.py
```diff
@@ -964,3 +964,14 @@ def check_order(vert, triarr):
         else:
             correct_vert[nb] = i
     return correct_vert
+
+def hex_color(color):
+    if color[0] == "#":
+        color = color[1:]
+
+    r = int("0x" + color[0: 2], 0) / 256
```
skoudoro: replace `/` by  `//` to avoid a float value. 

Same comment below.

namanb009: r, g, and b are supposed to have a float value between 0 and 1. Hence, I have used / operator

namanb009:
@skoudoro The utils.py file in this branch just went missing, I don't know-how. So I added it again on my own and now I'm stuck
Can you help me out here?

skoudoro:
Hi @namanb009, 

it looks like you renamed `utils.py` to `colormap.py`. What you should do is:

- go to the folder where there is `setup.py`
- git remove `colormap.py`
- cd fury
- git checkout `utils.py `
- git push

and you should be ok

PR #218, opened 2020-03-29
Changed options attribute from list to dict and updated respective tests

devmodi154:
This is the new PR, a fix for issue #76 as per discussion.

### DIFF
--- a/fury/tests/test_ui.py
+++ b/fury/tests/test_ui.py
@@ -736,13 +736,13 @@ def test_ui_checkbox_default(interactive=False):
                                 position=(10, 10), checked_labels=[])
 
     old_positions = []
-    for option in checkbox_test.options:
+    for option in checkbox_test.options.values():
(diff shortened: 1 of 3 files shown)


skoudoro:
Thank you for the update! I will get back to you as soon as I test it

skoudoro:
supersede and close #152

devmodi154:
@skoudoro I think I can add `get_option` function as mentioned issue before merging this if could help me with a doubt about it

skoudoro:
Sorry, I merged before answering your question.  

I do not think we need a `get_option`, it will be more confusing. The class has already a public attribute `option`.  the get/set method would be useful if our attribute was private/protected (meaning `_option` or  `__option`).

skoudoro:
So, it was a mistake from me on the issue. I had another idea initially but this solution is ok.
thank you @devmodi154

Issue #219, opened 2020-03-30
Issue occur when I Start testing the project


csharp489:
When I start running the test, it gives me these errors .Please help me
code: 
`pytest -svv fury`
Output: 
```python
PS C:\Users\rishabh\FURY> pytest -svv fury                                                                                                                                          =============================================================================== test session starts ===============================================================================
platform win32 -- Python 3.6.2, pytest-5.3.2, py-1.8.1, pluggy-0.13.1 -- c:\python36\python.exe
cachedir: .pytest_cache
rootdir: C:\Users\rishabh\FURY
plugins: cov-2.8.1, forked-1.0.2, localserver-0.5.0, xdist-1.31.0
collected 29 items / 6 errors / 23 selected

===================================================================================== ERRORS ======================================================================================
___________________________________________________________________ ERROR collecting fury/tests/test_actors.py ____________________________________________________________________
ImportError while importing test module 'C:\Users\rishabh\FURY\fury\tests\test_actors.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
c:\python36\lib\site-packages\_pytest\python.py:492: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
c:\python36\lib\site-packages\py\_path\local.py:701: in pyimport
    __import__(modname)
c:\python36\lib\site-packages\_pytest\assertion\rewrite.py:143: in exec_module
    exec(co, module.__dict__)
fury\tests\test_actors.py:11: in <module>
    from fury import actor, window
fury\actor.py:16: in <module>
    import fury.primitive as fp
fury\primitive.py:6: in <module>
    from scipy.spatial import transform
E   ImportError: cannot import name 'transform'
_________________________________________________________________ ERROR collecting fury/tests/test_interactor.py __________________________________________________________________
ImportError while importing test module 'C:\Users\rishabh\FURY\fury\tests\test_interactor.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
c:\python36\lib\site-packages\_pytest\python.py:492: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
c:\python36\lib\site-packages\py\_path\local.py:701: in pyimport
    __import__(modname)
c:\python36\lib\site-packages\_pytest\assertion\rewrite.py:143: in exec_module
    exec(co, module.__dict__)
fury\tests\test_interactor.py:7: in <module>
    from fury import actor, window, interactor
[...]

PR #220, opened 2020-03-30
bumping scipy version to 1.2.0

Nibba2018:
Referencing #219 , bumping minimum scipy version requirement from `0.9` to `1.2.0` for fixing `scipy.spatial.transform` module not found error.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,5 +1,5 @@
 # List required packages in this file, one per line.
 numpy>=1.7.1
 vtk>=8.1.2
-scipy>=0.9
+scipy>=1.2.0
 pillow>=5.4.1


skoudoro:
Thank you @Nibba2018! merging

Nibba2018:
Glad to help 😊

csharp489:
I want to know that why is @Nibba2018  updating scipy version to 1.2.0 rather than 1.4.1.
scipyversion
Is their any specific answer for this @skoudoro

skoudoro:
All software wants to be compatible with most versions. if you put only the last version, you reduce your amount of users. For example, many companies prefer an old stable version than the new/latest one and you do not want to lose them.

skoudoro:
I hope it answers your question @csharp489

Nibba2018:
@csharp489 as @skoudoro mentioned that the `scipy.spatial.transform` module was defined in `1.2.0`, thus setting the minimum scipy version to `1.2.0` will be enough as the later versions will already be having it.

> ok, I just checked, `transform` module is available from Scipy 1.2. However, you have Scipy 1.1.0. So :
> 
>     * you need to update your Scipy version
> 
>     * we need to create a PR and update Scipy minimum version.
> 
> 
> Thank you for the feedback.

PR #222, opened 2020-03-31
Attempt to refactor scrolling in FileMenu2D

devmodi154:
PR against the issue viz.ui.FileMenu2D #16. To improve scrolling in `FileMenu2D` that was already done in `ListBox2D` component.


skoudoro:
Thank you for doing this @devmodi154,

Can you have a look at this PR @MarcCote? Thank you

fury/ui.py
```diff
@@ -3817,6 +3839,105 @@ def scroll_drag_callback(self, i_ren, _obj, _rect_obj):
...
+                        file_names.append(file)
+        file_names.sort(key=lambda s: s.lower())
+        return file_names
+
+    def set_slot_colors(self):
+        """ Sets the text color of the slots based on the type of element
+        they show. Blue for directories and green for files.
+        """
+        for idx, slot in enumerate(self.slots):
+            # print("hello")
```
MarcCote: Debugging print to be removed.

fury/ui.py
```diff
@@ -3529,7 +3530,8 @@ def __init__(self, values, position=(0, 0), size=(100, 300),
                  unselected_color=(0.6, 0.6, 0.6),
                  scroll_bar_active_color=(0.6, 0.2, 0.2),
                  scroll_bar_inactive_color=(0.9, 0.0, 0.0),
-                 background_opacity=1.):
+                 background_opacity=1., directory_path=None,
```
MarcCote: I don't understand why `ListBox2D` needs a `directory_path` and `extensions` variables. That should go/stay in `FileMenu2D` or, if needed, in a subclass of `ListBox2D` to have a specialized list, e.g. `FileListBox2D` or something.

devmodi154: Hey @MarcCote! Thanks for reviewing! 
I understood your point in this comment, but your previous [comment](https://github.com/fury-gl/fury/pull/222#pullrequestreview-384797475) is not so clear. Can you please explain a bit more?

MarcCote: I wrote the comment you are referring to while trying to remember what was the issue raised by #16 (possibly outdated). Here, I think I found a better alternative that leverages `ListBox2D.on_change` callback to change directory.
https://github.com/devmodi154/fury/pull/1

Caveat: it's not dealing with multi-selection (I just take/look at the first selected item), if it is a directory, then move to it. We probably need a double-click event instead (not implemented yet - a similar idea as for `UI.on_left_mouse_button_dragged` might be doable).

MarcCote: @skoudoro we might need a double-click callback ^

skoudoro: `might` :smile: I agree, I already had the need for this feature.  I just created an issue as a reminder (#226)

devmodi154: Thanks for explaining @marscos! The scrolling refactor issue has been resolved but I had a doubt, how about we make ScrollBar a separate component? 
As scrolling can be needed for many other components. Example: A text box has a large text.

MarcCote: Good catch. I like the idea of having a ScrollBar component. :)

devmodi154:
@MarcCote I have tried to separate the ScrollBar component. I wanted to know two things :
- Is this a good direction for the idea of ScrollBar class.
- Would we want the functionality of scrolling using mouse events inside the ScrollBar class or should it belong to list itself?

fury/ui.py
devmodi154: @MarcCote I have tried to separate the ScrollBar component. I wanted to know two things :
- Is this a good direction for the idea of ScrollBar class.
- Would we want the functionality of scrolling using mouse events inside the ScrollBar class or should it belong to list itself?

MarcCote: It could be a good thing, yes. I see you have made an attempt at it. I'll look at it this week.

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
             option_.deselect()
 
         option.select()
         self.checked_labels = [option.label]
         self.on_change(self)
 
 
+class ScrollBar(UI):
+
+    def __init__(self, ListBox2D):
```
skoudoro, line 3518: It can be a panel and `ListBox2D` is a panel improved, so replace `ListBox2D` by `panel`

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+        self.listbox = ListBox2D
```
skoudoro, line 3519: same comment as above

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+        self.reverse_scrolling = self.listbox.reverse_scrolling
```
skoudoro, line 3520: should be an option not dependent on listbox

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+        if not denom:
+            denom += 1
+        super(ScrollBar, self).__init__()
+        self.scroll_step_size = (self.listbox.slot_height *
+                                 self.listbox.nb_slots -
+                                 self.scroll_bar.height) / denom
+        self.scroll_bar_active_color = self.listbox.scroll_bar_active_color
+        self.scroll_bar_inactive_color = self.listbox.scroll_bar_inactive_color
+        self.scroll_bar.color = self.listbox.scroll_bar_inactive_color
+        self.scroll_bar.opacity = self.listbox.background_opacity
```
skoudoro, line 3532: same comment as above. All this variable should not be linked to Listbox.

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+        self.reverse_scrolling = self.listbox.reverse_scrolling
+
+        denom = len(self.listbox.values) - self.listbox.nb_slots
+        if not denom:
+            denom += 1
+        super(ScrollBar, self).__init__()
+        self.scroll_step_size = (self.listbox.slot_height *
+                                 self.listbox.nb_slots -
+                                 self.scroll_bar.height) / denom
+        self.scroll_bar_active_color = self.listbox.scroll_bar_active_color
```
skoudoro, line 3529: `self.scroll_bar_active_color` is really repetitive. We already know that we are on the scrollbar class. Can you change the variable name? Just call it `active_color`.

Same comment for all variables above

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+        self.scroll_bar_inactive_color = self.listbox.scroll_bar_inactive_color
+        self.scroll_bar.color = self.listbox.scroll_bar_inactive_color
+        self.scroll_bar.opacity = self.listbox.background_opacity
+        self.scroll_init_position = 0
+
+    def _setup(self):
+        scroll_bar_height = self.listbox.nb_slots * (
+            self.listbox.panel_size[1] - 2 * self.listbox.margin) \
+            / len(self.listbox.values)
+        self.scroll_bar = Rectangle2D(size=(int(self.listbox.panel_size[0]/20),
```
skoudoro, line 3539: call it just `bar`

fury/ui.py
```diff
@@ -3504,14 +3505,174 @@ def __init__(self, labels, checked_labels, padding=1, font_size=18,
...
+                                      scroll_bar_height))
+        if len(self.listbox.values) <= self.listbox.nb_slots:
+            self.scroll_bar.set_visibility(False)
+
+        self.listbox.panel.add_element(
```
skoudoro, line 3544: do not use `listbox`. You should find a way to connect it differently

ganimtron-10:
Closing with reference to #816.

PR #223, opened 2020-03-31
[Fix] sphere winding issue 

### LINKED ISSUE #137: Issues with provided spheres
Hello,

Using ``primitive.prim_sphere(name='repulsion724', gen_faces=False)`` does not provide correct information to then calculate normals. Although symmetric724 does.
Furthermore, ``gen_faces=True`` shifts the sphere away from its center.
Similarly, most provided spheres have the same problem.

Thank you!

skoudoro:
This PR fix #137 concerning the winding issue and the displacement. All sphere normals are correct now

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -887,9 +887,9 @@ def triangle_order(vertices, faces):
 
     Parameters
     ----------
-    vertices : np.ndarray
+    vertices : ndarray
         array of vertices making up a shape
-    faces : np.ndarray
+    faces : ndarray
         array of triangles
 
     Returns
@@ -920,12 +920,12 @@ def change_vertices_order(triangle):
 
     Parameters
     ----------
-    triangle : np.ndarray, shape(1, 3)
+    triangle : ndarray, shape(1, 3)
         array of 3 vertices making up a triangle
 
     Returns
     -------
-    new_triangle: np.ndarray, shape(1, 3)
+    new_triangle : ndarray, shape(1, 3)
         new array of vertices making up a triangle in the opposite winding
         order of the given parameter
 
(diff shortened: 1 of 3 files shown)


fury/utils.py
```diff
@@ -998,22 +997,22 @@ def update_actor(actor):
...
 
 def get_bounds(actor):
-    """Returns Bounds of actor.
+    """Return Bounds of actor.
 
     Parameters
-    --------------
-    actor : actor
+    ----------
+    actor : vtk.vtkActor
```
Garyfallidis: Why vtk.vtkActor? Just say fury.actor

Garyfallidis: It could be vtkProp etc.

fury/utils.py
```diff
@@ -882,100 +882,99 @@ def normals_from_v_f(vertices, faces):
...
     Parameters
-    -------------
+    ----------
     actor : actor
 
     Returns
-    ---------
-    vertices : ndarray
+    -------
+    vertices : np.ndarray
```
Garyfallidis: We always say ndarray why to say now np.ndarray?

skoudoro: using the `np.`before permit to be recognize by intersphinx and the user can click and be forwarded to the right documentation (here, the numpy doc)

Garyfallidis: Clearly not necessary - everyone knows numpy these days

skoudoro: I can not agree with this statement. I would say everybody knows how to go to NumPy doc. 
Instead of doing a google search, this notation forwards you automatically to the right place in Numpy doc. 
I will put it back even if I do not agree with your previous statement.

Garyfallidis: Well if you feel very strong about it you can make another PR where you change this everywhere not only here.

PR #224, opened 2020-03-31
WIP: Picking tutorial

Garyfallidis:
Picking objects in actors


Garyfallidis:
There is a better version at #224

PR #225, opened 2020-03-31
update tutorial folder structure

skoudoro:
Small tutorial update:

- update tutorial folder structure
- Replace `Other Features` section by `Actors` section


Garyfallidis:
I suggest that the actors section be renamed to basic as now we are mixing both showing actors, windows and animations in the actions section.

PR #227, opened 2020-04-01
[Fix] update streamlines default color

### LINKED ISSUE #135: Backward compatibilities problem with streamtube
## Description

After playing with `dipy_horizon`, you can see that streamlines are in Black&White instead of color when you use the master branch instead of the release 0.4.0.
 
The recent PR (#129) change the original behavior so :

- we have to get back to the original behavior
- find a workaround to manage the new color features  (look at the old one [here](https://github.com/fury-gl/fury/blob/maint/v0.4.0/fury/actor.py#L445))

skoudoro:
Due to backward compatibilities, I revert a previous change. By default, the streamlines will be colored with a standard orientation colormap instead of no color.

fix #135


Garyfallidis:
Works thanks @skoudoro !

Issue #226, opened 2020-04-01
Adding DoubleClick event


skoudoro:
As suggested by [this comment](https://github.com/fury-gl/fury/pull/222#discussion_r401137696), It would be good to add the DoubleClick feature. This callback can be really useful for many UI elements.

Nibba2018:
I would like to work on this issue.

PR #231, opened 2020-04-03
[WIP] Adding Double Click Callback

Nibba2018:
Referencing #226 , Adding Double Left Click Callback.
I will be adding examples & unittests soon.

Let me know if I am doing it wrong.
Thank You.

### DIFF
--- a/fury/interactor.py
+++ b/fury/interactor.py
@@ -177,11 +177,11 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
             self.initial_state = self.trackball_camera.GetInteractor().GetEventPosition()
             if self.initial_state == self.click_history:
                 # stop single click event here...
+                self.event.abort()
                 print("Double Clicked Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
-                self.event.abort()
+                    prop.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
             else:
                 # print("Initial state:", self.initial_state)
                 print("Single Click Detected.")
@@ -208,11 +208,11 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
                     self.propagate_event(evt, prop)
                 self.nb_left_clicks = 0
             else:
+                self.event.abort()
                 print("Double Click Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
-                self.event.abort()
+                    prop.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
 
         self.trackball_camera.OnLeftButtonDown()
 
--- a/fury/interactor.py
+++ b/fury/interactor.py
@@ -142,8 +142,8 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
             self.on_mouse_wheel_forward(obj, evt)
         elif evt == "MouseWheelBackwardEvent":
             self.on_mouse_wheel_backward(obj, evt)
-        elif evt == "LeftButtonDoubleClickEvent":
-            self.on_left_button_double_click(obj, evt)
+        # elif evt == "LeftButtonDoubleClickEvent":
+        #     self.on_left_button_double_click(obj, evt)
 
         self.event.reset()  # Event fully processed.
 
@@ -157,13 +157,13 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         if not self.event.abort_flag:
             self.trackball_camera.OnLeftButtonDown()
 
-    def on_left_button_up(self, _obj, evt):
-        self.left_button_down = False
-        self.propagate_event(evt, *self.selected_props["left_button"])
-        self.selected_props["left_button"].clear()
-        self.trackball_camera.OnLeftButtonUp()
+    # def on_left_button_up(self, _obj, evt):
+    #     self.left_button_down = False
+    #     self.propagate_event(evt, *self.selected_props["left_button"])
+    #     self.selected_props["left_button"].clear()
+    #     self.trackball_camera.OnLeftButtonUp()
 
-    def on_left_button_double_click(self, _obj, evt):
+    def on_left_button_up(self, _obj, evt):
         self.left_button_down = False
         prop = self.get_prop_at_event_position()
 
--- a/fury/interactor.py
+++ b/fury/interactor.py
@@ -87,8 +87,6 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         self.reset_pixel_distance = 5
         self.click_history = []
 
-        self.DoubleClickEvent = vtk.vtkCommand.UserEvent+1
-
         self.selected_props = {"left_button": set(),
                                "right_button": set(),
                                "middle_button": set()}
@@ -181,7 +179,7 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
                 print("Double Clicked Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
+                    self.propagate_event("DoubleClickEvent", prop)
             else:
                 # print("Initial state:", self.initial_state)
                 print("Single Click Detected.")
@@ -212,7 +210,7 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
                 print("Double Click Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
+                    self.propagate_event("DoubleClickEvent", prop)
 
         self.trackball_camera.OnLeftButtonDown()
 
@@ -361,7 +359,6 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         self.AddObserver("RightButtonReleaseEvent", self._process_event)
         self.AddObserver("MiddleButtonPressEvent", self._process_event)
         self.AddObserver("MiddleButtonReleaseEvent", self._process_event)
-        self.AddObserver("DoubleClickEvent", self._process_event)
 
         # Windows and special events.
         # TODO: we ever find them useful we could support them.
--- a/fury/interactor.py
+++ b/fury/interactor.py
@@ -87,6 +87,8 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         self.reset_pixel_distance = 5
         self.click_history = []
 
+        self.DoubleClickEvent = vtk.vtkCommand.UserEvent+1
+
         self.selected_props = {"left_button": set(),
                                "right_button": set(),
                                "middle_button": set()}
@@ -145,8 +147,8 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
             self.on_mouse_wheel_forward(obj, evt)
         elif evt == "MouseWheelBackwardEvent":
             self.on_mouse_wheel_backward(obj, evt)
-        # elif evt == "LeftButtonDoubleClickEvent":
-        #     self.on_left_button_double_click(obj, evt)
+        elif evt == "DoubleClickEvent":
+            self.on_double_click(obj, evt)
 
         self.event.reset()  # Event fully processed.
 
@@ -175,13 +177,14 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
             self.initial_state = self.trackball_camera.GetInteractor().GetEventPosition()
             if self.initial_state == self.click_history:
                 # stop single click event here...
-                print("Double Clicked... [Aborts previous single click]")
+                print("Double Clicked Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.propagate_event(evt, prop)
+                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
+                self.event.abort()
             else:
                 # print("Initial state:", self.initial_state)
-                print("Single Clicked...")
+                print("Single Click Detected.")
                 if prop is not None:
                     # Single Clicked Events...
                     self.selected_props["left_button"].add(prop)
@@ -198,19 +201,29 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
 
             if dist_moved > self.reset_pixel_distance:
                 self.click_history = final_state
-                print("Single Clicked...")
+                print("Single Clicked Detected.")
                 if prop is not None:
                     # Single Clicked Events...
                     self.selected_props["left_button"].add(prop)
                     self.propagate_event(evt, prop)
                 self.nb_left_clicks = 0
             else:
-                print("Double Clicked... [Aborts previous single click]")
+                print("Double Click Detected. [Aborts previous single click]")
                 self.nb_left_clicks = 0
                 if prop is not None:
-                    self.propagate_event(evt, prop)
+                    self.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
+                self.event.abort()
 
         self.trackball_camera.OnLeftButtonDown()
+
+    def on_double_click(self, obj, evt):
+        print("Inside double click Method")
+        prop = self.get_prop_at_event_position()
+        if prop is not None:
+            self.propagate_event(evt, prop)
+
+        if not self.event.abort_flag:
+            self.trackball_camera.OnLeftButtonDown()
 
     def on_right_button_down(self, _obj, evt):
         self.right_button_down = True
@@ -348,6 +361,7 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         self.AddObserver("RightButtonReleaseEvent", self._process_event)
         self.AddObserver("MiddleButtonPressEvent", self._process_event)
         self.AddObserver("MiddleButtonReleaseEvent", self._process_event)
+        self.AddObserver("DoubleClickEvent", self._process_event)
 
         # Windows and special events.
         # TODO: we ever find them useful we could support them.
--- a/fury/interactor.py
+++ b/fury/interactor.py
@@ -170,27 +170,28 @@ class CustomInteractorStyle(vtk.vtkInteractorStyleUser):
         self.nb_left_clicks += 1
         prop = self.get_prop_at_event_position()
 
-        initial_state = self.trackball_camera.GetInteractor().GetEventPosition()
+        if self.nb_left_clicks == 1:
+            self.initial_state = self.trackball_camera.GetInteractor().GetEventPosition()
+            print("Initial state:", self.initial_state)
+            print("Single Clicked...")
 
-        print("Initial state:", initial_state)
-
-        final_state = self.trackball_camera.GetInteractor().GetEventPosition()
-
-        print("Final state:", final_state)
-
-        x_dist = final_state[0] - initial_state[0]
-        y_dist = final_state[1] - initial_state[1]
-
-        dist_moved = int((x_dist**2 + y_dist**2)**0.5)
-
-        if dist_moved > self.reset_pixel_distance:
-            self.nb_left_clicks = 1
-        
         if self.nb_left_clicks == 2:
-            print("Double Clicked...")
-            self.nb_left_clicks = 0
-            if prop is not None:
-                self.propagate_event(evt, prop)
+            final_state = self.trackball_camera.GetInteractor().GetEventPosition()
+            print("Final state:", final_state)
+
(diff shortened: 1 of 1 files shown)


Nibba2018:
@skoudoro I need some suggestions here...

According to this [example](https://lorensen.github.io/VTKExamples/site/Cxx/Interaction/DoubleClick/) they have overridden the OnLeftButtonDown by inheriting from vtkInteractorStyleTrackballCamera class. But the rest of the callbacks are implemented in a different way. So should I inherit both vtkInteractorStyleUser and vtkInteractorStyleTrackballCamera and implement the double click? or should I create a separate class for it?

I am also unable to implement my double click callback on an actor hence not being able to test my implementation. I need some help here with that.

Thanks.

fury/interactor.py
```diff
@@ -142,6 +142,8 @@ def _process_event(self, obj, evt):
             self.on_mouse_wheel_forward(obj, evt)
         elif evt == "MouseWheelBackwardEvent":
             self.on_mouse_wheel_backward(obj, evt)
+        elif evt == "LeftButtonDoubleClickEvent":
```
skoudoro, line 145: this event does not exist in vtk. We need to create this custom event. I recommend you to comment this part for the moment

fury/interactor.py
```diff
@@ -161,6 +163,22 @@ def on_left_button_up(self, _obj, evt):
         self.selected_props["left_button"].clear()
         self.trackball_camera.OnLeftButtonUp()
 
+    def on_left_button_double_click(self, _obj, evt):
```
skoudoro, line 166: this function will never be called. you should be called trough the `leftbuttondown`

fury/interactor.py
```diff
@@ -161,6 +163,22 @@ def on_left_button_up(self, _obj, evt):
...
+        self.left_button_down = False
+        prop = self.get_prop_at_event_position()
+
+        final_state = initial_state = []
+        self.trackball_camera.OnLeftButtonDown()
+        self.trackball_camera.GetInteractor.GetEventPosition(initial_state)
```
skoudoro, line 172: missing parenthesis for `GetInteractor`. this is a function

fury/interactor.py
```diff
@@ -161,6 +163,22 @@ def on_left_button_up(self, _obj, evt):
...
+        self.trackball_camera.OnLeftButtonUp()
+
+        self.trackball_camera.OnLeftButtonDown()
+        self.trackball_camera.GetInteractor.GetEventPosition(final_state)
```
skoudoro, line 176: missing parenthesis for `GetInteractor`. this is a function

fury/interactor.py
Nibba2018: hello @skoudoro , can you explain a bit more, i did not understand.

Nibba2018: Understood

Nibba2018:
Here's a test that I am running to test double click implementation where I change the color of the cube when double clicked on it.

```python
from fury import actor, window, interactor
import numpy as np

cube = actor.cube(np.random.rand(1, 3), np.random.rand(1,3), (1,1,1), heights=np.random.rand(5))

def change_color(iren, obj):
	obj.SetColor(np.random.rand(3))
	iren.force_render()

interactor_style = interactor.CustomInteractorStyle()
interactor_style.add_active_prop(cube)
interactor_style.add_callback(cube, "LeftButtonPressEvent", change_color)

current_size = (800, 800)
showm = window.ShowManager(size=current_size, title="Double Click Test")
showm.scene.add(cube)
showm.start()
```
Apparantly I am having an error message being printed which I am not able to understand why. The output:
```bash
Initial state: (390, 379)
Final state: (390, 379)
Initial state: (390, 379)
Final state: (390, 379)
Double Clicked...
interactor is none
event name is LeftButtonPressEvent
```
`interactor is none`
`event name is LeftButtonPressEvent`

skoudoro:
1. Can you differentiate `LeftButtonPressEvent` and `DoubleClick`? For example, on DoubleClick, switch color between red and blue, for a single click switch between yellow and purple. 

2. You do not need `interactor_style = interactor.CustomInteractorStyle()`. This is created by `ShowManager` and if you want to access it, you just need to do `showm.iren`. This will avoid your error

3. `showm.iren.add_callback (cube, ...)`

nice progress!

Nibba2018:
Hello @skoudoro , I tried `showm.iren.add_callback(...)` but it gives me an Attribute Error
```python
  File "double_click.py", line 13, in <module>
    showm.iren.add_callback(cube, "LeftButtonPressEvent", change_color)
AttributeError: 'vtkRenderingOpenGL2Python.vtkXRenderWindowInteract' object has no attribute 'add_callback'
```

fury/interactor.py
```diff
@@ -142,25 +144,57 @@ def _process_event(self, obj, evt):
             self.on_mouse_wheel_forward(obj, evt)
         elif evt == "MouseWheelBackwardEvent":
             self.on_mouse_wheel_backward(obj, evt)
+        # elif evt == "LeftButtonDoubleClickEvent":
+        #     self.on_left_button_double_click(obj, evt)
```
Nibba2018, line 149: To differentiate Single clicks and double clicks wouldn't i need custom events like these?

Nibba2018: Or we could pass two methods, one for Single click and the other for double clicks...

skoudoro:
> I tried showm.iren.add_callback(...) but it gives me an Attribute Error

ok, need to check, maybe it is `showm.style.add_callback` instead

fury/interactor.py
skoudoro: I believe we can pass 2 methods, to test, and then, we will have to create a new event.
doubleclick should cancel singleclick action/propagation when it is detected

Nibba2018:
Hello @skoudoro , I am successfully able to detect and differentiate between single clicks & double clicks. But I am unable to understand how to abort or stop the propagation of the previously initiated single click event. Can you give me some idea where I should look for a solution... Thank you.

Nibba2018:
Also, will adding the line `self.AddObserver("LeftButtonDoubleClickEvent", self._process_event)` sufficient enough to create a new event for double click? or do I need to do something else?

skoudoro:
> Hello @skoudoro , I am successfully able to detect and differentiate between single clicks & double clicks. But I am unable to understand how to abort or stop the propagation of the previously initiated single click event. Can you give me some idea where I should look for a solution... Thank you.

Hi @Nibba2018, That's good news! To stop the propagation,you need to use `self.event.abort()`.
But, to use this method at the right place can be tricky, I need to go deeper into it.

> Also, is the line self.AddObserver("LeftButtonDoubleClickEvent", self._process_event) sufficient enough to create a new event for double click? or do I need to do something else?

No, it is not because this custom event is never invoked. We need to create a custom event named `LeftButtonDoubleClickEvent` and then invoke it on our interactor. You will have to look at VTK examples to see how to manage a new custom event.

Nibba2018:
Hello @skoudoro , I did some research and have come up with the following conclusions:

The custom event can be implemented by subclassing [vtkCommand](https://vtk.org/doc/nightly/html/classvtkCommand.html#details). 
I was also able to find an [example](https://github.com/Kitware/VTK/blob/master/Common/Core/Testing/Python/TestCommand.py) related to vtkCommand, but I am not sure how it works or how it is being implemented. I will be looking into it further.
I also found some contradictions regarding subclassing vtkCommand [here](https://vtk.org/Wiki/VTK/Python_Wrapper_Enhancement#Wrap_vtkCommand_and_allow_it_to_be_subclassed), but I am not sure whether it's outdated or not.

I will try implementing vtkCommand subclasses and see if it works.

skoudoro:
Hi @Nibba2018,

Thank you for your research and all this information. You should look at [this nice example](https://github.com/Beastmaster/itk-python-example/blob/master/vtkUserDefinedEvent.py) too.
Basically, it seems that you need to follow this step:
1. create DoubleClickEvent in the init of our custom interactor style : `self.DoubleClickEvent = vtk.vtkCommand.UserEvent+1
2. invoke this event on the leftclick fonction when you detect it: `self.invokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
3. connect DoubleClickEvent with your double click function.

Does it make sense?

Nibba2018:
Thank you @skoudoro , It makes it very clear. I will implement this right away.

Nibba2018:
@skoudoro I am using this following for testing:
```python3
from fury import actor, window, interactor
import numpy as np

cube = actor.cube(np.random.rand(1, 3), np.random.rand(1,3), (1,1,1), heights=np.random.rand(5))

def single_click(iren, obj):
	print("Single Click")
	iren.force_render()

def double_click(iren, obj):
	print("Double Click")
	iren.force_render()

current_size = (800, 800)
showm = window.ShowManager(size=current_size, title="Double Click Test")
showm.scene.add(cube)
showm.style.add_callback(cube, "LeftButtonPressEvent", single_click)
showm.style.add_callback(cube, "DoubleClickEvent", double_click)
showm.start()
```
The changes are able to detect and differentiate clicks but are not able to propagate Double Click Event Callback.

Nibba2018:
@skoudoro , I ran the debugger and found that during single clicks, the control moves from `self.propagate` to `prop.InvokeEvent` and then to `_add_callback` function. But in the case of Double Clicks, the control doesn't go to `_add_callback` after the `prop.InvokeEvent` has been executed...

skoudoro:
Interesting, I have no idea at this point, I need to go deeper.

Maybe @MarcCote have an idea about the DoubleClik

skoudoro:
I suppose you are missing a `self.InvokeEvent(...)` to propagate the event to himself (the interactor)

`prop.InvokeEvent` propagates the event to all dependent actor

So you need both

Nibba2018:
@skoudoro , Initially I tried with `self.InvokeEvent` but it didn't work, so I thought I should be using `prop.InvokeEvent` instead as it was previously implemented in `self.propagate_event`

fury/interactor.py
```diff
@@ -142,25 +147,84 @@ def _process_event(self, obj, evt):
...
+
+        if self.nb_left_clicks == 1:
+            self.initial_state = self.trackball_camera.GetInteractor().GetEventPosition()
+            if self.initial_state == self.click_history:
+                # stop single click event here...
+                self.event.abort()
+                print("Double Clicked Detected. [Aborts previous single click]")
+                self.nb_left_clicks = 0
+                if prop is not None:
+                    prop.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
```
MarcCote, line 184: What you want here is: `self.propagate_event("DoubleClickEvent", prop)`.

Also, from my experience, the proper way to call `InvokeEvent` is `prop.InvokeEvent('DoubleClickEvent')`. But use our centralized method `self.propagate_event`.

fury/interactor.py
```diff
@@ -83,6 +83,11 @@ def __init__(self):
         self.right_button_down = False
         self.middle_button_down = False
         self.active_props = set()
+        self.nb_left_clicks = 0
+        self.reset_pixel_distance = 5
+        self.click_history = []
+
+        self.DoubleClickEvent = vtk.vtkCommand.UserEvent+1
```
MarcCote, line 90: I don't think you need that for our custom interactor.

fury/interactor.py
```diff
@@ -142,25 +147,84 @@ def _process_event(self, obj, evt):
...
+                    # Single Clicked Events...
+                    self.selected_props["left_button"].add(prop)
+                    self.propagate_event(evt, prop)
+                self.nb_left_clicks = 0
+            else:
+                self.event.abort()
+                print("Double Click Detected. [Aborts previous single click]")
+                self.nb_left_clicks = 0
+                if prop is not None:
+                    prop.InvokeEvent(self.DoubleClickEvent, 'DoubleClickEvent')
```
MarcCote, line 215: Same comment as above.

fury/interactor.py
```diff
@@ -297,6 +361,7 @@ def SetInteractor(self, interactor):
         self.AddObserver("RightButtonReleaseEvent", self._process_event)
         self.AddObserver("MiddleButtonPressEvent", self._process_event)
         self.AddObserver("MiddleButtonReleaseEvent", self._process_event)
+        self.AddObserver("DoubleClickEvent", self._process_event)
```
MarcCote, line 364: From what I could gather VTK doesn't truly support DoubleClickEvent by default, so this observer will never fire up. That's why people implemented the double click behavior from scratch like you just did (i.e. tracking clicks count).

So, you can remove this line and should still be working fine.

Nibba2018:
Hello @MarcCote , thank you for the review I will apply the changes and will let you know if I have any questions.

Nibba2018:
@MarcCote Thank for the suggestions

Nibba2018:
Hello @MarcCote , looks like single clicks arent working the way they are supposed to and hence causing tests to fail. Failed tests start from `test_ui_line_slider_2d_horizontal_bottom` and then eventually get a `segmentation fault` error. Not sure why this is happening.

MarcCote:
Hi @Nibba2018 I've made some refactoring of your code and made a PR to your branch https://github.com/Nibba2018/fury/pull/1

Note, I'm not sure we care about the click radius actually (i.e. allowing slight movement of the mouse between the clicks), so I've commented it out. The code I proposed is assuming the mouse doesn't move between the clicks. Maybe I'm wrong, though, but it does feel natural when trying it with the script below.

Note2: I had to keep track of custom event (e.g., LeftButtonDoubleClickEvent, etc) since VTK will map all unknown event to the **same** `NoEvent`.

Here's the testing script I've been using to debug the code:
```python
from fury import actor, window, interactor
import numpy as np


cube1 = actor.cube([(0, 0, 0)], np.random.rand(1,3), (1,1,1), heights=3)
cube2 = actor.cube([(0, 0, 0)], np.random.rand(1,3), (1,0.5,0), heights=3)
cube2.SetVisibility(False)

def left_single_click(iren, obj):
    print("> Left Single Click <")
    iren.force_render()

def left_double_click(iren, obj):
    print("> Left Double Click <")
    cube1.SetVisibility(not bool(cube1.GetVisibility()))
    cube2.SetVisibility(not bool(cube2.GetVisibility()))
    iren.force_render()

def right_single_click(iren, obj):
    print("> Right Single Click <")
    iren.force_render()

def right_double_click(iren, obj):
    print("> Right Double Click <")
    cube1.SetVisibility(not bool(cube1.GetVisibility()))
    cube2.SetVisibility(not bool(cube2.GetVisibility()))
    iren.force_render()

def middle_single_click(iren, obj):
    print("> Middle Single Click <")
    iren.force_render()

def middle_double_click(iren, obj):
    print("> Middle Double Click <")
    cube1.SetVisibility(not bool(cube1.GetVisibility()))
    cube2.SetVisibility(not bool(cube2.GetVisibility()))
    iren.force_render()


current_size = (800, 800)
showm = window.ShowManager(size=current_size, title="Double Click Test")
showm.scene.add(cube1, cube2)
showm.style.add_callback(cube1, "LeftButtonPressEvent", left_single_click)
showm.style.add_callback(cube2, "LeftButtonPressEvent", left_single_click)
showm.style.add_callback(cube1, "LeftButtonDoubleClickEvent", left_double_click)
showm.style.add_callback(cube2, "LeftButtonDoubleClickEvent", left_double_click)
showm.style.add_callback(cube1, "RightButtonPressEvent", right_single_click)
showm.style.add_callback(cube2, "RightButtonPressEvent", right_single_click)
showm.style.add_callback(cube1, "RightButtonDoubleClickEvent", right_double_click)
showm.style.add_callback(cube2, "RightButtonDoubleClickEvent", right_double_click)
showm.style.add_callback(cube1, "MiddleButtonPressEvent", middle_single_click)
showm.style.add_callback(cube2, "MiddleButtonPressEvent", middle_single_click)
showm.style.add_callback(cube1, "MiddleButtonDoubleClickEvent", middle_double_click)
showm.style.add_callback(cube2, "MiddleButtonDoubleClickEvent", middle_double_click)
showm.start()
```

Nibba2018:
How did it pass codecov tests? Aren't I supposed to add tests for the newly added changes?

MarcCote:
@skoudoro in Codecov's post it says "No coverage uploaded for pull request base". What can we do about that?

Nibba2018:
@MarcCote  I think the `test_interactor.py` doesn't work on Windows and Mac OS maybe that's why tests for it are disabled?

skoudoro:
> @skoudoro in Codecov's post it says "No coverage uploaded for pull request base". What can we do about that?

We can do anything @MarcCote. Codecov is just lost since it didnot run on your branch before the merge. It can not compare this commit ea98aba from @Nibba2018 and your commit 6a2a782.

It should come back to normal after a new commit I suppose.

fury/interactor.py
```diff
@@ -145,9 +158,49 @@ def _process_event(self, obj, evt):
...
+        # if self.history[before_last_event]["prop"] \
+        #       != self.history[last_event]["prop"]:
+            # return False
+
+        # dx = self.history[before_last_event]["pos"][0]\
+        #  - self.history[last_event]["pos"][0]
+        # dy = self.history[before_last_event]["pos"][1]\
+        #  - self.history[last_event]["pos"][1]
+        # if (dx**2 + dy**2) > self.click_radius_square:
+            # return False
```
MarcCote, line 181: We need to decide what we actually want as the double click behavior.

I think it is fine as it is right now, i.e. it is assuming a click radius of 0 because if the user moves their mouse, there will be MouseMoveEvent added to the history between the click events preventing the *DoubleClickEvent to be invoked.

fury/interactor.py
```diff
@@ -331,10 +401,17 @@ def _callback(_obj, event_name):
             # Update event information.
             interactor_ = self.GetInteractor()
             if interactor_ is not None:
-                self.event.update(event_name, interactor_)
+                # self.event.update(event_name, interactor_)
```
MarcCote, line 404: We can delete this line since it was moved in `_process_event`

fury/interactor.py
```diff
@@ -331,10 +401,17 @@ def _callback(_obj, event_name):
...
             else:
                 print('interactor is none')
                 print('event name is', event_name)
 
+        if vtk.vtkCommand.GetEventIdFromString(event_type) == 0:
+            if event_type not in self.event2id:
+                self.event2id[event_type] = \
+                    vtk.vtkCommand.UserEvent + len(self.event2id) + 1
+
+            event_type = self.event2id[event_type]
```
MarcCote, line 415: Needs a comment describing this section: dealing with custom events not defined in VTK.

fury/interactor.py
Nibba2018: Hello @MarcCote , I agree with your points. The current implementation feels natural to me as well. I think we should remove this portion.

fury/interactor.py
Nibba2018: okay :+1:

fury/interactor.py
```diff
@@ -77,13 +79,17 @@ def __init__(self):
...
+        self.event2id = {}  # To map an event's name to an ID.
 
         # Define some interaction states
         self.left_button_down = False
         self.right_button_down = False
         self.middle_button_down = False
         self.active_props = set()
 
+        self.history = deque(maxlen=10)  # Events history.
+        # self.click_radius_square = 5 ** 2  # In pixel.
```
MarcCote, line 91: We can also remove this one.

fury/interactor.py
MarcCote: Great. Let remove it then. @skoudoro, thoughts?

skoudoro: I agree with your point too, it is fine as it is. I just tested it and it works well. Thank you @MarcCote and @Nibba2018!
I just think that we need to add them to the UI like we did [here](https://github.com/fury-gl/fury/blob/master/fury/ui.py#L99).
Furthermore, there is a small bug to fix. It seems when you doubleclick on the "background",  there is no object so we need to manage `None` case as you can see below.
Capture d’écran 2020-05-25 à 06 43 43

skoudoro:
Thank you for the quick update @Nibba2018.

It works well. I wonder whether we should add the notion of time for the double click behavior. Currently, if I have a delay of 10sec, 30sec or 1min between 2 clicks, the double click event popup which is not correct. Maybe we need a timer that empties the history. 

You can try it: left click, wait 30sec, left click -> and the DoubleClick event is triggered

What do you think @Nibba2018 and @MarcCote?

skoudoro:
A second thought, I think we should keep that for a new PR. This version is good as it is.

MarcCote:
I agree, a timer would be nice but is better suited for a new PR.

MarcCote:
Regarding the tests. See https://github.com/Nibba2018/fury/pull/2. Running your test with the provided `test_double_click_events.log.gz` with XVFB on my machine produced the same error as with Travis. I made a new one and change the test a bit to make it easier to distinguish between the different double clicks.

MarcCote:
@skoudoro is it expected that some tests on Windows fail?

skoudoro:
Yes @MarcCote. Due to the new VTK 9. Since there are no wheels for Linux, tests passed normally. This is why I fixed them on #245.

skoudoro:
I would like to say this PR is ready to go. But I prefer to wait for #245 to be merged and we will have to rebase this PR to be sure to have no surprise.

Thanks @Nibba2018 and @MarcCote for this work. Let's wait for now.

skoudoro:
Can you rebase your PR @Nibba2018? It should be ok now.

skoudoro:
Thank you for the rebase @Nibba2018. 

Any additional comments? I will merge this by the end of the day.

PR #232, opened 2020-04-04
Improve packaging

skoudoro:
It seems that some static files were not included during the packaging.
This PR update the directive for including them.


Garyfallidis:
Ready to merge if nobody has anything to add.

PR #248, opened 2020-05-31
Setting Bounding Box for TextBlock2D

Nibba2018:
Working example of setting a BBox for a `vtkTextActor` using `vtkTextRepresentation` and `vtkTextWidget`:

```python
import vtk

renderer = vtk.vtkRenderer()
renderer.SetBackground(0.1, 0.2, 0.4)

render_window = vtk.vtkRenderWindow()
render_window.AddRenderer(renderer)
render_window.SetSize(400, 400)

interactor = vtk.vtkRenderWindowInteractor()
interactor.SetRenderWindow(render_window)

# Create the TextActor
text_actor = vtk.vtkTextActor()
text_actor.SetInput("This is a test")
text_actor.GetTextProperty().SetColor((0, 1, 1))

# Create the text representation. Used for positioning the text_actor
text_representation = vtk.vtkTextRepresentation()

text_representation.GetPositionCoordinate().SetCoordinateSystemToDisplay()
text_representation.GetPosition2Coordinate().SetCoordinateSystemToDisplay()

text_representation.GetPositionCoordinate().SetValue(200, 200)
text_representation.GetPosition2Coordinate().SetValue(200, 200)

text_representation.SetShowBorderToOff()

text_actor.GetTextProperty().SetFontSize(0)

# Create the TextWidget
# Note that the SelectableOff method MUST be invoked!
# According to the documentation :
#
# SelectableOn/Off indicates whether the interior region of the widget can be
# selected or not. If not, then events (such as left mouse down) allow the user
# to "move" the widget, and no selection is possible. Otherwise the
# SelectRegion() method is invoked.
text_widget = vtk.vtkTextWidget()
text_widget.SetRepresentation(text_representation)
text_widget.SetInteractor(interactor)
text_widget.SetTextActor(text_actor)
text_widget.SelectableOff()
text_widget.ResizableOff()
text_widget.On()

print(text_widget.GetTextActor())
print(text_actor.GetPositionCoordinate(), text_actor.GetPosition2Coordinate())

interactor.Initialize()
render_window.Render()
interactor.Start()
```

### DIFF
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -1084,7 +1084,7 @@ class TextBlock2D(UI):
 
     def __init__(self, text="Text Block", font_size=18, font_family='Arial',
                  justification='left', vertical_justification="bottom",
-                 bold=False, italic=False, shadow=False, size=(100, 100),
+                 bold=False, italic=False, shadow=False, size=None,
                  color=(1, 1, 1), bg_color=None, position=(0, 0)):
         """
         Parameters
@@ -1115,10 +1115,14 @@ class TextBlock2D(UI):
             Size (width, height) in pixels of the text bounding box.
         """
         super(TextBlock2D, self).__init__(position=position)
-        self.resize(size)
+        if size is not None:
+            self.actor.SetTextScaleModeToProp()
+            self.resize(size)
+        else:
+            self.actor.SetTextScaleModeToNone()
+            self.font_size = font_size
         self.color = color
         self.background_color = bg_color
-        self.font_size = font_size
         self.font_family = font_family
         self.justification = justification
         self.bold = bold
@@ -1129,7 +1133,6 @@ class TextBlock2D(UI):
 
     def _setup(self):
         self.actor = vtk.vtkTextActor()
-        self.actor.SetTextScaleModeToProp()
         self.actor.GetPosition2Coordinate().SetCoordinateSystemToViewport()
         self._background = None  # For VTK < 7
         self.handle_events(self.actor)
@@ -1208,6 +1211,7 @@ class TextBlock2D(UI):
         size : int
             Text font size.
         """
+        self.actor.SetTextScaleModeToNone()
         self.actor.GetTextProperty().SetFontSize(size)
 
     @property
@@ -1451,6 +1455,9 @@ class TextBlock2D(UI):
     def _get_size(self):
         if self._background is not None:
             return self._background.size
+
+        if not self.actor.GetTextScaleMode():
+            return self.font_size * 1.2, self.font_size * 1.2
 
         size = np.array([0, 0])
         size[0] = self.actor.GetPosition2()[0] - self.actor.GetPosition()[0]
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -1080,6 +1080,8 @@ class TextBlock2D(UI):
         Makes text italicised.
     shadow : bool
         Adds text shadow.
+    size : (int, int)
+        Size (width, height) in pixels of the text bounding box.
     """
 
     def __init__(self, text="Text Block", font_size=18, font_family='Arial',
@@ -1137,7 +1139,7 @@ class TextBlock2D(UI):
         self.handle_events(self.actor)
 
     def resize(self, size):
-        """Resize the button.
+        """Resize TextBlock2D.
 
         Parameters
         ----------
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -1145,7 +1145,7 @@ class TextBlock2D(UI):
         size : (int, int)
             Text bounding box size(width, height) in pixels.
         """
-
+        self.actor.SetTextScaleModeToProp()
         position2 = self.position + size
         self.actor.SetPosition2(*position2)
 
--- a/fury/ui.py
+++ b/fury/ui.py
@@ -1082,10 +1082,11 @@ class TextBlock2D(UI):
         Adds text shadow.
     """
 
-    def __init__(self, text="Text Block", font_size=18, font_family='Arial',
+    def __init__(self, text="Text Block", font_size=0, font_family='Arial',
                  justification='left', vertical_justification="bottom",
-                 bold=False, italic=False, shadow=False,
-                 color=(1, 1, 1), bg_color=None, position=(0, 0)):
+                 bold=False, italic=False, shadow=False, border=False,
+                 color=(1, 1, 1), bg_color=None, bbox_size=(100, 100),
+                 position=(0, 0)):
         """
         Parameters
         ----------
@@ -1111,7 +1112,13 @@ class TextBlock2D(UI):
             Makes text italicised.
         shadow : bool
             Adds text shadow.
+        border : bool
+            show bounding box.
+        bbox_size : (int, int)
+            width & height of bounding box
         """
+        self.bbox_size = bbox_size
+        self.border = border
         super(TextBlock2D, self).__init__(position=position)
         self.color = color
         self.background_color = bg_color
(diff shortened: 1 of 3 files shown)


Nibba2018:
Hello @skoudoro , I am not able to understand how to refactor this example code to make it work with the current FURY API. The current implementation here gives no error but doesn't work as expected.

fury/ui.py
```diff
@@ -1071,9 +1071,9 @@ class TextBlock2D(UI):
         Adds text shadow.
     """
 
-    def __init__(self, text="Text Block", font_size=18, font_family='Arial',
+    def __init__(self, text="Text Block", font_size=None, font_family='Arial',
                  justification='left', vertical_justification="bottom",
-                 bold=False, italic=False, shadow=False,
+                 bold=False, italic=False, shadow=False, position2=None,
```
skoudoro, line 1076: `position2` should not be a parameter. You should add a parameter `size` to respect the API with the other UI.  With position and size, you can compute and update `position2`.

fury/ui.py
```diff
@@ -1100,11 +1100,20 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
             Makes text italicised.
         shadow : bool
             Adds text shadow.
+        border : bool
+            show bounding box.
+        bbox_size : (int, int)
+            width & height of bounding box
```
skoudoro, line 1106: you can remove the comment above and and size

fury/ui.py
```diff
@@ -1420,6 +1429,30 @@ def _set_position(self, position):
...
+        """ Set text actor position2.
+
+        Parameters
+        ----------
+        position : (float, float)
+            The new position. (x, y) in pixels.
+        """
+        self.actor.SetPosition2(*position2)
+        if self._background is not None:
+            self._background.SetPosition2(*self.actor.GetPosition2())
```
skoudoro, line 1454: I do not think we should make this accessible to the user. The user already have size to update. I think you can remove this block (`_set_position2` and `position2`)

fury/ui.py
```diff
@@ -1420,6 +1429,30 @@ def _set_position(self, position):
...
+
     def _get_size(self):
```
skoudoro, line 1456: You should update `_get_size`

fury/ui.py
Nibba2018: HI @skoudoro , got your point. In that case I have to add a `@size.setter` method right?

fury/ui.py
Nibba2018: sure

fury/ui.py
Nibba2018: Okay!

fury/ui.py
Nibba2018: Sure, also updating this will most likely break the ListBoxItem2D text-overflow fix. So should I fix that within this PR itself?

fury/ui.py
```diff
@@ -1100,11 +1100,18 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
...
         super(TextBlock2D, self).__init__(position=position)
         self.color = color
         self.background_color = bg_color
-        self.font_size = font_size
+        if font_size is not None:
+            self.font_size = font_size
+        if size is not None:
+            self.size = size
+        if font_size is None and size is None:
+            self.font_size = 18
```
skoudoro, line 1114: I do not understand why you need to do all of that. You could add default value in the `__init__`. Can you explain your chaoice ? thank you

fury/ui.py
```diff
@@ -1100,11 +1100,18 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
             Makes text italicised.
         shadow : bool
             Adds text shadow.
+        size : (int, int)
+            Size (width, height) in pixels of the Text.
```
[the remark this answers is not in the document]
Nibba2018: Hello @skoudoro , travis CI is failing because of the text overflow fix that I did in #173 . There I used the `self.size` parameter and now as I have changed the defination of `_get_size`, its failing. I will get that fixed no problem.

fury/ui.py
Nibba2018: Actually, the problem is when I set the `position2` coordinate, the actor ignores the font parameter. Therefore when I set the default size to `(100, 100)`, the size of the actor remains the same(100, 100) no matter how big the font size is.

Nibba2018: I was testing if such if statements would help or not but it didnt

fury/ui.py
```diff
@@ -1100,11 +1100,18 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
             Makes text italicised.
         shadow : bool
             Adds text shadow.
+        size : (int, int)
+            Size (width, height) in pixels of the text bounding box.
         """
         super(TextBlock2D, self).__init__(position=position)
+        if size is not None:
+            self.actor.SetTextScaleModeToProp()
```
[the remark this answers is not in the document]
Nibba2018: Actually I did that incase the user wanted to shift `Scaling Mode` during runtime.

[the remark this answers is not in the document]
Nibba2018: True, will change that

fury/ui.py
```diff
@@ -1100,11 +1100,18 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
...
+            self.resize(size)
+        else:
+            self.actor.SetTextScaleModeToNone()
```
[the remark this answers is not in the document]
Nibba2018: Sure

Nibba2018:
Attaching images for blog.
fury
fury2

fury/ui.py
```diff
@@ -1111,11 +1114,18 @@ def __init__(self, text="Text Block", font_size=18, font_family='Arial',
             Makes text italicised.
         shadow : bool
             Adds text shadow.
+        size : (int, int)
+            Size (width, height) in pixels of the text bounding box.
         """
         super(TextBlock2D, self).__init__(position=position)
+        self.scene = None
+        self.bg = bool(bg_color)
```
skoudoro, line 1122: `self.have_bg` instead of `self.bg`

skoudoro:
Also, It seems, you have to update `fury\tests\test_pick.py:63` because it fails sometimes. Please, Try to make this test more robust with your change

fury/ui.py
Nibba2018: Hello @skoudoro , Currently, the background adjusts itself with the font size of the Text Actor, i.e the background expands if the font size is bigger. Do you want me to prevent that and generate a warning instead?

skoudoro: I do not have this behavior. When I increase the font_size, the background does not expands.

Nibba2018: Did you expand it externally? like `text.font_size = 100` ?

skoudoro: Yes

Nibba2018: Oh yes, I did not add that in `font_size.setter`. Also, what should the behaviour be like if the user passes both `size` and `font_size` parameter?

skoudoro: Priority to `size`, font_size should adapt.

Nibba2018: Okay, for now, I am preventing the user from updating the font size if the text overflows the background.

skoudoro: Great! Thanks, I will play with bit and keep you updated

skoudoro: It does not work for me, I got the following error:

```python
In [3]: from fury import window, ui
In [4]: te = ui.TextBlock2D(position=(50, 50), size=(100, 100), bg_color=(1,1,0))  
In [5]: s = window.ShowManager()                                                                                                         

In [6]: s.scene.add(te)                                                                                                                  

In [7]: s.start()                                                                                                                        

In [8]: te.color =(0,0,0)                                                                                                                

In [9]: s.start()    
In [10]: te.font_size = 50                                                                                                               
ERROR: In /Users/prabhu/src/git/VTKPythonPackage/standalone-build/VTK-source/Rendering/Core/vtkTextActor.cxx, line 902
vtkOpenGLTextActor (0x7fd5f293e120): No render window available: cannot determine DPI.

ERROR: In /Users/prabhu/src/git/VTKPythonPackage/standalone-build/VTK-source/Rendering/Core/vtkTextActor.cxx, line 902
vtkOpenGLTextActor (0x7fd5f293e120): No render window available: cannot determine DPI.

ERROR: In /Users/prabhu/src/git/VTKPythonPackage/standalone-build/VTK-source/Rendering/Core/vtkTextActor.cxx, line 164
vtkOpenGLTextActor (0x7fd5f293e120): UpdateRectangle failed to do so
```

Does it work for you ?

Nibba2018: Yeah gives me these errors when I close the window and Interactor names are printed when I click on the text block. Can you try running it as a script?

skoudoro: Because of this error, the UI is losing the interactor, this is why we see the message from this line https://github.com/fury-gl/fury/blob/master/fury/interactor.py#L385.

It is strange, we need to understand why today.

Nibba2018: @skoudoro , The problem seems to occur when you execute multiple `sm.start()` commands. The same thing happens when you do it with the `fury-gl/master branch` as well.

Nibba2018: Also, the vtk warnings appear when you try to set the font size after closing the window. When the window is closed, no such scene exists. The bounding box constraint check that I added needs the scene to exist to calculate the size of the text actor and therefore throws the following error.

skoudoro: Ok, thank you for your feedback. I will go ahead and merge it.

Nibba2018: Thank you @skoudoro

Issue #254, opened 2020-06-09
unexpected order transparent behavior [VTK9] [Ubuntu 18.04]


MarcCote:
Also, `pytest -svv fury/tests/test_window.py::test_order_transparent` is failing on VTK9.
Here's the screenshot corresponding to the following line in the test
```python
    arr = window.snapshot(scene, fname='green_front.png',
                          offscreen=True, order_transparent=False)
```
![green_front](https://user-images.githubusercontent.com/660004/83987029-dc5cea80-a90c-11ea-97f6-ff0db7f1e9ba.png)

compared to what we had for VTK <9
![green_front](https://user-images.githubusercontent.com/660004/83987089-11693d00-a90d-11ea-94a7-8f394acc9917.png)

PR #258, opened 2020-06-12
Updating test of order_transparency for compatibility with vtk 9

Garyfallidis:
This is in response to https://github.com/fury-gl/fury/issues/254


skoudoro:
Does it mean that the code is correct but the test was not @Garyfallidis?

Thank you for the update!

skoudoro:
Thank you @Garyfallidis, merging

PR #263, opened 2020-07-12
[NF] Deprecated parameters

skoudoro:
This PR adds a decorator to deprecate one or multiple parameters from a function.

# Examples

```python
# The deprecation warnings are not shown in the following examples.
# To deprecate a positional or keyword argument
>>> from fury.deprecator import deprecated_params
>>> @deprecated_params('sig', 'sigma', '0.3')
...    def test(sigma):
...        return sigma
>>> test(2)
2
>>> test(sigma=2)
2
>>> test(sig=2)
/Users/koudoro/miniconda3/envs/fury-env-37/bin/ipython:1: ArgsDeprecationWarning: "sig" was deprecated

* deprecated from version: 1.0
* Use argument "sigma" instead.
2

# It is also possible to replace multiple arguments. The ``old_name``,
# ``new_name`` and ``since`` have to be `tuple` or `list` and contain the
# same number of entries::
>>> @deprecated_params(['a', 'b'], ['alpha', 'beta'], since=['0.2', 0.4], until=['0.7', '0.6'])
...    def test(alpha, beta):
...        return alpha, beta
>>> test(a=2, b=3)
/Users/koudoro/miniconda3/envs/fury-env-37/bin/ipython:5: ArgsDeprecationWarning: "a" was deprecated

* deprecated from version: 0.2
* Use argument "alpha" instead.
  
/Users/koudoro/miniconda3/envs/fury-env-37/bin/ipython:5: ArgsDeprecationWarning: "b" was deprecated

* deprecated from version: 0.4
* Use argument "beta" instead.
(2, 3)
```

### DIFF
--- a/fury/deprecator.py
+++ b/fury/deprecator.py
@@ -11,6 +11,7 @@ the Nibabel package for the copyright and license terms.
 import functools
 import warnings
 import re
+from inspect import signature
 from fury import __version__
 from fury.optpkg import optional_package
 # packaging.version.parse is a third-party utility but is used by setuptools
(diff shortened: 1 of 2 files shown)


skoudoro:
This PR is ready to go!

Issue #264, opened 2020-07-14
Re-implementation of Text Overflow in ListBox2D


Nibba2018:
## Description
Due to the newer implementation of TextBlock2D and its size property in #248 , the Text Overflow bug fix previously implemented in #173 isn't relevant anymore and therefore needs to be re-implemented. #173 was a workaround for the incorrect size calculation given by `TextBlock2D.size`, but the current implementation provides the correct value. 

## Way to reproduce
- [x] Code example
```python
from fury import ui, window
values = ["Violet", "Indigo", "Blue", "Green", "Yellow", "Orange", "Red"]
listBox = ui.ListBox2D(values, position=(70, 40), size=(250, 150))
sm = window.ShowManager(size=(400, 200))
sm.scene.add(listBox)
sm.start()
```
- [x] Relevant images (if any)
**Previously:**
![image](https://user-images.githubusercontent.com/29832615/87430962-8f52ef00-c603-11ea-98ac-a6fc44aaca17.png)

**Currently:**
![image](https://user-images.githubusercontent.com/29832615/87430848-6b8fa900-c603-11ea-87b8-327f6e7f2ee0.png)

skoudoro:
Good catch! Are you working on it?

Nibba2018:
@skoudoro Should I work on this or should I continue with ComboBox and Tab UI?

skoudoro:
1. Combobox
2. Fix this issue
3. back to Tab UI

PR #268, opened 2020-07-19
Clip Text Overflow

Nibba2018:
* This PR is meant to reimplement text-overflow in ListBox2D. It solves #264 .
* A new method is added to `fury.ui` to maintain abstraction. Users can use this method to clip their `TextBlock2D` text for a given width.


fury/utils.py
```diff
@@ -1073,3 +1073,43 @@ def get_bounds(actor):
 
     """
     return actor.GetMapper().GetInput().GetBounds()
+
+
+def clip_overflow(textblock, width):
```
skoudoro, line 1078: This function should stay in UI, later, we will split the UI modules into multiple files.

fury/utils.py
```diff
@@ -1073,3 +1073,43 @@ def get_bounds(actor):
...
+    start_ptr = 0
+    end_ptr = len(original_str)
+    prev_bg = textblock.have_bg
+    textblock.have_bg = False
+
+    if textblock.size[0] == width or textblock.size[0] <= width:
+        textblock.have_bg = prev_bg
+        return original_str
+
+    else:
```
skoudoro, line 1103: `else` not needed, you can remove it, it will help for the lisibility

fury/utils.py
```diff
@@ -1073,3 +1073,43 @@ def get_bounds(actor):
...
+        while start_ptr < end_ptr:
```
skoudoro, line 1104: I wonder about the impact of this `while` for performance.  `clip_overflow` seems to be called often. I need to test it. As a temporary solution, I will be fine but we will have to find a better approach.

fury/utils.py
Nibba2018: This should be outside of any class definations right?

skoudoro: Yes

fury/utils.py
Nibba2018: The complexity of this is around `log2(length)`. The problem is each character has a different height and width.
Also the width of two characters within a string isn't equal to the sum of their individual widths.

skoudoro:
Ok, it works as expected, no issue. Thank you @Nibba2018! I will wait that you move the function and it will be ready for merge

Nibba2018:
@skoudoro , codecov patch seems to fail. Any Idea how I can fix that?

PR #279, opened 2020-08-05
Decreasing the size of the sun in solarsystem tutorial

mlraglin:
This pull request contains a small change to the recent ``viz_solar_system.py`` tutorial. The size of the sun actor is decreased to half its original size.


skoudoro:
Thanks for the update @mlraglin

PR #283, opened 2020-08-06
Standardize colors across API

### LINKED ISSUE #242: Standardize colors across API
Some times we use colors between 0-1 and others between 0-255.  We should use the same range across the API.

### LINKED ISSUE #243: Single Color in Primitive Does Not Work
## Description
Assigning a single color to colors in get_actor_from_primitive does not copy the colors to each vertex.

## Way to reproduce
```python
import numpy as np
from fury import window, utils, actor, primitive
import itertools

vertices, triangles = primitive.prim_frustum()

colors = 255 * np.random.rand(*vertices.shape)

point_actor = actor.point(vertices, point_radius=0.01, colors=colors/255.)
# this does not work
colors = np.array([255, 0. , 0.])
# this does work
# colors = np.array([[255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.],
#                    [255, 0. , 0.]])
frustum_actor = utils.get_actor_from_primitive(vertices=vertices, triangles=triangles, colors=colors, backface_culling=False)

scene = window.Scene()
scene.add(point_actor)
scene.add(actor.axes())
scene.add(frustum_actor)
# frustum_actor.GetProperty().SetOpacity(0.3)
scene.set_camera(position=(10, 5, 7), focal_point=(0, 0, 0))

window.show(scene,size=(600,600), reset_camera=False)
```

### LINKED ISSUE #271: Some issues with actor.box after release 0.6.0
Hello,

- In actor.py, the parameter "scale" in box should be "scales" (lines 1600 and 1611);
- The colors should be at the range of [255, 0, 255]. Otherwise it does not work.

Thank you all for the great work pushing this forward!

skoudoro:
The goal of this PR is to standardize colors across the API [0 -1]

fix #271
fix #242
fix #243 
supersede and fix #269

This PR should be merged after #282

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -446,7 +446,8 @@ def set_polydata_colors(polydata, colors, array_name="colors"):
     """
     vtk_colors = numpy_support.numpy_to_vtk(colors, deep=True,
                                             array_type=vtk.VTK_UNSIGNED_CHAR)
-    vtk_colors.SetNumberOfComponents(3)
+    nb_components = colors.shape[1]
+    vtk_colors.SetNumberOfComponents(nb_components)
     vtk_colors.SetName(array_name)
     polydata.GetPointData().SetScalars(vtk_colors)
     return polydata
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -539,9 +539,9 @@ def get_actor_from_primitive(vertices, triangles, colors=None,
         XYZ coordinates of the object
     triangles: (Nx3) ndarray
         Indices into vertices; forms triangular faces.
-    colors: (Mx3) or (Mx4) ndarray
+    colors: (Nx3) or (Nx4) ndarray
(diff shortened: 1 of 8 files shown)


fury/utils.py
```diff
@@ -539,9 +539,9 @@ def get_actor_from_primitive(vertices, triangles, colors=None,
         XYZ coordinates of the object
     triangles: (Nx3) ndarray
         Indices into vertices; forms triangular faces.
-    colors: (Nx3) ndarray
-        N is equal to the number of lines. Every line is coloured with a
-        different RGB color.
+    colors: (Mx3) or (Mx4) ndarray
```
Garyfallidis, line 542: Nx3 or Nx4

PR #286, opened 2020-08-12
GSoC blogs for Third Evaluation.

Nibba2018:
PR to add blogs written during third evaluation.


skoudoro (commented):
Thanks, I just need to check it after generation and I will merge this PR beginning of next week

PR #288, opened 2020-08-14
Shader manager

### LINKED ISSUE #260: Changes to shader API in VTK 9
When running FURY shader code with VTK 9, I now see warnings like this:

```
vtkOpenGLPolyDataMapper:227   WARN| 
vtkOpenGLPolyDataMapper::AddShaderReplacement was deprecated for VTK 9.0 and will be removed in a future version.  
Use vtkOpenGLShaderProperty::AddShaderReplacement instead.
```

The new implementation comes from [this commit](https://github.com/Kitware/VTK/commit/1b005a294d34f5628dc029fa943047cff9629f22), where the new classes and some examples of their use can be found.

We probably don't need to do anything until:

- FURY moves to a requirement of `vtk>=9.0.0` or
- VTK removes the feature instead of just deprecating it

skoudoro:
Beginning of adding utility functions to standardize and help the shader management. Still some work to do:

- [x] Add tests
- [x] Add callback
- [x] Add attributes

Feel free to comment and add more ideas. 

fix #260

### DIFF
--- a/fury/shaders/base.py
+++ b/fury/shaders/base.py
@@ -21,7 +21,7 @@ SHADERS_BLOCK = {
 
 
 def add_shader_to_actor(actor, shader_type, impl_code="", decl_code="",
-                        block="", keep_default=True, replace_first=True,
+                        block="valuepass", keep_default=True, replace_first=True,
                         replace_all=False, debug=False):
     """Apply your own substitutions to the shader creation process.
 
@@ -41,7 +41,12 @@ def add_shader_to_actor(actor, shader_type, impl_code="", decl_code="",
         shader declaration code, should be a string or filename
         by default None
     block : str, optional
-        section name to be replaced, by default
+        section name to be replaced. vtk use of heavy string replacments to
+        to insert shader and make it flexible. Each section of the shader
+        template have a specific name. For more informations:
+        https://vtk.org/Wiki/Shaders_In_VTK. The possible values are:
+        position, normal, light, tcoord, color, clip, camera, prim_id,
+        valuepass. by default valuepass
     keep_default : bool, optional
         keep the default block tag to let VTK replace it with its default
         behavior. By default True
--- a/fury/shaders/base.py
+++ b/fury/shaders/base.py
@@ -22,8 +22,8 @@ SHADERS_BLOCK = {
 
 
 def add_shader_to_actor(actor, shader_type, impl_code="", decl_code="",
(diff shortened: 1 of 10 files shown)


lenixlobo:
Hi! 
This is very easy to use . 

In documentation, I think we should also explain the shaders blocks in short
As a new user, it can be quite confusing why the blocks exist and vtk doesn’t have much documentation on the same.

lenixlobo:
Hi! 
This is very easy to use . 

In documentation, I think we should also explain the shaders blocks in short
As a new user, it can be quite confusing why the blocks exist and vtk doesn’t have much documentation on the same.

fury/shaders/base.py
```diff
@@ -0,0 +1,136 @@
...
+        raise ValueError(msg)
+
+    block_dec = block + "::Dec"
+    block_impl = block + "::Impl"
+
+    if keep_default:
+        decl_code = block_dec + "\n" + decl_code
+        impl_code = block_impl + "\n" + impl_code
+
+    if debug:
```
guaje, line 79: Nice one! This will be extremely useful.

fury/shaders/base.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Parameters
+    ----------
+    actor : vtkActor
+        Rendered Object
+    arr : ndarray
+        array to link to vertices
+    arr_name : str
+        data array name
+    attr_name : str
```
guaje, line 122: Could we just keep one of the 2 params names in the signature of the function and if we need to differentiate between the 2 names do it by appending a fixed prefix to one of the 2 strings. My suggestion would be to leave the users to define the attribute they have to retrieve in the vertex shader and leave this function to manage the name of the vtk array.

fury/shaders/base.py
Garyfallidis: ;)

fury/shaders/base.py
skoudoro: 👍

skoudoro:
PR ready for review @guaje @Garyfallidis @lenixlobo @dmreagan.

Garyfallidis:
I suggest not using ``add_`` in these functions. If you have add then you need to have remove too.
What about using ``shader_to_actor`` and ``attribute_to_actor``.

Garyfallidis:
Merging. On another PR we will have to update the experimental shaders too.

PR #289, opened 2020-08-18
Rename parameter scale to scales

skoudoro:
Quick PR to rename the parameter `scale` to `scales` before it becomes a habit.


docs/examples/viz_advanced.py
```diff
@@ -264,7 +264,7 @@ def win_callback(obj, _event):
 # Finally, please set the following variable to ``True`` to interact with the
 # datasets in 3D.
 
-interactive = False
+interactive = True
```
Garyfallidis, line 267: Should this be True?

skoudoro: good catch! updated

guaje:
The primitive module still has some scale params in singular: https://github.com/fury-gl/fury/blob/master/fury/primitive.py#L50

skoudoro:
ok,  I will update that, thanks @guaje!

guaje:
> ok, I will update that, thanks @guaje!

Also the superquadric function: https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1864

guaje:
> > ok, I will update that, thanks @guaje!
> 
> Also the superquadric function: https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1864

skoudoro:
ok, all fixed @guaje. thank you for the review. Let me know if you see something else concerning `scales`

guaje:
> ok, all fixed @guaje. thank you for the review. Let me know if you see something else concerning `scales`

Sure, I'm checking again all the files. So far, I've found this:

- https://github.com/fury-gl/fury/pull/289/files#diff-8da821ba4630b32206fafffa55273b6dR1332

guaje:
I can't find any missing usage of the word "scale". Merging.

Issue #274, opened 2020-07-24
SDF rendering bug for low scale values


lenixlobo:
In the SDF actor, when the scale of the actor is set below 0.01 some deformation is observed. As the value is further decreased, the deformation gets worse

lenixlobo:
Example : for no. of actors = 100000, scale = 0.001
![bug1](https://user-images.githubusercontent.com/20311706/88391950-ca5ddb00-cdd8-11ea-8ecb-6ada27356c5b.png)

PR #307, opened 2020-09-15
Fixed translating and scaling issues on billboard and SDF actors

guaje:
This PR fixes:

 - Scaling issue on the billboard actor.
 - Translating issue on the billboard and SDF actors after upgrading to VTK 9.0.1.

|Updated Billboard actor |Updated SDF actor |
|---|---|
|Issue1|Issue2|

### DIFF
--- a/fury/shaders/billboard_dec.frag
+++ b/fury/shaders/billboard_dec.frag
@@ -1,3 +1,2 @@
 /* Billboard  fragment shader declaration */
-in vec3 centeredVertexMC;
-in float scalingFactor;
+in vec3 normVertexMCVSOutput;
--- a/fury/shaders/billboard_dec.frag
+++ b/fury/shaders/billboard_dec.frag
@@ -1,2 +1,2 @@
 /* Billboard  fragment shader declaration */
-in vec3 normVertexMCVSOutput;
+in vec3 normalizedVertexMCVSOutput;
--- a/fury/shaders/billboard_dec.vert
+++ b/fury/shaders/billboard_dec.vert
@@ -1,5 +1,4 @@
 /* Billboard  vertex shader declaration */
 in vec3 center;
 
-out vec3 centeredVertexMC;
-out float scalingFactor;
+out vec3 normVertexMCVSOutput;
--- a/fury/shaders/billboard_dec.vert
+++ b/fury/shaders/billboard_dec.vert
@@ -1,4 +1,4 @@
 /* Billboard  vertex shader declaration */
 in vec3 center;
 
-out vec3 normVertexMCVSOutput;
+out vec3 normalizedVertexMCVSOutput;
--- a/fury/shaders/billboard_dec.vert
+++ b/fury/shaders/billboard_dec.vert
@@ -1,4 +1,5 @@
 /* Billboard  vertex shader declaration */
 in vec3 center;
+
 out vec3 centeredVertexMC;
 out float scalingFactor;
--- a/fury/shaders/billboard_impl.vert
+++ b/fury/shaders/billboard_impl.vert
@@ -1,11 +1,12 @@
 /* Billboard  vertex shader implementation */
-centeredVertexMC = vertexMC.xyz - center;
+centeredVertexMC = vertexMC.xyz - center; // 1st Norm. [-scale, scale]
 scalingFactor = 1. / abs(centeredVertexMC.x);
-centeredVertexMC *= scalingFactor;
-
-vec2 billboardSize = vec2(scalingFactor, scalingFactor);
-
+float size = abs(centeredVertexMC.x) * 2;
+centeredVertexMC *= scalingFactor; // 2nd Norm. [-1, 1]
+vec2 billboardSize = vec2(size, size); // Fixes the scaling issue
 vec3 cameraRightWorldSpace = vec3(MCVCMatrix[0][0], MCVCMatrix[1][0], MCVCMatrix[2][0]);
 vec3 cameraUpWorldSpace = vec3(MCVCMatrix[0][1], MCVCMatrix[1][1], MCVCMatrix[2][1]);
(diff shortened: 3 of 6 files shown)


skoudoro:
That's awesome! Thank you for fixing that. It is working like a charm on my macOS. What a tricky issue!

I will merge this PR later in the day, waiting for Travis CI.

Thank you @guaje

skoudoro (approved):

dmreagan:
Nice catch, @guaje! However, it does not fix #274, as @skoudoro suggests above.

fury/shaders/billboard_dec.vert
```diff
@@ -1,4 +1,5 @@
 /* Billboard  vertex shader declaration */
 in vec3 center;
+
 out vec3 centeredVertexMC;
```
dmreagan, line 4: I'd like to see these follow the VTK naming convention. So this would be `centeredVertexMCVSOutput`.

fury/shaders/billboard_dec.vert
```diff
@@ -1,4 +1,5 @@
...
 out float scalingFactor;
```
dmreagan, line 5: and this would be `scalingFactorVSOutput`

fury/shaders/billboard_impl.vert
```diff
@@ -1,11 +1,12 @@
...
-vec2 billboardSize = vec2(scalingFactor, scalingFactor);
-
+float size = abs(centeredVertexMC.x) * 2;
+centeredVertexMC *= scalingFactor; // 2nd Norm. [-1, 1]
+vec2 billboardSize = vec2(size, size); // Fixes the scaling issue
 vec3 cameraRightWorldSpace = vec3(MCVCMatrix[0][0], MCVCMatrix[1][0], MCVCMatrix[2][0]);
 vec3 cameraUpWorldSpace = vec3(MCVCMatrix[0][1], MCVCMatrix[1][1], MCVCMatrix[2][1]);
-vec3 vertexPositionWorldSpace = center + cameraRightWorldSpace * billboardSize.x * centeredVertexMC.x + cameraUpWorldSpace * billboardSize.y * centeredVertexMC.y;
+vec3 vertexPositionWorldSpace = center +
+cameraRightWorldSpace * billboardSize.x * centeredVertexMC.x +
```
dmreagan, line 10: indent

skoudoro:
It fixes the issue @dmreagan.

A scale of 0.001 is too small to see it but if you try 0.01 for example, it works. Try to decrease the scale gradually and you will see that the effect of this fix

fury/shaders/billboard_impl.vert
```diff
@@ -1,11 +1,12 @@
...
+cameraUpWorldSpace * billboardSize.y * centeredVertexMC.y;
```
dmreagan, line 11: indent

guaje:
> Nice catch, @guaje! However, it does not fix #274, as @skoudoro suggests above.

Yes, that's because @lenixlobo is applying the scaling on the GLSL code. That's why I didn't link this PR to that issue. I just fixed the inverted scaling on the billboard actor, but the translation issue should be fixed for both actors.

fury/tests/test_actors.py
```diff
@@ -1294,9 +1294,12 @@ def test_superquadric_actor(interactive=False):
 def test_billboard_actor(interactive=False):
```
dmreagan, line 1294: This test doesn't appear to test anything

skoudoro:
You see how tiny it is with 0.01
```
    scene = window.Scene()
    scene.background((1, 1, 1))
    centers = np.array([[1, 0, 0], [0, 1, 0], [0, 1, 0]])
    directions = np.array([[0, 1, 0], [1, 0, 0], [0, 0, 1]])
    scales = np.array([1, 2, 3]) * 0.01
    primitive = ['sphere', 'ellipsoid', 'torus']

    sdf_actor = actor.sdf(centers, directions,
                          colors, primitive, scales)
    scene.add(sdf_actor)
    scene.add(actor.axes())
    if interactive:
        window.show(scene)
```
Capture d’écran 2020-09-15 à 18 32 44

fury/shaders/billboard_dec.vert
guaje: I have a question regarding this variable. We are passing this variable to the fragment shader however it is not used in any of the current examples we have. Should we restrict its usage only to the vertex shader code?

fury/tests/test_actors.py
guaje: Thanks for noticing this

fury/shaders/billboard_dec.vert
dmreagan: If it isn't used anywhere, then it is probably safe to stop passing it through. Perhaps it is just left over from some experiment and isn't really needed.

fury/shaders/billboard_dec.frag
```diff
@@ -1,3 +1,2 @@
 /* Billboard  fragment shader declaration */
-in vec3 centeredVertexMC;
-in float scalingFactor;
+in vec3 normVertexMCVSOutput;
```
skoudoro, line 2: Not sure that `norm` is a good name, We can easily confuse it with normal. I personally preferred `centered`

skoudoro (commented):
Any other comments? @dmreagan @lenixlobo?

Should we go ahead and merge this PR?

fury/shaders/billboard_dec.frag
guaje: Will `normalized` be a better choice? I think `centered` is not an accurate description of what that variable is supposed to be.

skoudoro: ok, it seems we have no choice to put the whole word (`normalized`)

skoudoro:
Any other comments? I will wait until tomorrow and then merge this PR.

skoudoro:
Ok, I'm going ahead and merge. Thank you @guaje

PR #330, opened 2020-10-06
WIP: Swarming/flocking simulation based on simple boids rules

nasimanousheh:
I added a simple example, a new file swarm.py and a test file. There is still some work to do and I need some help with making the rendering faster. Most time is spent for rotating the cones because of too many python calls.


Garyfallidis:
@guaje, @lenixlobo, @skoudoro, @Nibba2018 we need vertex shader version of vec2vec_rotmat function used here. A similar version can be used to solve the issues with slow rotations for bullet integration that @Nibba2018 had.

I have done something similar here https://github.com/fury-gl/fury/blob/master/docs/experimental/viz_shader_square.py#L68 and @skoudoro may have a better version of that using the new shader manager.

skoudoro:
Hi @nassim, Thank you for this awesome example

> There is still some work to do

Can you give us more details. Just to know when we can start to review.

> Most time is spent for rotating the cones

ok, we will look into this part.

guaje:
> @guaje, @lenixlobo, @skoudoro, @Nibba2018 we need vertex shader version of vec2vec_rotmat function used here. A similar version can be used to solve the issues with slow rotations for bullet integration that @Nibba2018 had.
 
@Garyfallidis and @nasimanousheh please review [this](https://github.com/nasimanousheh/fury/pull/2) PR addressing this request.

Issue #317, opened 2020-09-30
Track and fix warnings during tests. 


skoudoro:
When you run FURY tests, a bunch of warnings is popping up. The goal of this task is to track and fix them. Some of them are expected, others not. We will need many PR for this issue and more precisely, 1PR per module

TODO:
-------

- [x] fury/actor.py  (`pytest -svv fury/test_actor.py`)
- [x] fury/colormap.py  (`pytest -svv fury/test_colormap.py`)
- [x] fury/convert.py  (`pytest -svv fury/test_convert.py`)
- [x] fury/decorators.py  (`pytest -svv fury/test_decorators.py`)
- [x] fury/interactor.py  (`pytest -svv fury/test_interactor.py`)
- [x] fury/io.py  (`pytest -svv fury/test_io.py`)
- [ ] fury/layout.py  (`pytest -svv fury/test_layout.py`)
- [x] fury/ui.py  (`pytest -svv fury/test_ui.py`)
- [x] fury/pick.py  (`pytest -svv fury/test_pick.py`)
- [x] fury/primitive.py  (`pytest -svv fury/test_primitive.py`)
- [x] fury/transform.py  (`pytest -svv fury/test_transform.py`)
- [x] fury/window.py  (`pytest -svv fury/test_window.py`)
- [x] fury/utils.py  (`pytest -svv fury/test_utils.py`)

tushar5526:
```
interactive = False

    def test_superquadric_actor(interactive=False):
        scene = window.Scene()
        centers = np.array([[8, 0, 0], [0, 8, 0], [0, 0, 0]])
        colors = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
        directions = np.random.rand(3, 3)
        scales = [1, 2, 3]
        roundness = np.array([[1, 1], [1, 2], [2, 1]])
    
        sq_actor = actor.superquadric(centers, roundness=roundness,
                                      directions=directions,
                                      colors=colors.astype(np.uint8),
                                      scales=scales)
        scene.add(sq_actor)
        if interactive:
            window.show(scene)
    
        arr = window.snapshot(scene, offscreen=True)
        arr[arr > 0] = 255  # Normalization
        report = window.analyze_snapshot(arr, colors=255*colors.astype(np.uint8))
>       npt.assert_equal(report.objects, 3)
E       AssertionError: 
E       Items are not equal:
E        ACTUAL: 6
E        DESIRED: 3

[...]

PR #332, opened 2020-10-22
Fixing UI warnings

Nibba2018:
Context: #317 

The following warning:

`DeprecationWarning: Converting 'np.character' to a dtype is deprecated. The current result is 'np.dtype(np.str_)' which is not strictly correct. Note that 'np.character' is generally deprecated and 'S1' should be used.`

is generated when numpy arrays are converted to vtk arrays in `vtk.util.numpy_support.get_vtk_array_type`. This method is mainly used by the methods in `fury.utils` and `fury.actors`. This warning is generated when arrays contain character values in them, but while test I was not able to find any character values present in the array when the tests are run.

This warning is also the most recurring warning compared to the rest and also the only one in `fury.ui`. I created a file `pytest.ini` to ignore this particular warning and got the following results:

* All Fury test warnings reduced from `892` to `14`
* All Fury UI test warnings reduced from `108` to `0`

I am not sure if there is a better way to fix these.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore: Converting `np.character` to a dtype is deprecated. :DeprecationWarning


skoudoro (approved):
LGTM, thanks @Nibba2018!  I do not know why Azure pipeline failed so I just restarted them. but your PR is ready to go

Issue #309, opened 2020-09-15
Domino Physics Simulation


Nibba2018:
Add a Physics Simulation demo to showcase the following Domino effect:
![domino](https://i.imgur.com/aYhJRSX.gif)

To know more about Physics simulations refer to this [document](https://docs.google.com/document/d/1XJcG1TL5ZRJZDyi8V76leYZt_maxGp0kOB7OZIxKsTA/edit?usp=sharing).

SanjayMarreddi:
Hey @Nibba2018 ,

I would like to work on this issue. Could you please assign this to me and brief it out as well ! 

Thanks and Regards,
Sanjay Marreddi.

Nibba2018:
Hello @SanjayMarreddi , this issue is reserved for Hacktoberfest so if you are interested kindly reclaim it after 1st October.

SanjayMarreddi:
@Nibba2018 
Yeah I am Interested in Hacktoberfest. I will reclaim after 1st October.
Thanks for your Information.

SanjayMarreddi:
Hey @Nibba2018 
Could you please assign this to me now !

Nibba2018:
@SanjayMarreddi Sure! go Ahead

Nibba2018:
@SanjayMarreddi Any updates?

SanjayMarreddi:
@Nibba2018  I was too busy with Academic Exams and Other Events till Tomorrow..
I have done implementing this about 60-70%. I just need to make sure that Bricks follow the Pattern. I will try to complete this before the end of this October Month. 
Hope you Understand !
Thanks !!

Nibba2018:
@SanjayMarreddi Okay no problem :+1:

PR #334, opened 2020-10-29
Added Domino-Simulation-file for Review

SanjayMarreddi:
#309

### DIFF
--- a/docs/examples/physics_using_pybullet/viz_domino.py
+++ b/docs/examples/physics_using_pybullet/viz_domino.py
@@ -4,7 +4,8 @@ Domino Physics Simulation
 =====================
 
 This example simulation shows how to use pybullet to render physics simulations
-in fury. In this example we specifically render a series of Dominoes which are under Domino Effect.
+in fury. In this example we specifically render a series of Dominoes which are
+under Domino Effect.
 
 """
 ###############################################################################
@@ -14,7 +15,8 @@ from fury import window, actor, ui, utils
 import itertools
 import pybullet as p
 
-# Next, we initialize a pybullet client to render the physics. We use `DIRECT` mode to initialize pybullet without a GUI.
+# Next, we initialize a pybullet client to render the physics. 
+# We use `DIRECT` mode to initialize pybullet without a GUI.
 client = p.connect(p.DIRECT)
 
 # Apply gravity to the scene.
@@ -45,10 +47,11 @@ base = p.createMultiBody(
 p.changeDynamics(base, -1, lateralFriction=1, restitution=0.5)
 
 ###############################################################################
-# We define some global parameters of the Dominoes so that its easier for us to tweak the simulation.
+# We define some global parameters of the Dominoes so that its easier for 
+# us to tweak the simulation.
 
 domino_mass = 0.5
-domino_size = np.array([0.1,1,2])
+domino_size = np.array([0.1, 1, 2])
 
 number_of_dominoes = 10
 domino_centers = np.zeros((number_of_dominoes, 3))
@@ -65,7 +68,7 @@ domino_sizes[:] = domino_size
 domino_colors = np.random.rand(number_of_dominoes, 3)
 
 domino_coll = p.createCollisionShape(p.GEOM_BOX,
-                                    halfExtents=domino_size / 2)
+                                     halfExtents=domino_size / 2)
 
 # We use this array to store the reference of domino objects in pybullet world.
 dominos = np.zeros(number_of_dominoes, dtype=np.int8)
@@ -74,20 +77,20 @@ centers_list = np.zeros((number_of_dominoes, 3))
 
 # Adding the dominoes
 for i in range(number_of_dominoes):
-    center_pos = np.array([(i*0.99)-5.5,0.4,1])
+    center_pos = np.array([(i*0.99)-5.5, 0.4, 1])
     domino_centers[i] = center_pos
-    domino_orns[i] = np.array([0, 0, 0, 1]) 
+    domino_orns[i] = np.array([0, 0, 0, 1])
     dominos[i] = p.createMultiBody(baseMass=domino_mass,
-                                      baseCollisionShapeIndex=domino_coll,
-                                      basePosition=center_pos,
-                                      baseOrientation=domino_orns[i])
+                                   baseCollisionShapeIndex=domino_coll,
+                                   basePosition=center_pos,
+                                   baseOrientation=domino_orns[i])
     p.changeDynamics(dominos[i], -1, lateralFriction=0.2, restitution=0.1)
 
 
 domino_actor = actor.box(centers=domino_centers,
-                        directions=domino_directions,
-                        scales=domino_sizes,
-                        colors=domino_colors)
+                         directions=domino_directions,
+                         scales=domino_sizes,
+                         colors=domino_colors)
 
 ###############################################################################
 # Now, we define a scene and add actors to it.
@@ -135,6 +138,7 @@ sec = np.int(num_vertices / num_objects)
 #   * Reshape it in a 3x3 matrix.
 # * Perform calculations to get the required position and orientation.
 # * Update the position and orientation.
+
 
 def sync_domino(object_index, multibody):
     pos, orn = p.getBasePositionAndOrientation(multibody)
@@ -187,13 +191,13 @@ def timer_callback(_obj, _event):
 
     # Get the position and orientation of the first domino.
     domino1_pos, domino1_orn = p.getBasePositionAndOrientation(dominos[0])
-    
+
     # Apply force on the First Domino (domino) above the Center of Mass.
     if apply_force:
         # Apply the force.
         p.applyExternalForce(dominos[0], -1,
                              forceObj=[100, 0, 0],
-                             posObj=domino1_pos + np.array([0,0,1.7]),
+                             posObj=domino1_pos + np.array([0, 0, 1.7]),
                              flags=p.WORLD_FRAME)
         apply_force = False
 
--- a/docs/examples/physics_using_pybullet/viz_domino.py
+++ b/docs/examples/physics_using_pybullet/viz_domino.py
@@ -1,26 +1,33 @@
+"""
+=====================
+Domino Physics Simulation
+=====================
+
+This example simulation shows how to use pybullet to render physics simulations
+in fury. In this example we specifically render a series of Dominoes which are under Domino Effect.
+
+"""
+###############################################################################
 # The following imports are necessary for physics simulations
 import numpy as np
 from fury import window, actor, ui, utils
 import itertools
 import pybullet as p
 
-# Instantiate Pybullet client.
+# Next, we initialize a pybullet client to render the physics. We use `DIRECT` mode to initialize pybullet without a GUI.
 client = p.connect(p.DIRECT)
 
 # Apply gravity to the scene.
-gravity_x = 0
-gravity_y = 0
-gravity_z = -10
-p.setGravity(gravity_x, gravity_y, gravity_z, physicsClientId=client)
+p.setGravity(0, 0, -10, physicsClientId=client)
 
+###############################################################################
 # Base Plane Parameters
 base_size = np.array([15, 15, 0.2])
 base_color = np.array([1, 1, 1])
 base_position = np.array([0, 0, -0.1])
 base_orientation = np.array([0, 0, 0, 1])
 
-
-# Creating BASE Plane
+# Render a BASE plane to support the Dominoes.
 base_actor = actor.box(centers=np.array([[0, 0, 0]]),
                        directions=[0, 0, 0],
                        scales=base_size,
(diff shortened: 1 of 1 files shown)


SanjayMarreddi:
@Nibba2018 Could you please give a Feedback on why It is giving Wrong Output.
Thanks !

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,238 @@
...
+# Adding the dominoes
+for i in range(n_dominoes):
+    center_pos = np.array([(i*0.99)-2.8,1,0])
+    brick_centers[i] = center_pos
+    brick_orns[i] = np.array([0, 0, 0, 1])
+    bricks[i] = p.createMultiBody(baseMass=brick_mass,
+                                      baseCollisionShapeIndex=brick_coll,
+                                      basePosition=center_pos,
+                                      baseOrientation=brick_orns[i])
+    p.changeDynamics(bricks[i], -1, lateralFriction=0.1, restitution=0.1)
```
Nibba2018, line 79: Your domino position calculation is a bit wrong. Your dominoes are being rendered inside the BASE plate and that why you can observe a weird bump during the start of the simulation.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,238 @@
+# The following imports are necessary for physics simulations
+import numpy as np
+from fury import window, actor, ui, utils
+import itertools
+import pybullet as p
+
+# Instantiate Pybullet client.
+client = p.connect(p.DIRECT)
+
```
Nibba2018, line 9: If you are not sure whats going on then you can do `client = p.connect(p.GUI)` to understand render in pybullet world.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,238 @@
...
+# A simpler but inaccurate approach is used here to update the position and
+# orientation.
+
+
+def sync_actor(actor, multibody):
+    pos, orn = p.getBasePositionAndOrientation(multibody)
+    actor.SetPosition(*pos)
+    orn_deg = np.degrees(p.getEulerFromQuaternion(orn))
+    actor.SetOrientation(*orn_deg)
+
```
Nibba2018, line 163: This method is to update the position and orientation of the entire actor. This is actually not necessary. You can render the first domino without any issues in `sync_bricks`.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,238 @@
...
+    if apply_force:
+        # Apply the force.
+        p.applyExternalForce(bricks[0], -1,
+                             forceObj=[5, 0, 0],
+                             posObj=brick1_pos,
+                             flags=p.WORLD_FRAME)
+        apply_force = False
+
+    # Set position and orientation of the first Brick.
+    sync_actor(brick_actor, bricks[0])
```
Nibba2018, line 211: All your dominoes are rendered by this single actor called `brick_actor`. So whatever modifications you make to this actor will be reflected on all the dominoes.
Thus, in this case when you pass `brick_actor`, `bricks` to `sync_actor` you are basically asking fury to modify all the dominoes with the position and orientation of the first domino, which is incorrect.

That's why you can observe that all of your dominoes are following the same trajectory.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,238 @@
...
+        p.getMatrixFromQuaternion(
+            p.getDifferenceQuaternion(orn, brick_orns[object_index])),
+        (3, 3))
+
+    vertices[object_index * sec: object_index * sec + sec] = \
+        (vertices[object_index * sec: object_index * sec + sec] -
+         brick_centers[object_index]) @ rot_mat + pos
+
+    brick_centers[object_index] = pos
+    brick_orns[object_index] = orn
```
Nibba2018, line 150: This method is responsible to update the position and orientation of the objects. So if according to physics the domino is supposed to be stationary then it will be stationary. So you can safely pass the first domino here.

Nibba2018 (commented):
@SanjayMarreddi I have added some comments below.

docs/examples/physics_using_pybullet/viz_domino.py
SanjayMarreddi: @Nibba2018 Yeah This is the place I am facing problem.. Now I have changed the Position and There is no bump in the start of the Simulation. **But I have a doubt why are the Domino's going below below the Base during Simulation?** I have kept center of the Base as `(0,0,0)`. Since Size of the bricks is `np.array([2,0.8, 0.2])` I have kept the y coordinate of the Bricks as 0.4 now like this -
```
for i in range(n_dominoes):
    center_pos = np.array([(i*0.99)-3.3,0.4,1.5])
```

**So Now when I am running the Code, Initially the bricks are lying exactly with their bottom ends on the base, But when Simulation is proceeding, The bottom ends of Bricks are going below the base !!!**

Please check the file now(I made commit in the same PR) and help me resolve this 

Thanks !

Nibba2018: image
@SanjayMarreddi, this is how your dominoes are being initialised in pybullet.

Nibba2018: I think you are getting your orientation wrong.

SanjayMarreddi: @Nibba2018 
I have corrected the Orientation Problem. **Now The Simulation is working Correctly in Pybullet World (GUI option) !**
But in the FURY World, The Dominoes are coming in Horizontal position because of which the Simulation in FURY world is not completely Correct. Where as in the Pybullet world they are fine and in Vertical Position.
I have committed my changes in the same PR.
Could you please tell me how to approach in resolving this !!

Thanks !

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,224 @@
...
+centers_list = np.zeros((n_dominoes, 3))
+
+# Adding the dominoes
+for i in range(n_dominoes):
+    center_pos = np.array([(i*0.99)-5,0.4,1])
+    brick_centers[i] = center_pos
+    if i == 0 : 
+        brick_orns[0] = np.array([-0.6 , -0.6, 1, 1])
+    else:
+        brick_orns[i] = np.array([1 ,1, 1, 1])
```
Nibba2018, line 76: @SanjayMarreddi This is the part of the code that is ruining the positioning of the bricks. These values are in `quaternion`. Remove this part and the simulation should work fine.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,224 @@
...
+    # Get the position and orientation of the first brick.
+    brick1_pos, brick1_orn = p.getBasePositionAndOrientation(bricks[0])
+    
+    # Apply force for the first step of simulation.
+    if apply_force:
+        # Apply the force.
+        p.applyExternalForce(bricks[0], -1,
+                             forceObj=[400, 0, 0],
+                             posObj=brick1_pos,
+                             flags=p.WORLD_FRAME)
```
Nibba2018, line 196: Here, you are applying the force at the center of mass of the domino so it will never rotate in this situation. You need to apply the force above the center of mass.

Nibba2018:
Peek 2020-11-01 23-48

docs/examples/physics_using_pybullet/viz_domino.py
SanjayMarreddi: @Nibba2018 
I kept the Orientation of the First Brick in a Slanted Position willingly. So that the Brick will yield to Domino effect even though we apply Force at COM. (We can even change this scenario by keeping the Brick Parallel only and apply force above COM as you said)

**But my Doubt is the Following:-
Currently, With the code (with out removing above 73-76 lines), The Simulation is working Correctly in the PyBullet World(GUI). But in the FURY World, The Dominoes are coming in "Horizontal position" because of which the Simulation in FURY world is not completely Correct.** 

Please find the Images of the Same:-

1) In the Pybullet :
Screenshot (191)

2) In the Fury:
Screenshot (192)

Nibba2018: @SanjayMarreddi In these lines you are explicitly assigning values to `brick_orns` which is only used by pybullet and not by fury. If you need to make changes to the fury world then you have to update `brick_directions` as well. Now as you are only updating `brick_orns` instead of updating both, it means that your simulation in fury will have a negative offset(in orientation) compared to the one in pybullet.

Just remove those line and you will notice that pybullet and fury are both in sync. Pybullet and Fury uses different methods to represent orientations and we haven't figured out a way to map those together. So, we let `sync_bricks` handle the orientation situation during runtime.

SanjayMarreddi: @Nibba2018 
I understood your explanation. But the problem is If I remove those 4 lines, I could not see any thing in both the FURY and PYBULLET WORLD !

So I am trying to change `brick_directions`

Nibba2018: @SanjayMarreddi right for that you need to set `brick_directions[:] = np.array([1.57, 0, 0])` and `brick_orns[i] = np.array([0, 0, 0, 1])`

SanjayMarreddi: @Nibba2018  Basically I started from this only (In the first Commit of PR).
Since it is Initializing the things in a Wrong way, I started changing the Orientation to   
```
brick_orns[i] = np.array([1 ,1, 1, 1])
``` 
to make the Bricks Vertical in Pybullet (GUI).

In order to apply force on the COM on 1st Brick, I tilted the Orientation of 1st Brick and finally changed to -

```    
if i == 0 : 
        brick_orns[0] = np.array([-0.6 , -0.6, 1, 1])
    else:
        brick_orns[i] = np.array([1 ,1, 1, 1])
```
So Now it is working Fine in GUI

and then I tried Multiple Directions for FURY world to fix orientation problem in FURY like -
```
brick_directions[:] = np.array([1.57, 0, 0])
brick_directions[:] = np.array([0, 1, 1]) # This is giving better result
```

The result corresponding to `brick_directions[:] = np.array([0, 1, 1])` is as follows:-

Screenshot (197)

**The only thing left is to make this similar to the Pybullet by changing `brick_directions[:] = np.array([0, 1, 1])` to a better value. I have been trying soo hard for this.. Please suggest the direction.**

Thanks!

Nibba2018: @SanjayMarreddi Is it really necessary to tilt the domino? I mean you are already applying a force in `timer_callback`. Trying to tilt it will complicate things. I have made a PR on your repo kindly check.

SanjayMarreddi: @Nibba2018 

Okay Got it.. Thanks .!

So finally, The entire Problem between the Syncing of FURY and PYBULLET occurred because of Including the Tilt in the First domino. So to rectify this we kept it parallel only and applied force at a point above COM..

SanjayMarreddi:
@Nibba2018 
I have made all the Final Changes. Could you please review it.

Also Could you tell me more about `TravisCI - Pull Request` automated Test.
I am failing in that test. To make my PR merged, I need to pass all the tests I think?

Thanks !

skoudoro:
Good progress here!  do not forget to address all pep8 issues reported here[https://github.com/fury-gl/fury/pull/334#issuecomment-718590194] @SanjayMarreddi.

Also, you will need to document and explain a bit more your tutorial. It is missing a title and explanation. Please, look at the other tutorials to get an idea of the format.

No worries about Travis, the error does not seem to be related to your PR. I will test it a bit later.

SanjayMarreddi:
@skoudoro 
Sure ! I will address all the pep8 issues Reported.
Also I will document and explain the Tutorial completely by Tomorrow night, Hope that is fine.

Thanks !

Nibba2018:
@SanjayMarreddi  Also make sure to change your variable names from `brick_*` to `domino_*`.

SanjayMarreddi:
@Nibba2018 Sure I will do that as well..

SanjayMarreddi:
@Nibba2018 @skoudoro 
Please review the PR Now.
I resolved all the PEP Issues, Added the Title , Improved Documentation and Explanation of Tutorial, Changed the Variable Names too..

Thanks !

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,228 @@
+"""
+=====================
+Domino Physics Simulation
+=====================
```
Nibba2018, line 4: The `====` must cover the entire text.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,228 @@
...
+
+This example simulation shows how to use pybullet to render physics simulations
+in fury. In this example we specifically render a series of Dominoes which are
+under Domino Effect.
+
+"""
+###############################################################################
```
Nibba2018, line 11: I don't think this line is necessary :thinking: . As the doc-string itself will be considered as a different section.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,228 @@
...
+# Render a BASE plane to support the Dominoes.
+base_actor = actor.box(centers=np.array([[0, 0, 0]]),
+                       directions=[0, 0, 0],
+                       scales=base_size,
+                       colors=base_color)
+
+base_coll = p.createCollisionShape(p.GEOM_BOX,
+                                   halfExtents=base_size / 2)
+
+# half of the actual size.
```
Nibba2018, line 41: This comment should be above line 38.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,228 @@
...
+    for idx, domino in enumerate(dominos):
+        sync_domino(idx, domino)
+    utils.update_actor(domino_actor)
+
+    # Simulate a step.
+    p.stepSimulation()
+
+    # Exit after 2000 steps of simulation.
+    if cnt == 2000:
+        showm.exit()
```
Nibba2018, line 214: Reduce `2000` to `300` steps.

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,228 @@
...
+showm.add_timer_callback(True, 1, timer_callback)
+
+interactive = False
+
+# start simulation
+if interactive:
+    showm.start()
+
+window.record(scene, out_path="domnio_simulation.png", size=(900, 768))
+###############################################################################
```
Nibba2018, line 228: The `#####` line isn't necessary. also set `out_path="viz_domino.png"` to make it same as the other examples.

Nibba2018 (commented):
Great Job! @SanjayMarreddi . The simulation was very well made. I have added some comments below.

SanjayMarreddi:
@Nibba2018 
Made the Above Modifications. Please Review it !

Thanks!

Nibba2018 (approved):
LGTM! :rocket: . I would like to request @skoudoro to have a look.

skoudoro:
Thank you for the work @SanjayMarreddi  and thanks for the review @Nibba2018!

I will look into it by the end of the week and keep you updated

SanjayMarreddi:
@skoudoro Kindly look into this PR & provide feedback.. Will be waiting for your update. Thanks !!

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,226 @@
...
+p.changeDynamics(base, -1, lateralFriction=1, restitution=0.5)
+
+###############################################################################
+# We define some global parameters of the Dominoes so that its easier for
+# us to tweak the simulation.
+
+domino_mass = 0.5
+domino_size = np.array([0.1, 1, 2])
+
+number_of_dominoes = 10
```
skoudoro, line 55: If I change the number of dominoes,  it will be good if you could update the size of the base. Otherwise, it looks strange

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+# Calculate the vertices of the dominos.
+vertices = utils.vertices_from_actor(domino_actor)
+num_vertices = vertices.shape[0]
+num_objects = domino_centers.shape[0]
+sec = np.int(num_vertices / num_objects)
+
+###############################################################################
+# ==============
```
skoudoro, line 128: `===` should go until the end of `Dominoes`, otherwise there is a error on doc generation

docs/examples/physics_using_pybullet/viz_domino.py
```diff
@@ -0,0 +1,226 @@
...
+# Syncing Dominoes
+# ==============
```
skoudoro, line 130: same comment as above

skoudoro (commented):
Hi @SanjayMarreddi,

Overall, it looks good to me, Nice job! See below for some comments but I think it is ready to go soon. Sorry for the late review.

docs/examples/physics_using_pybullet/viz_domino.py
skoudoro: you can check the simulation when you change the value to 20.

SanjayMarreddi:
@skoudoro 
I made all the above Modifications suggested by you. Could you please re-review it now ?
Thanks !

skoudoro:
thank you for the update @SanjayMarreddi. This first version looks good. merging

PR #340, opened 2020-11-21
Pass OAuth token through header

### LINKED ISSUE #315: Deprecation notice for authentication via URL query parameters
An update is needed on [`github_tools.py`](https://github.com/fury-gl/fury/blob/master/docs/source/ext/github_tools.py) script.

We need to use the Authorization HTTP header instead as using the `access_token` query parameter is deprecated.

More information [here](https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param)

tushar5526:
Closes #315

### DIFF
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -19,7 +19,7 @@ import operator
 from datetime import datetime, timedelta
 from subprocess import check_output
 
-from urllib.request import urlopen
+from urllib.request import urlopen, Request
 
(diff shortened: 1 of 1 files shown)


skoudoro (commented):
Hi @tushar5526,

Overall, It looks good to me.  I just need to test this locally and then we can merge it. 

Thanks @tushar5526!

skoudoro:
Unfortunately, I can not test it right now. So I will go ahead and merge it. Thank you @tushar5526.

PR #342, opened 2020-11-22
From Travis to Github Actions 

### LINKED ISSUE #338: From travis (pricing model changed) to github Actions ?
The pricing model of Travis changed. Our macOS matrix stopped to work if we do not pay. here more informations: https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing.

We need to read this carefully all the changes and we need to remove/replace our macOS matrix

> We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment. Please include:

> Your account name and VCS provider How many credits (build minutes) you’d like to request (should your run out of credits again you can repeat the process to request more or discuss a renewable amount)

Also, this is not good sign

skoudoro:
This PR should fix #338. The goal is to switch off Travis and start Github Actions.


skoudoro:
- I need to squash a bit multiple commits.
- Xvfb works 🥳 
- There is a strange version of FURY, as we can see [here](https://github.com/fury-gl/fury/pull/342/checks?check_run_id=1446179815#step:10:340) that's why CI's are failing. I do not understand where this strange version number of fury comes from.

arokem:
Just a hunch: you might want to change your working directory after installing, and before running the tests, so that Python picks up the version from the installation and not from the source.

skoudoro:
This PR is done, anyone wants to have a look?

@arokem, my version problem was due to this [2 lines](https://github.com/fury-gl/fury/pull/342/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R68-R69). Only the commit was pulled, but with `versionner` you need the tag to have the right version. These 2 lines permit to fetch the whole repo

Garyfallidis:
Done! Thx!

PR #348, opened 2020-12-09
Added center forwarding in billboard shaders.

guaje:
This simple PR forwards the content of the variable center from the vertex shader to the fragment shader.


skoudoro:
Thanks @guaje!

PR #349, opened 2020-12-25
Implemented Saving frames to GIF and tested on a Tutorial

### LINKED ISSUE #323: GIF support
Add support for saving frames as gif. Make sure the color and quality is saved correctly.

SanjayMarreddi:
Closes #323  

I have Implemented Saving frames to GIF Image and Tested on this tutorial :  [Using a timer](https://fury.gl/latest/auto_tutorials/01_introductory/viz_timers.html#sphx-glr-auto-tutorials-01-introductory-viz-timers-py) 

Here is the GIF generated by my Code : 
Bubbles

@skoudoro Could you please look into it.
Also, I think I need to generalize it right ? I feel Incorporating it in the `window.py` would be better & using it like 
```
window.save_as_gif(path, some other args)
```
With my current way of converting into GIF, while Generalizing it, we need to pass more arguments to `save_as_gif` and even we need to restrict user to write the code in a specific way. Could you please give me an Idea for generalising it ?

Thanks !!

### DIFF
--- a/fury/tests/test_gif.py
+++ b/fury/tests/test_gif.py
@@ -1,0 +1,87 @@
+"""
+In this File, I am testing the way of Saving Frames as GIF on the Tutorial named `Using a timer` 
+"""
+
+import os
+import itertools
+import numpy as np
+from PIL import Image, ImageDraw
+from fury import window, actor, ui
+
+
+xyz = 10 * np.random.rand(100, 3)
+colors = np.random.rand(100, 4)
+radii = np.random.rand(100) + 0.5
+
+
+os.mkdir("temp")
+
+
+# This function does the Execution of the Code with the Threshold for the Counter provided.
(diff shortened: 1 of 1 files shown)


skoudoro (commented):
Hi @SanjayMarreddi,

Thank you for looking into this.

- It will be good to save the gif without using the disk (save on the disk).
- `save_as_gif` function should be on `fury.io` module.
- `save_as_gif` should have as parameter `fname` and a 3D or 4D volume which are the time frame.

I think you are close to a reasonable solution. Thanks!

SanjayMarreddi:
@skoudoro Looks like someone approached with a better Idea (#358). I really appreciate it. Should I close this PR ? Thanks !!

ShivrajSS-git:
is this still open?

PR #351, opened 2021-01-01
Opacity bug fix for point and sphere actors

### LINKED ISSUE #335: _opacity argument for point doesn't seem to work
## Description
The "_opacity" argument for fury.actor.point doesn't seem to work. Manipulating the values for the _opacity doesn't seem to change the opacity of the rendered points.
From what I could figure from the documentation, this occurs as the function which generates the points (fury.actor.point) utilises the sphere function (fury.actor.sphere) in generating the points and the function for generating spheres doesn't have any means to manipulate the opacity of the spheres that it generates.

## Way to reproduce
- [x] Code example
```
import numpy as np

scene = window.Scene()
coor1 = np.array([[0, 0, 0]])
coor2 = np.array([[1, 0, 0]])
point_actor1 = actor.point(coor1, window.colors.coral, _opacity=0.00, point_radius=0.2)          #point on the left
point_actor2 = actor.point(coor2, window.colors.coral, _opacity=1.00, point_radius=0.2)          #point on the right
scene.add(point_actor1)
scene.add(point_actor2)
window.show(scene)
window.record(scene, size=(300, 300), out_path="vr.png")

```

From the code, it's evident that the desired output requires that the point on the left be transparent (opacity=0) and the point on the right be opaque. But the output rendered has both the points as opaque.
- [x] Relevant images (if any)

vr

- [x] Operating system and versions 
{'fury_version': '0.6.1', 'pkg_path': "C:\\Users\\Tzu's Pc\\AppData\\Local\\Programs\\Python\\Python38\\lib\\site-packages\\fury", 'commit_hash': 'ee4e85c8f1358105700c156b20ba1db83a004261', 'sys_version': '3.8.6 (tags/v3.8.6:db45529, Sep 23 2020, 15:52:53) [MSC v.1927 64 bit (AMD64)]', 'sys_executable': "C:\\Users\\Tzu's Pc\\AppData\\Local\\Programs\\Python\\Python38\\python.exe", 'sys_platform': 'win32', 'numpy_version': '1.19.2', 'scipy_version': '1.5.3', 'vtk_version': '9.0.1', 'matplotlib_version': '3.3.2'}

SunTzunami:
Fixes #335 

Earlier, 
vr

Now, 
vr2

Earlier, opacity argument wasn't manipulating the opacity of point actors and sphere actors. That issue is resolved by this fix. Unit tests are included for the same.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1386,7 +1386,7 @@ def dots(points, color=(1, 0, 0), opacity=1, dot_size=5):
     return aPolyVertexActor
 
 
-def point(points, colors, _opacity=1., point_radius=0.1, theta=8, phi=8):
+def point(points, colors, point_radius=0.1, phi=8, theta=8, opacity=1.):
     """Visualize points as sphere glyphs
 
     Parameters
@@ -1394,10 +1394,10 @@ def point(points, colors, _opacity=1., point_radius=0.1, theta=8, phi=8):
     points : ndarray, shape (N, 3)
     colors : ndarray (N,3) or tuple (3,)
     point_radius : float
-    theta : int
     phi : int
+    theta : int
     opacity : float, optional
-        Takes values from 0 (fully transparent) to 1 (opaque)
+        Takes values from 0 (fully transparent) to 1 (opaque). Default is 1.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


skoudoro (approved):
Hi @SunTzunami,

It looks good to me. I just need to try it locally and then I will go ahead and merge your PR.

Thank you for this update

skoudoro:
After looking more in detail into your fix, I think we should avoid using `actor.GetProperty().SetOpacity(opacity)`. It seems not compatible with our architecture. 
Ideally, in the function `actor.sphere`, we would check if `opacity != 1`.  if yes, we should update the color array by adding a dimension if it is an Nx3, or by updating the last dimension if it is an Nx4.

I will merge your PR because it fixes the `actor.point` but we need a new PR to improve this version.

skoudoro:
thank you @SunTzunami

SunTzunami:
Hi @skoudoro. To make these fixes, I read the documentation for line actor ([https://fury.gl/latest/_modules/fury/actor.html#line](url)) as transparency worked for line actor. Line actor too uses `actor.GetProperty().SetOpacity(opacity)` to manipulate opacity so I assumed it'll be fine to use that for sphere and point actors too.

skoudoro:
In both cases, it needs to be improved @SunTzunami. 

if you look [here](https://github.com/fury-gl/fury/blob/master/fury/tests/test_actors.py#L890), the 3 spheres have 3 differents opacities : `0.3, 04, 0.99`. This is how we setup the opacities. `actor.GetProperty().SetOpacity(opacity)`  do not update this array and override our values.  Each object can have a different opacity which is impossible with `actor.GetProperty().SetOpacity(opacity)`.

SunTzunami:
@skoudoro thanks for the info. I'll try making a better fix.

PR #357, opened 2021-01-11
Patches: vulnerable code that can lead to RCE

Aju100:
Hi, there in this new PR, I have removed the vulnerable functions call used in the code which can allow leading Remote Code execution. Instead of using urrlib, we can use Requests for it which can solve the issue for it. For more info, you can look through [bandit official docs](https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html#b310-urllib-urlopen) and [StackOverflow.](https://stackoverflow.com/questions/48779202/audit-url-open-for-permitted-schemes-allowing-use-of-file-or-custom-schemes)

But one thing is we have to harder URL for there in order to request it. Thank you

### DIFF
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -19,7 +19,7 @@ import operator
 from datetime import datetime, timedelta
 from subprocess import check_output
 
-from urllib.request import urlopen, Request
+from urllib.request import Request
 
 # ----------------------------------------------------------------------------
 # Globals
@@ -40,6 +40,7 @@ GH_TOKEN = os.environ.get('GH_TOKEN', '')
 # Functions
 # ----------------------------------------------------------------------------
 
+req = urllib.request.Request('http://www.url.com')
 
 def fetch_url(url):
     req = Request(url)
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -11,7 +11,6 @@ from shutil import copyfileobj
 import tarfile
 import zipfile
 
-#from urllib.request import urlopen
 import urllib.request
 
 # Set a user-writeable file-system location to put files:
@@ -128,6 +127,8 @@ def check_sha(filename, stored_sha256=None):
             raise FetcherError(msg)
 
 url = urllib.request.Request('http://www.example.com')
+
+
 def _get_file_data(fname, url):
     with contextlib.closing(urlopen(url)) as opener:
         try:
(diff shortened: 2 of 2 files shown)


docs/source/ext/github_tools.py
```diff
@@ -40,6 +40,7 @@
 # Functions
 # ----------------------------------------------------------------------------
 
+req = urllib.request.Request('http://www.url.com')
```
skoudoro, line 43: - `'http://www.url.com'` : Can you explain what is this url?
- Could you explain why `req` should be a global variable.

docs/source/ext/github_tools.py
```diff
@@ -19,7 +19,7 @@
 from datetime import datetime, timedelta
 from subprocess import check_output
 
-from urllib.request import urlopen, Request
```
skoudoro, line 22: Could you explain why you remove `urlopen`?  The link that you provide does not do it and I do not understand your change.

skoudoro (commented):
Hi @Aju100,

Thank you for trying to improve the security.

Your change is breaking the whole codebase and the tests are not passing anymore.
Can you look deeper into the problem? I recommend you to run the tests locally. You can check the FURY documentation to see how to run them. it will help you to debug.

Thank you

docs/source/ext/github_tools.py
Aju100: As I mention it's a vulnerable calling function that can lead to open up the file too instead of URL only.

docs/source/ext/github_tools.py
Aju100: yes, it has broken out the whole codebase. First of all, req must be a global variable to look at our URL through which we can fetch or use it. It will be better to know about the URL link which we are using it. Thank you

skoudoro: As you can see [here](https://github.com/fury-gl/fury/blob/89aaa7eddc645979047c6740b24f86e489085aeb/docs/source/ext/github_tools.py#L45), `Request` does not to be global for doing this fix.

I still do not understand what is `http://www.url.com`.  Why this URL? why pointing to this place? when going to this url there is nothing.

Aju100: Url should be this one https://api.github.com/repos/{0}/stats/contributors. I think we can disclose or reject this PR for this issue. But in order to know the issue behind it, you can still use the [Bandit](https://github.com/PyCQA/bandit) tool to find out the vulnerabilities on the fury repository. Here is the [link.](https://github.com/PyCQA/bandit)

PR #365, opened 2021-01-27
Closes #352

### LINKED ISSUE #352: Update opacity management in actor.sphere
After looking more in detail into your fix, I think we should avoid using `actor.GetProperty().SetOpacity(opacity)`. It seems not compatible with our architecture. 
Ideally, in the function `actor.sphere`, we would check if `opacity != 1`.  if yes, we should update the color array by adding a dimension if it is an Nx3, or by updating the last dimension if it is an Nx4.

I will merge your PR because it fixes the `actor.point` but we need a new PR to improve this version.

_Originally posted by @skoudoro in https://github.com/fury-gl/fury/issues/351#issuecomment-755368331_

tantheta01:
Added fix for opacity, unit tests included.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1462,6 +1462,15 @@ def sphere(centers, colors, radii=1., phi=16, theta=16,
         src.SetThetaResolution(theta)
         src.SetPhiResolution(phi)
 
+    if isinstance(colors, (list, tuple)):
+        colors = (colors[0], colors[1], colors[2], opacity)
+    elif isinstance(colors, np.ndarray):
+        if colors.ndim==1 or colors.shape[1]==1:
+            colors.reshape(-1,)
+            colors = np.array([colors[0], colors[1], colors[2], opacity])
+        else:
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Hi @tantheta01, 

Could you fix the pep8? Otherwise, it looks ready to go.

Thanks!

tantheta01:
I have resolved PEP8 issues. Thank You

skoudoro:
> I have resolved PEP8 issues. Thank You

Unfortunately, there are still some of them as you can see [here](https://github.com/fury-gl/fury/pull/365#issuecomment-768002953).

SunTzunami:
Hi @skoudoro, @tantheta01! I was working with line actor recently and I think its opacity could use modification too as it uses 
`    actor.GetProperty().SetOpacity(opacity)` to manipulate opacity. Check the documentation [here](https://fury.gl/latest/_modules/fury/actor.html#line).

skoudoro:
Thank you for the feedback @SunTzunami. We might change strategy and use only the color to set up the opacity. So let's keep it as it for the line `actor` for the moment.

PR #368, opened 2021-01-31
Solving The Minor Documentation Error

ganimtron-10:
In the documentation of colormap.py, the documentation regarding the auto argument in the create_colormap function had minor errors as its function interpolate from [0,1] but Docs wrote it as [0,10].


skoudoro:
Good catch! Thank you for this @ganimtron-10. Merging

PR #369, opened 2021-02-01
Added Special Character Support

16bitmood:
Special character support for textbox2d (like: '!@#$%^&*()')


skoudoro (commented):
Hi @16bitmood,

Thank you for adding this feature! we will look into it and get back to you.

fury/ui.py
```diff
@@ -1896,6 +1901,7 @@ class LineSlider2D(UI):
     active_color : (float, float, float)
         Color of the handle when it is pressed.
     """
+
     def __init__(self, center=(0, 0),
                  initial_value=50, min_value=0, max_value=100,
                  length=200, line_width=5,
```
Nibba2018, line 1907: Are these extra new lines necessary?

fury/ui.py
```diff
@@ -2244,6 +2250,7 @@ class LineDoubleSlider2D(UI):
         Color of the handles when they are pressed.
 
     """
+
     def __init__(self, line_width=5, inner_radius=0, outer_radius=10,
                  handle_side=20, center=(450, 300), length=200,
```
Nibba2018, line 2255: same as above and below few new lines.

Nibba2018 (changes requested):
Hi @16bitmood , thanks for creating this PR, I have tested various special characters and they work fine on `TextBox2D`.

But I don't understand your unit test. Please be a bit more elaborate. For me a window pops up and then quickly closes without any keystrokes.

I would suggest you to add some assertions at the end to test whether the key presses are correctly registered/saved or not. For e.g:

* type a string including printable characters.
* close the window.
* then assert if the message parameter contains the same string of the characters or not.

Thanks!

PR #371, opened 2021-02-02
Textbox2d  special character support

16bitmood:
Re-recorded tests and removed spaces.


Nibba2018 (changes requested):
> Hi @16bitmood , thanks for creating this PR, I have tested various special characters and they work fine on `TextBox2D`.
> 
> But I don't understand your unit test. Please be a bit more elaborate. For me a window pops up and then quickly closes without any keystrokes.
> 
> I would suggest you to add some assertions at the end to test whether the key presses are correctly registered/saved or not. For e.g:
> 
> * type a string including printable characters.
> * close the window.
> * then assert if the message parameter contains the same string of the characters or not.
> 
> Thanks!

My review for your PR stays the same as it was for your previous PR. Also, you can push changes directly to this branch to update this PR. It's not necessary for you to create separate PRs for each individual change. Thanks!

16bitmood:
Hey @Nibba2018 , thanks for the review!
I'm getting some issues recording the tests, (and I'm not sure if it's just my machine) Can you check if these issues are also on your system?
- I'm getting a weird issue while recording `test_ui_textbox`. if you enter some specific letters like `e` and `I` sometimes you get a warning:
```
2021-02-09 14:18:14.392 (  39.748s) [        B1AF1740]          vtkObject.cxx:545   WARN| Passive observer should not call AddObserver or RemoveObserver in callback.
```
- Also, when you record a test, they get saved in these directories, is that the expected behaviour?
  - ~/.local/lib/python3.8/site-packages/fury
  - env/lib/python3.8/site-packages/fury
- Even after you load tests from that path, their playback speed does not match with how I recorded them 
On a side note, the textbox looks really small, is that intended?

https://user-images.githubusercontent.com/63354262/107341175-a8ae1200-6ae4-11eb-820a-a5f4a25ee615.mp4

Nibba2018:
> I'm getting some issues recording the tests, (and I'm not sure if it's just my machine) Can you check if these issues are also on > your system?

Hi @16bitmood , good catch! There are some specific characters which are generating this warning and hence interrupting the recording process. I was able to reproduce the same on my environment:

```{'fury_version': '0.6.1.post138.dev0+gc1e169a', 'pkg_path': '/home/nibba/OSS/fury/fury', 'commit_hash': 'c1e169a1724373da8092c17e521fda2a30106171', 'sys_version': '3.7.9 (default, Aug 31 2020, 12:42:55) \n[GCC 7.3.0]', 'sys_executable': '/home/nibba/miniconda3/envs/fury-gl/bin/python', 'sys_platform': 'linux', 'numpy_version': '1.19.4', 'scipy_version': '1.5.4', 'vtk_version': '9.0.1', 'matplotlib_version': '3.3.3', 'dipy_version': '1.3.0'}```.

*Same issue is observed with the latest master commit as well.*

Characters responsible for such issues are(so far): `qei`

The previous(latest master commit) test had the following characters: `ahjfhsffsjkabjsdfbhdasjufhdjasdsjdshajfdhhad` which didn't include them hence it was not caught by the test.

I would recommend you to create an Issue referring to this comment.

> Also, when you record a test, they get saved in these directories, is that the expected behaviour?

That is strange, recorded files must get saved in `fury/data/files/`

> Even after you load tests from that path, their playback speed does not match with how I recorded them.

This is expected as recordings are sped up during playback.

> On a side note, the textbox looks really small, is that intended?

Yes, you can see the source of the test to find out why.

16bitmood:
> I would recommend you to create an Issue referring to this comment.

:+1:,  added that issue.
> That is strange, recorded files must get saved in fury/data/files/

Yes, that's what I thought it should do, I'll look into why it's not saving correctly

Thanks for the detailed feedback.

skoudoro:
Thanks for the review @Nibba2018  and thanks for the issue @16bitmood. 

I can reproduce the issue too. I need to look deeper into the recording and see what's going on. But feel free to introspect it!

skoudoro (approved):
This PR seems ready to go. Thank you again @16bitmood and for creating the issue. I just spent time on the issue and it seems tricky. no easy solution. 

Do you have any other comments @Nibba2018? if no, go ahead and merge it

Issue #64, opened 2019-03-25
Restructuring ui and actors module


skoudoro:
## Description

`ui.py` and `actors.py` start to be huge so it would good to start some thinking on this file restructuration. We need to convert them to package and submodule. 

It would be good to do a short meeting (30-45min) this week or next week to talk about it. If you have some suggestion, feel free to describe your proposition here.

What do you think @MarcCote @ranveeraggarwal @Garyfallidis @dmreagan? Thank you in advance for your feedback.

ranveeraggarwal:
I see a few approaches here:

1. 2D and 3D elements in separate (two) files. But as the elements grow, we'll be back to the same issue.
2. Separate file for each top level elements (the derivative of an element can join the same file). The advantage here is that filenames will document the range of elements we have, but this might suffer from the problem of file bloat in general.
3. Categorize the elements and have separate files for each category. For instance, FileMenu (and future menus) can fall in the menus file while the ring and the line sliders can be in `sliders.py`. 

Furthermore, is there a nice way to do something like forward declaration of classes in Python? Or perhaps a file-level docstring, so that one doesn't have to go through the entire file to get the range of features we provide?

MarcCote:
I thought I had already replied to this issue. My bad :(.

I'm all for splitting the huge `ui.py` into multiple files. I don't have a strong opinion about the separation but shorter files are always better IMO.

@ranveeraggarwal about forward declarations, you mean like listing the available classes/components in the `__init__.py` for instance?

ranveeraggarwal:
@MarcCote yeah importing the available APIs in `__init.py__` sounds right. That would also ensure that we maintain the flat-ness of the UI API (`import x from ui` instead of `import x from ui.component_name`).

PR #374, opened 2021-02-10
ENH: Refactor actor.odf_slicer for increased performances

CHrlS98:
Hi,
I did some refactoring of `actor.odf_slicer` with the objective to make it faster and less memory greedy.
Basically:
* I replaced `for` loops by `numpy` array operations;
* I added the option to supply ODFs in SH coefficients, when given a B matrix;
* I added a method to change the sphere used for visualization when using SH coefficients;
* I changed the `numpy` method called in `utils.set_polydata_triangle` for one that completes faster;
* I applied the affine transform to the grid as well as to the ODFs themselves, as suggested [here](https://github.com/fury-gl/fury/issues/10);
* I moved the class `OdfSlicerActor` to the new module `fury.actors.odf_slicer` to make some space in `fury.actor` module;
* I wrote tests.

If there is some ODF data available somewhere, I would gladly add an example too.

Below are some screenshots showing new functionalities.

**Real-time ODF volume slicing**
real-time-slicing

**Changing sphere resolution during execution**
update-sphere

### DIFF
--- a/fury/actors/odf_slicer.py
+++ b/fury/actors/odf_slicer.py
@@ -46,7 +46,7 @@ class OdfSlicerActor(vtk.vtkActor):
         to be expressed in SF coefficients.
     """
     def __init__(self, odfs, sphere, indices, scale, norm, radial_scale,
-                 global_cm, colormap, opacity, affine=None, B=None):
+                 shape, global_cm, colormap, opacity, affine=None, B=None):
         self.vertices = sphere.vertices
         self.faces = self._reorder_faces(sphere.faces)
         self.odfs = odfs
@@ -54,6 +54,7 @@ class OdfSlicerActor(vtk.vtkActor):
         self.B = B
         self.radial_scale = radial_scale
         self.colormap = colormap
+        self.grid_shape = shape
         self.global_cm = global_cm
 
         # declare a mask to be instantiated in slice_along_axis
@@ -73,13 +74,9 @@ class OdfSlicerActor(vtk.vtkActor):
                 self.odfs /= np.abs(self.odfs).max(axis=-1, keepdims=True)
             self.odfs *= scale
 
-        # Set dimensions of 3D volume to dimensions of the smallest
-        # grid containing the maximum indices in `indices`
-        self.grid_shape = tuple(np.array(self.indices).max(axis=-1) + 1)
-
         # Compute world coordinates of an affine is supplied
-        self.is_world = affine is not None
-        if self.is_world:
+        self.affine = affine
+        if self.affine is not None:
             self.w_verts = self.vertices.dot(affine[:3, :3])
             self.w_pos = apply_affine(affine, np.asarray(self.indices).T)
 
@@ -150,6 +147,8 @@ class OdfSlicerActor(vtk.vtkActor):
             raise ValueError('Can\'t update sphere when'
                              ' using SF coefficients.')
         self.vertices = sphere.vertices
+        if self.affine is not None:
+            self.w_verts = self.vertices.dot(self.affine[:3, :3])
         self.faces = self._reorder_faces(sphere.faces)
         self.B = B
 
@@ -184,7 +183,7 @@ class OdfSlicerActor(vtk.vtkActor):
         """
         Get the position of non-zero voxels inside `mask`.
         """
-        if self.is_world:
+        if self.affine is not None:
             return self.w_pos[mask[self.indices]]
         return np.asarray(self.indices).T[mask[self.indices]]
 
@@ -192,7 +191,7 @@ class OdfSlicerActor(vtk.vtkActor):
         """
         Get the sphere directions onto which is projected the signal.
         """
-        if self.is_world:
+        if self.affine is not None:
             return self.w_verts
         return self.vertices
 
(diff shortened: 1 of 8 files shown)


skoudoro (commented):
Thank you for this nice update @CHrlS98. 

We will look into this. For backward compatibility, we might need to keep the same order of all function parameters. I wonder why the order has been changed. We do not want to break some scripts so we will need to find a way to warn the users. 

Concerning the move to a new module, we plan to do it for all actors as you can see in this issue: https://github.com/fury-gl/fury/issues/64.

Concerning example, there is no ODF data available somewhere. Feel free to generate it, and create a PR in https://github.com/fury-gl/fury-data.  All our example datas are in this place. Then, we create a specific fetcher. 

What is the size of your volume in your demo gif?

Thank again for this good work

CHrlS98:
Hi @skoudoro 
The primary reason why I changed the order of arguments is because I feel like `sphere` should be mandatory, as it crashes when it is not supplied.

 I will look into `fury-data` and add an ODF volume for examples.

The volume shown in the screenshots is a 111 x 105 x 138 order 8 fODF SH coefficients volume of 133.0 MB.

Thanks

skoudoro:
> The primary reason why I changed the order of arguments is because I feel like sphere should be mandatory, as it crashes when it is not supplied.

I agree that the sphere should be defined. So maybe we can just provide a default value instead of `None` and keep the parameters in order

CHrlS98:
> I agree that the sphere should be defined. So maybe we can just provide a default value instead of `None` and keep the parameters in order

The thing is the method expects a `DIPY Sphere`, but `DIPY` is not a dependency of `FURY`. Therefore we can't set the default to some `DIPY` sphere. We would need some workaround to supply a sphere without requiring `DIPY`.

I'll try to think of something.

skoudoro:
From `DIPY SPHERE`,  you need only the vertices and faces. However, this is already present in FURY. 

look at `[fury.primitive.prim_sphere](https://github.com/fury-gl/fury/blob/master/fury/primitive.py#L256)`, it will help.

CHrlS98:
Nice, did not know that, adding a default will be easier than I thought.
I'll do it soon enough!

CHrlS98:
Hi,
I switched the parameter order back to what it was and added a default sphere. I also made a PR [here](https://github.com/fury-gl/fury-data/pull/7) to make available an ODF volume I'll use for my example.

fury/actors/odf_slicer.py
```diff
@@ -0,0 +1,257 @@
...
+                rescaled = sf - sf.min(axis=-1, keepdims=True)
+                rescaled[range_sf > 0] /= range_sf[range_sf > 0][..., None]
+                all_colors =\
+                    create_colormap(rescaled.ravel(), self.colormap) * 255
+            else:
+                all_colors = np.tile(np.array(self.colormap).reshape(1, 3),
+                                     (sf.shape[0]*sf.shape[1], 1))
+        else:
+            all_colors = np.tile(np.abs(self.vertices)*255, (len(sf), 1))
+        return all_colors.astype(np.uint8)
```
skoudoro, line 257: I wonder if this block can be an independent function. I feel like I saw something similar in our codebase. I just can not remember where....

fury/actors/odf_slicer.py
```diff
@@ -0,0 +1,257 @@
...
+        sph_dirs = self._get_sphere_directions()
+        sf = self._get_sf(self.mask)
+
+        all_vertices = self._get_all_vertices(offsets, sph_dirs, sf)
+        all_faces = self._get_all_faces(len(offsets), len(sph_dirs))
+        all_colors = self._generate_color_for_vertices(sf)
+
+        set_polydata_triangles(polydata, all_faces)
+        set_polydata_vertices(polydata, all_vertices)
+        set_polydata_colors(polydata, all_colors)
```
skoudoro, line 179: there is a lot of deep copy here. I think this part can be easily optimized. There are some examples in the `network` examples.

For now, it is fine, but can you add a small note/TODO. It deserves its own PR

skoudoro (commented):
Hi @CHrlS98,

The code is really clean, Thank you for that. The tutorial is working well. I just have a small comment below but apart from that, I think it is ready to go.

After merging, I will tweet your gif to promote your nice work, I think it can be useful for many people and bring other contributors.

Thanks! 

ps: We will fix the design of the combobox.

fury/actors/odf_slicer.py
CHrlS98: @skoudoro How about we move it to `colormap.py` so it is some colormap utility function for generating a color array given some values and a colormap?

CHrlS98:
@skoudoro Thanks for the nice comment,
I addressed your requests but didn't do anything about `_generate_color_for_vertices` yet. I think it could make sense to move it to `colormap.py` although it isn't clear to me what other methods could benefit from this.

On another note, I did a new GIF that I think you'll prefer. I removed the combo box and added transparency to the sagittal and axial slices. I also disabled the mouse capture.
better-slicing

skoudoro:
Ok, I tested it and it is good to go. 

Concerning `_generate_color_for_vertices`, I believe it was `line` or `surface` actors. I will create an issue for that. We have to review color and colormap in general. Sometimes the initialization is between [0-1] and sometimes [0-255]. There is a uniformization task to do.

Thank you for the updated gif. What is your Twitter handle?

Concerning checkboxes, you should look at dipy_horizon. we transform the `X, Y, Z` labels to flat buttons which avoid having checkboxes. The buttons hide/show the slices and sliders.

CHrlS98:
@skoudoro Alrighty thanks,
I'll look into `dipy_horizon` for flat buttons :+1: 
My twitter: **@CHrlS_98**

See you around!

PR #375, opened 2021-02-13
Visuals for some parametric 2D functions

SunTzunami:
I had an idea of making some simple demonstrations of how one can plot 2D parametric functions using FURY. Some functions cannot be plotted by using the cartesian equations but can be easily rendered by using the parametric equations.
Parametric equations are of the form-
x = x(u, v), y = y(u, v), z = z(u, v)

Functions visualized - 
1. Gabriel's Horn

Gabriel's Horn

2. Klein Bottle

Klein bottle

3. Möbius Strip

Mobius Strip

4. Roman Surface

Roman Surface

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -475,10 +475,10 @@ def contour_from_label(data, affine=None, color=None):
         opacity = color[:, -1]
         color = color[:, :-1]
     else:
-        opacity = np.ones((nb_surfaces, 1)).astype(float)
+        opacity = np.ones((nb_surfaces, 1)).astype(np.float)
 
     for i, roi_id in enumerate(unique_roi_id):
-        roi_data = np.isin(data, roi_id).astype(int)
+        roi_data = np.isin(data, roi_id).astype(np.int)
         roi_surface = contour_from_roi(roi_data, affine,
                                        color=color[i],
                                        opacity=opacity[i])
@@ -951,9 +951,9 @@ def tensor_slicer(evals, evecs, affine=None, mask=None, sphere=None, scale=2.2,
             .format(evals.shape, evecs.shape))
 
     if mask is None:
-        mask = np.ones(evals.shape[:3], dtype=bool)
+        mask = np.ones(evals.shape[:3], dtype=np.bool)
     else:
-        mask = mask.astype(bool)
+        mask = mask.astype(np.bool)
 
     szx, szy, szz = evals.shape[:3]
 
@@ -962,7 +962,7 @@ def tensor_slicer(evals, evecs, affine=None, mask=None, sphere=None, scale=2.2,
             self.mapper = None
 
         def display_extent(self, x1, x2, y1, y2, z1, z2):
-            tmp_mask = np.zeros(evals.shape[:3], dtype=bool)
+            tmp_mask = np.zeros(evals.shape[:3], dtype=np.bool)
             tmp_mask[x1:x2 + 1, y1:y2 + 1, z1:z2 + 1] = True
             tmp_mask = np.bitwise_and(tmp_mask, mask)
 
@@ -1050,19 +1050,19 @@ def _tensor_slicer_mapper(evals, evecs, affine=None, mask=None, sphere=None,
     all_xyz = []
     all_faces = []
     for (k, center) in enumerate(ijk):
-        ea = evals[tuple(center.astype(int))]
+        ea = evals[tuple(center.astype(np.int))]
         if norm:
             ea /= ea.max()
         ea = np.diag(ea.copy())
 
-        ev = evecs[tuple(center.astype(int))].copy()
+        ev = evecs[tuple(center.astype(np.int))].copy()
         xyz = np.dot(ev, np.dot(ea, vertices.T))
 
         xyz = xyz.T
         all_xyz.append(scale * xyz + center)
         all_faces.append(faces + k * xyz.shape[0])
 
-        cols[k, ...] = np.interp(cfa[tuple(center.astype(int))], [0, 1],
+        cols[k, ...] = np.interp(cfa[tuple(center.astype(np.int))], [0, 1],
                                  [0, 255]).astype('ubyte')
 
     all_xyz = np.ascontiguousarray(np.concatenate(all_xyz))
@@ -1096,8 +1096,8 @@ def _tensor_slicer_mapper(evals, evecs, affine=None, mask=None, sphere=None,
 
 
 def peak_slicer(peaks_dirs, peaks_values=None, mask=None, affine=None,
-                colors=(1, 0, 0), opacity=1., linewidth=1, lod=False,
-                lod_points=10 ** 4, lod_points_size=3, symmetric=True):
+                colors=(1, 0, 0), opacity=1., linewidth=1,
+                lod=False, lod_points=10 ** 4, lod_points_size=3):
     """Visualize peak directions as given from ``peaks_from_model``.
 
     Parameters
@@ -1115,10 +1115,13 @@ def peak_slicer(peaks_dirs, peaks_values=None, mask=None, affine=None,
     colors : tuple or None
         Default red color. If None then every peak gets an orientation color
         in similarity to a DEC map.
+
     opacity : float, optional
         Takes values from 0 (fully transparent) to 1 (opaque)
+
     linewidth : float, optional
         Line thickness. Default is 1.
+
     lod : bool
         Use vtkLODActor(level of detail) rather than vtkActor.
         Default is False. Level of detail actors do not render the full
@@ -1127,10 +1130,6 @@ def peak_slicer(peaks_dirs, peaks_values=None, mask=None, affine=None,
         Number of points to be used when LOD is in effect. Default is 10000.
     lod_points_size : int
         Size of points when lod is in effect. Default is 3.
-    symmetric: bool, optional
-        If True, peaks are drawn for both peaks_dirs and -peaks_dirs. Else,
-        peaks are only drawn for directions given by peaks_dirs. Default is
-        True.
 
     Returns
     -------
@@ -1152,7 +1151,7 @@ def peak_slicer(peaks_dirs, peaks_values=None, mask=None, affine=None,
     grid_shape = np.array(peaks_dirs.shape[:3])
 
     if mask is None:
-        mask = np.ones(grid_shape).astype(bool)
+        mask = np.ones(grid_shape).astype(np.bool)
 
     class PeakSlicerActor(vtk.vtkLODActor):
         def __init__(self):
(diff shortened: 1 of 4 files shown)


skoudoro (commented):
Hi @SunTzunami,

Thank you for this. As you can see [here](https://www.paraview.org/Wiki/VTK/Examples/Python/GeometricObjects/ParametricObjectsDemo), VTK has already a lot of parametric functions. Those parametric functions use surface instead of the point actor. 

I suggest creating one actor, that can return the desired parametric functions. 
Also, it would be better to have:
- 1 tutorial with this new function and all object in the scene
- 1 tutorial with what you just created to explain more in details the magic. 

I don't know if I am clear, so do not hesitate if you have any questions.
Thanks!

SunTzunami:
Hi @skoudoro ! Thanks for the review! In this program, I made a function. Is this what you require the actor to be like? 

actor1

The function I've used in this example uses line actor to generate the parametric surfaces. It's best to use surface actor I feel but surface actor doesn't generate surface properly. Even line actor has its limitations (it can't generate all surfaces, I've implemented only those surfaces that could be generated properly by the line actor). Point actor can generate all surfaces but it's too computationally taxing and slow. 

Rendering same surface (Dini's surface) by using all three actors, you can see the difference- 

1. Surface actor

Dini_surface1

2. Line actor

Dini_surface2

3. Point actor (correct rendering)

Dini_surface3

The surface actor and line actor are not rendering correctly. The point actor renders properly (it takes more time than line and surface actors to be rendered though).

SunTzunami:
Hi @skoudoro ! I figured out how to use surface actor to generate the parametric surfaces- triangulation is different for parametric surfaces. Earlier I was triangulating it normally for cartesian coordinates but when using parametric equations, we have to triangulate the parameters (u and v) too. After triangulating the surfaces properly, the result was as follows- 

parametric_surfaces

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    def sin(x):
+        return np.sin(x)
+
+    def cos(x):
+        return np.cos(x)
+
+    npoints = 600
+
+    def scaler(arr):
```
skoudoro, line 46: I think, it would be better to call it normalize.

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+        arr = np.array(arr)
+        maxima = np.max(arr)
+        minima = np.min(arr)
+        return (arr - maxima)/(maxima - minima)
+
+    if (name == "mobius_strip"):
```
skoudoro, line 52: no parenthesis needed

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+        y = coords[1] + scale*(1 + v/2 * cos(u/2)) * sin(u)
+        z = coords[2] + scale*v/2 * sin(u/2)
+
+        xyz = np.array([(a, b, c) for (a, b, c) in zip(x, y, z)])
+        if cmap:
+            v = scaler(z)
+            colors = colormap.create_colormap(v, name=cmap_name)
+        return(actor.line([xyz], colors=colors, opacity=opacity))
+
+    elif (name == "kleins_bottle"):
```
skoudoro, line 69: same comment as above

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+                               80*cos(u)**5*cos(v)*sin(u) +
+                               80*cos(u)**7*cos(v)*sin(u)))
+        z = coords[2] + scale*(2/15*(3 + 5*cos(u)*sin(u))*sin(v))
+        if cmap:
+            v = scaler(y)
+            colors = colormap.create_colormap(v, name=cmap_name)
+        xyz = np.array([(a, b, c) for (a, b, c) in zip(x, y, z)])
+        return(actor.line([xyz], colors=colors, opacity=opacity))
+
+    elif (name == 'roman_surface'):
```
skoudoro, line 92: same comment as above

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+        y = coords[1] + scale*a*cos(u)*sin(u)*cos(v)
+        z = coords[2] + scale*a*cos(u)**2*cos(v)*sin(v)
+
+        if cmap:
+            v = scaler(z)
+            colors = colormap.create_colormap(v, name=cmap_name)
+        xyz = np.array([(a, b, c) for (a, b, c) in zip(x, y, z)])
+        return(actor.line([xyz], colors=colors, opacity=opacity))
+
+    elif (name == 'boys_surface'):
```
skoudoro, line 110: same comment as above

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
...
+# Importing necessary modules
+
+from fury import window, actor, ui, colormap
+import numpy as np
+
+###############################################################################
+# Function that returns the parametric surface specified by the user.
+
+
+def parametric_func(name, colors=(1, 1, 1), coords=[0, 0, 0],
```
skoudoro, line 11: - call it just `parametric`
- This function should go in `actor` module. Also, it will be good to add a test for this function
- `centers` instead of `coords`
- `ntype` instead of `name`
- It should handle many centers and many `ntype` at the same time. I would recommend to look at the other actors like sphere, square, etc... You can create in one call, multiple objects.

docs/examples/viz_parametric_surfaces_v2.py
```diff
@@ -0,0 +1,250 @@
...
+    Returns
+    -------
+    vtkActor
+
+    """
+    def sin(x):
+        return np.sin(x)
+
+    def cos(x):
+        return np.cos(x)
```
skoudoro, line 44: same comment as above

docs/examples/viz_parametric_surfaces_v2.py
```diff
@@ -0,0 +1,250 @@
...
+        u = u.reshape(-1)
+        v = v.reshape(-1)
+        points2D = np.vstack([u, v]).T
+        tri = Delaunay(points2D)
+
+        x = coords[0] + scale*(1 + v/2 * cos(u/2)) * cos(u)
+        y = coords[1] + scale*(1 + v/2 * cos(u/2)) * sin(u)
+        z = coords[2] + scale*v/2 * sin(u/2)
+
+        xyz = np.array([(a, b, c) for (a, b, c) in zip(x, y, z)])
```
skoudoro, line 67: each element is a primitive for us. you should look at `fury.primitive`.  Every primitive returns vertices and faces. Then, the actor call the primitive and build the actor.

You should look at `actor.square` or `actor.superquadrics` to understand how it works and reproduce it with your case. Let us know if you have any question

skoudoro (changes requested):
Hi @SunTzunami,

Great for surface! There is still a lot of work but you are on the good path. You need to create a specific actor. All this code can not stay in the tutorial. You should look at:
- What is a primitive in FURY, 
- How we create an actor from a primitive. 
This is exactly your case here. [`actor.superquadric`](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1755) is a good example to get inspiration for you. You will need to add tests for primitive and actor. This will allow to set up the direction, scale, etc...
and then, update your tutorial. Let us know if you have any questions.

SunTzunami:
Thanks for the review @skoudoro ! I'll work upon creating an actor.

SunTzunami:
Hi @skoudoro ! I tried using the primitive to generate the actors. I'm getting the right shape for all the surfaces but the rendering is kinda puzzling- the shapes seem to look different when viewed from different angles. I don't know why this is happening.  Some examples- 

1. Mobius Strip

mobius

2. Klein's Bottle

klein

skoudoro:
It will be hard to guess without code. Please, push your last version.

SunTzunami:
I added the code @skoudoro . PTAL.

skoudoro (commented):
I think
- this function: https://github.com/fury-gl/fury/blob/master/fury/primitive.py#L24
- this link https://stackoverflow.com/a/28372331

might help you.

SunTzunami:
Thanks @skoudoro !! I'll read about it.

SunTzunami:
Hi @skoudoro ! I tried using ConvexHull and faces_from_sphere_vertices functions but none yield the desired result-

With Delaunay- 

Delaunay


With faces_from_sphere_vertices function (which uses ConvexHull)-

faces_from_sphere_vertices

With ConvexHull- 

ConvexHull

I can't figure out how to proceed.

docs/examples/viz_parametric_surfaces.py
filipinascimento: I think this name is ok, `scaler` can be used as a short name for `maxmin_scaler` while `normalization` is more appropriate if the array is being divided by its sum or magnitude.

filipinascimento:
> With Delaunay-
> 
> Delaunay
> 
> I can't figure out how to proceed.

For the Delaunay surface, an easy way to fix that is disabling backface culling. You just need to call this for the actor:
```python
actor.GetProperty().BackfaceCullingOff()
```
I'm not sure if convex hull methods can help in this case as these curves are 2D manifolds that can only be properly embedded to 4D.

docs/examples/viz_parametric_surfaces_v2.py
```diff
@@ -0,0 +1,250 @@
...
+        Set it to true if colormap is being mapped to the surface.
+    cmap_name : string, optional
+        Stores the name of the colormap specified by the user, cmap should be
+        set to True for the colormap to be rendered.
+
+    Returns
+    -------
+    vtkActor
+
+    """
```
filipinascimento, line 39: There are three files in the example and at least two different techniques working (Delaunay triangulation and point-based). It would be great to merge these different techniques to render the curves into a single function and have a parameter to select between them.

docs/examples/viz_parametric_surfaces.py
```diff
@@ -0,0 +1,180 @@
+###############################################################################
+# Importing necessary modules
+
+from fury import window, actor, ui, colormap
+import numpy as np
+
+###############################################################################
+# Function that returns the parametric surface specified by the user.
+
+
```
filipinascimento, line 10: It would be great if you can provide a description of the demos here.

docs/examples/viz_parametric_surfaces_v2.py
```diff
@@ -0,0 +1,250 @@
...
+    def cos(x):
+        return np.cos(x)
+
+    npoints = 400
+
+    def scaler(arr):
+        arr = np.array(arr)
+        maxima = np.max(arr)
+        minima = np.min(arr)
+        return (arr - maxima)/(maxima - minima)
```
filipinascimento, line 52: Not being used in this file.

Issue #360, opened 2021-01-19
Dashline / dotted line actor


skoudoro:
It will be good to add this new actor.

We can add an option to the line actor or create a specific actor. This actor should contain an option for this type of dash (point, line, etc)

tushar5526:
Hey @skoudoro, I would like to add this feature.

skoudoro:
Feel free to go ahead!

carrycooldude:
Is this issue is open? @skoudoro

carrycooldude:
> Thanks for the interest @carrycooldude, but I'm already working on this issue.
> But, I highly encourage you to look into other equally interesting issues in the fury codebase.

No issue brother but let me how you do that 😊...

haran2001:
You can start by having a look at the visualization toolkit documentation and playing around with various actors, images and textures. Later you can compare and contrast techniques from VTK with those implemented in fury and suggest/find some room for improvement.
You can always clarify doubts from the maintainers via texts and/or meets, as they are very kind and highly encouraging.
I hope this helps.

sid-008:
Is anyone still working on this issue? If not can I start working on it? @skoudoro

skoudoro:
You can go ahead @SaraSaadoun. 

Please, look at the comments of the first implementation #380 to not reproduce the same error. 

Thanks !

PR #380, opened 2021-02-16
Added Actor for dotted line 

haran2001:
As there is no PR yet for issue #360, I've given it a try using VTK.

dotted_line
 
I tried to add some basic parameters, and am looking forward to add anymore if necessary.
I have also added the necessary tests.

Please do review them and let me know of any room for improvement.


fury/actor.py
```diff
@@ -754,6 +754,80 @@ def callback(_caller, _event, calldata=None):
     return actor
 
 
+def dashed_line(p_initial =[0.0, 0.0, 0.0], p_final=[0.0, 0.0, 0.0],
```
skoudoro, line 757: Please, do not provide default value for `start_pos` and `end_pos`.

fury/actor.py
```diff
@@ -754,6 +754,80 @@ def callback(_caller, _event, calldata=None):
...
+
+    # Create a polydata object
+    point = vtk.vtkPolyData()
+
+    # Set the points and vertices as the geometry/topology of the polydata
+    point.SetPoints(points)
+    point.SetVerts(vertices)
+    mapper.SetInputData(point)
+
+    actor = vtk.vtkActor()
```
skoudoro, line 824: We have a `point` actor and `line` actor. Can you use them instead of going from scratch? Thank you

skoudoro (commented):
Hi @haran2001,

Thank you for doing this. Can you fix all the pep8 issues? Also, you are only using points for the dashed line. It would be good if you could manage the line. Look at `line` and `point` actor.

Thank you

fury/actor.py
haran2001: Shall I set them as None? or should I leave it completely?

fury/actor.py
haran2001: Sure, I will try to implement it using the suggested inbuilt actors.
Thank you very much for reviewing.

haran2001:
Hi @skoudoro , I'm encountering a strange issue while using the fury line actor for a **dashed** line
As per the following code, I expect two separate lines,

import numpy as np

scene = window.Scene()
colors = [1.0, 0.0, 0.0]
c = actor.line(np.random.rand(2, 3), colors)
scene.add(c)
d = actor.line(np.random.rand(2, 3), colors)
scene.add(d)
window.show(scene) 


But the following is the result

test


As you can see, there seems to be a faint connection between the two lines.
This prevents me from creating a dashed line using the fury line actor, as the faint connections appear and make the dashed line look like a whole line.
I think we can solve this by using VTK methods directly instead just for the dashed line.

However fury point actors work just fine for the **dotted** line.

dotted_line2


Let me know how I should proceed and if you are fine with using VTK only for the dashed line.
Thank you.

SunTzunami:
Hi @haran2001! I was using line actor recently, I think this solves your dilemma-

```
scene = window.Scene()
colors = [1.0, 0.0, 0.0]
c = actor.line([np.random.rand(2, 3)], colors)
scene.add(c)
d = actor.line([np.random.rand(2, 3)], colors)
scene.add(d)
window.show(scene) 
```

Notice the square brackets.

haran2001:
I never noticed this detail, thanks a ton @SunTzunami !

haran2001:
Hi, I have utilized fury actors and implemented two functions to carry out the dashed/dotted line features
1. dashed_line() -->

dash_line_final

2. dotted_line() -->

dotted_line_final
 

Please do go through them and let me know of any appropriate changes, if necessary.
The required tests have also been added.

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
     return actor
 
 
+def dashed_line(start_pos, end_pos, num_lines=10, line_color=[1.0, 0.0, 0.0], line_fraction=0.5):
+     """Create an actor for one or more dashed lines.
+
+    Parameters
+    ------------
+    start_pos :  list (1, 3)
+    end_pos :  list (1, 3)
```
skoudoro, line 763: to be consistent with the rest of the code, it should be a `np.array`. For an example, you can look at other actors

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+    num_lines :  integer
```
skoudoro, line 764: `num_lines :  int, optional` instead

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+    line_color :  list (1, 3)
```
skoudoro, line 765: same comment as above

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+    line_fraction :  float (between 0 and 1)
```
skoudoro, line 766:
```python 
line_fraction :  float, optional
     it should be between 0 and 1
```

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+        Dashed Line.
+
+    Examples
+    ----------
+    >>> from fury import window, actor
+    >>> scene = window.Scene()
+    >>> start_pos = [.0, 0.0, 0.0]
+    >>> end_pos = [1.0, 1.0, 0.0]
+    >>> num_lines = 20
+    >>> line_color = [0.0, 1.0, 0.0]
```
skoudoro, line 780: to update with `np.array`

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+    >>> line_fraction = 0.5
+    >>> c = actor.dashed_line(start_pos, end_pos, num_lines, line_color,line_fraction)
+    >>> scene.add(c)
+    >>> window.show(scene)
+
+    """
+    start_pos = np.array(start_pos)
+    end_pos = np.array(end_pos)
```
skoudoro, line 788: implicit conversion not necessary, let's keep it simple

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+def dotted_line(start_pos, end_pos, num_points=10, point_color=[1.0, 0.0, 0.0], point_radius=0.01):
+        """Create an actor for one or more dotted lines.
+
+    Parameters
+    ------------
+    start_pos :  list (1, 3)
+    end_pos :  list (1, 3)
+    num_points :  integer
+    point_color :  list (1, 3)
+    point_radius :  float
```
skoudoro, line 812: same comment as above

fury/actor.py
```diff
@@ -754,6 +754,95 @@ def callback(_caller, _event, calldata=None):
...
+    >>> from fury import window, actor
+    >>> scene = window.Scene()
+    >>> start_pos = [0.0, 0.0, 0.0]
+    >>> end_pos = [1.0, 1.0, 0.0]
+    >>> num_points = 20
+    >>> point_color = [1.0, 0.0, 0.0]
+    >>> point_radius = 0.01
+    >>> c = actor.dotted_line(start_pos, end_pos, num_points, point_color,point_radius)
+    >>> scene.add(c)
+    >>> #window.show(scene)
```
skoudoro, line 830: same comment as above

fury/tests/test_actors.py
```diff
@@ -869,6 +869,85 @@ def test_points(interactive=False):
...
+
+    arr = window.snapshot(scene)
+    report = window.analyze_snapshot(arr,
+                                     colors=[point_color])
+    npt.assert_equal(report.objects, 10)
+
+    scene.add(c)
+    arr = window.snapshot(scene)
+    report = window.analyze_snapshot(arr, colors=(1, 0, 0))
+    npt.assert_equal(report.colors_found, [True])
```
skoudoro, line 925: duplicated with above?

skoudoro (commented):
Thank you for the update @haran2001!

You can see below some quick comments but I will go through this PR later in the day. Please, try to fix the pep8 issue as you can see [here](https://github.com/fury-gl/fury/pull/380#issuecomment-779815136)

haran2001:
Hi , I have made the requested changes .
However I'm having some difficulty in troubleshooting the checks.
If you have any suggestions regarding the same, then do let me know. 
Thank You.

Nibba2018:
@haran2001 Can you explain the test you are trying to implement? thanks.

haran2001:
Both tests are quite similar @Nibba2018 , they first render an actor and check  if the actors have the properties they are expected to have.

dotted_line()
1. Number of actors = 1
2. Number of dots in the line = 10
    (This is done by getting all the objects of a given color (red in this case) and checking for their number (using report.objects))
3. Color of the dots = (1.0, 0.0, 0.0) = 'red'
    
dashed_line()
1. Check for the number of actors = 1
2. Check for the number of dashed lines in the actor = 10 
     (This is done by getting all the objects of a given color (red in this case) and checking for their number (using report.objects))

skoudoro:
Below the error that you can see [here](https://github.com/fury-gl/fury/pull/380/checks?check_run_id=1936788656#step:10:886)
```python
    def test_dotted_line(interactive=False):
        start_pos = np.array([1.0, 0.0, 0.0])
        end_pos = np.array([0.0, 1.0, 0.0])
        point_radius = 0.01
        num_points = 10
        point_color = np.array([1.0, 0.0, 0.0])
        c = actor.dotted_line(start_pos, end_pos,
                              num_points, point_color, point_radius)
    
        scene = window.Scene()
        scene.add(c)
    
        if interactive:
            window.show(scene, order_transparent=True)
    
        arr = window.snapshot(scene)
        report = window.analyze_snapshot(arr,
                                         colors=[point_color])
        npt.assert_equal(report.objects, 10)
    
        scene.add(c)
        arr = window.snapshot(scene)
        report = window.analyze_snapshot(arr, colors=(1, 0, 0))
>       npt.assert_equal(report.colors_found, [True])
E       AssertionError: 
E       Items are not equal:
E       item=0
E       
E        ACTUAL: False
E        DESIRED: True
```

skoudoro:
I would recommend saving the snapshot as png to make sure you have the expected image

haran2001:
I have removed the check for color as its was producing an unexpected result.
Thank you @skoudoro for pointing it out.

skoudoro (commented):
Hi @haran2001,

Could you explain why the test for the color did not work? it works for other actors so it will be good if we understand what is wrong here. 

`dashed_line` actor manages only 1 line. it will be the only actor with this behavior so we should avoid this. Could you improve your actor to manage multiple dashed lines with different colors. We should be able to do this:
```python
start_pos = np.array([[1.0, 0.0, 0.0], [2.0, 0.0, 0.0]])
    end_pos = np.array([[0.0, 1.0, 0.0], [0.0, 2.0, 0.0]])
    line_fraction = 0.5
    num_lines = 10
    line_color = np.array([[1.0, 0.0, 0.0], [0, 0, 1]])
    c = actor.dashed_line(start_pos, end_pos,
                          num_lines, line_color, line_fraction)
```

We just create 2 parallels dashed lines. However, it does not work with the current code. Can you look at it? Thank you

haran2001:
Yes, i will look into this

fury/actor.py
```diff
@@ -754,6 +754,97 @@ def callback(_caller, _event, calldata=None):
...
+    >>> num_points, point_color,point_radius)
+    >>> scene.add(c)
+    >>> #window.show(scene)
+    """
+    dp = (end_pos - start_pos) / num_points
+    point_lis = []
+    count = 0
+
+    while count < num_points:
+        point_lis.append(list(start_pos + (count * dp)))
```
Garyfallidis, line 841: This is slow. Please avoid using lists. Use numpy arrays instead. Allocate memory in the beginning.

haran2001:
Hi, I have made the requested changes.
Now multiple lines can be initialized and both the methods only use numpy arrays.
1. dashed_line()

dashed_line

2. dotted_line()
 
dotted_line

Do let me know of any further changes, if necessary.

Garyfallidis:
Please look at this option too. Using a texture to create the dashed line. 
https://lorensen.github.io/VTKExamples/site/Python/Rendering/StippledLine/
You can use it in the following way with ``fury.actor.line``. 

    import numpy as np
    from fury.actor import line
    # lines = [np.random.rand(10, 3), np.random.rand(20, 3)]
    lines = [np.array([[0, 0, 0], [50, 0, 0.]])]

    act = line(lines, colors=np.array([[1, 0, 0]]), lod=False)
    StippledLine(act, 0xA1A1, 2)

Supports many different patterns and it should be faster than the proposed approach
(for dashed lines).

haran2001:
Hi, I have tried to apply the stippled line method (texture) on the fury line actor, but I'm encountered the following issue.
The output of implementing Stippled_line() with fury.actor.line() is just a regular line. 

dashed_line_test1

I believe this is because the fury.actor.line actor using a VTK LODActor.
Using LOD actor fury.actor.line can return multiple actors at the same time, but the LODActor doesn't support application of textures as is evident from the docs for the actor. (There are no methods to get/set textures) 
https://vtk.org/doc/nightly/html/classvtkLODActor.html


An alternative to this approach is to use the VTK line source directly, which produces the following result. (Since the VTK line source supports get/set texture methods)

stippled_line

However the drawback to this idea is that it can return only **one actor at a time**, hence plotting multiple dashed lines at the same time would not be possible.

Presently I'm working on some other alternatives, however in the meanwhile any suggestions/ideas from the maintainers on how to proceed further will be much appreciated. 
Thank you.

haran2001:
(Code using VTK line source)


from fury import window
import vtk

def StippledLine(start_pos, end_pos, lineStipplePattern, lineStippleRepeat, colors):
    lines = vtk.vtkLineSource()
    lines.SetResolution(11)
    lines.SetPoint1(start_pos)
    lines.SetPoint2(end_pos)

    mapper = vtk.vtkPolyDataMapper()
    mapper.SetInputConnection(lines.GetOutputPort())

    actor = vtk.vtkActor()
    actor.SetMapper(mapper)
    actor.GetProperty().SetColor(colors)

    image = vtk.vtkImageData()
    texture = vtk.vtkTexture()

    dimension = 16 * lineStippleRepeat

    image.SetDimensions(dimension, 1, 1)
    image.AllocateScalars(vtk.VTK_UNSIGNED_CHAR, 4)
    image.SetExtent(0, dimension - 1, 0, 0, 0, 0)
    on = 255
    off = 0
    i_dim = 0
    while i_dim < dimension:
        for i in range(0, 16):
            mask = (1 << i)
            bit = (lineStipplePattern & mask) >> i
            value = bit
            if value == 0:
                for j in range(0, lineStippleRepeat):
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 0, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 1, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 2, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 3, off)
                    i_dim += 1
            else:
                for j in range(0, lineStippleRepeat):
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 0, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 1, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 2, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 3, on)
                    i_dim += 1

    texture.SetInputData(image)
    texture.InterpolateOff()
    texture.RepeatOn()

    actor.SetTexture(texture)
    return actor


scene = window.Scene()
start_pos = [0.0, 0.0, 0.0]
end_pos = [1.0, 1.0, 1.0]
c = StippledLine(start_pos, end_pos, 0xAAAA, 2, (1, 0, 0))
scene.add(c)
window.show(scene)

haran2001:
(Code using fury line actor)


import numpy as np
from fury.actor import line
import vtk
from fury import window

def StippledLine(act, lineStipplePattern, lineStippleRepeat):
    image = vtk.vtkImageData()
    texture = vtk.vtkTexture()

    dimension = 16 * lineStippleRepeat

    image.SetDimensions(dimension, 1, 1)
    image.AllocateScalars(vtk.VTK_UNSIGNED_CHAR, 4)
    image.SetExtent(0, dimension - 1, 0, 0, 0, 0)
    on = 255
    off = 0
    i_dim = 0
    while i_dim < dimension:
        for i in range(0, 16):
            mask = (1 << i)
            bit = (lineStipplePattern & mask) >> i
            value = bit
            if value == 0:
                for j in range(0, lineStippleRepeat):
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 0, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 1, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 2, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 3, off)
                    i_dim += 1
            else:
                for j in range(0, lineStippleRepeat):
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 0, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 1, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 2, on)
                    image.SetScalarComponentFromFloat(i_dim, 0, 0, 3, on)
                    i_dim += 1

    texture.SetInputData(image)
    texture.InterpolateOff()
    texture.RepeatOn()

    act.SetTexture(texture)

    return act


scene = window.Scene()
lines = [np.array([[0, 0, 0], [50, 0, 0.]])]

act = line(lines, colors=np.array([[1, 0, 0]]), lod=False)
StippledLine(act, 0xA1A1, 2)
scene.add(act)
window.show(scene)

skoudoro:
Thank you for your feedback @haran2001.

We will look into it and test it from our side.

haran2001:
Thank you @skoudoro,
I also wanted to mention that the present implementation is of complexity O ( n ) , where as the implementation via textures is of complexity O ( n ^ 2 ) which makes the current implementation more efficient. (at least for large n)
Do share your thoughts on this, if any.

haran2001:
> Thank you for your feedback @haran2001.
> 
> We will look into it and test it from our side.

Any update on this @skoudoro?

fury/actor.py
```diff
@@ -754,6 +754,103 @@ def callback(_caller, _event, calldata=None):
...
+
+    for start_pos, end_pos, x in zip(start_pos, end_pos, range(0, N)):
+        dp = (end_pos - start_pos) / num_points
+        p = start_pos.copy()
+        count = 0
+        while count < num_points:
+            arr[x * num_points + count] = np.array([p])
+            count = count + 1
+            p = p + dp
+    c = point(points=arr, colors=np.repeat(colors, num_points, axis=0),
```
nasimanousheh, line 849: Thank you @haran2001 for creating dotted line and dashed line. They are very useful. Here are my comments:
- It's better to have two options for creating actor in dotted_line:  "point" and "dots" actors. The point actor is slow because it is creating spheres with full geometry. People should be able to use ``actor.dots`` instead that does not generated 3D geometry. Also there is a fast sphere rendering option put that can be added at another PR.
- If num_points=2, it will create two points, the first point is located at start_pos, the second point is located in the middle of the line (not at end_pos). Here I showed two snapshots. To make sure if the second point is not located at end_pos, I created sphere actors for start_point and end_point with bigger radius. You will see that the second point is exactly in the middle of start_pos and end_pos. I think this should be corrected.
image

fury/tests/test_actors.py
```diff
@@ -869,6 +869,56 @@ def test_points(interactive=False):
     npt.assert_equal(report.objects, 3)
 
 
+def test_dotted_line(interactive=False):
+    start_pos = np.array([[0, 0, 0]])
+    end_pos = np.array([[1, 1, 1]])
+    radius = 0.01
+    num_points = 10
```
nasimanousheh, line 876: We can have two options here. 1. define the num_points directly. 2. or the length of segments and gaps are fixed and based on the length of line, the number of segments with gaps will increase properly depending on length of the line. 
For option 2, the size of segment and size of the gap (between segments) are fixed, so if we increase the length of line, we are just adding more segments with gap (with fixed size). The size of last segment that reaches the end_pos, can be flexible until it reaches the end_point. I sent you a GIF file for a better explanation. 
dashed_line

fury/actor.py
haran2001: Resetting the end position is an simple fix. However adding the dots option is difficult as the dot actor is implemented through a **single unstructured grid**  in VTK . This makes it difficult to plot multiple dotted lines with different colors and all the line end up with the same color of the grid actor.

fury/tests/test_actors.py
haran2001: Can you have a look at the line fraction parameter of dashed lines actor? It allows the developer to change the length (or space) without altering the number of segments.

haran2001:
Thank you for the detailed review @nasimanousheh

fury/actor.py
haran2001: Do you have any suggestions on how to approach this issue?

haran2001: The dots_actor parameter has been added, which uses fury dots to render the dotted line.

fury/tests/test_actors.py
haran2001: The line_fraction parameter allows the user to alter the spaces without changing the total number of segments, as shown below:

1. The dashed line with two segments with line fraction set to 25% (0.25)
25_percent

2. The dashed line with two segments with line fraction set to 50% (0.50)
50_percent

3. The dashed line with two segments with line fraction set to 75% (0.75)
75_percent

The number of segments is constant (two), but the spacing can be varied.

PR #385, opened 2021-02-23
fixed the example for superquadric function

SunTzunami:
Earlier, `scales` was incorrectly initialised and `colors` was not initialised at all, as a result, example was not working. That's fixed now.


skoudoro:
Thanks @SunTzunami!

PR #387, opened 2021-02-24
[fix]   Propagate update_actor

skoudoro:
This PR assures that we update all array when we call `update_actor`

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -1008,7 +1008,7 @@ def fix_winding_order(vertices, triangles, clockwise=False):
     return corrected_triangles
 
 
-def vertices_from_actor(actor):
+def vertices_from_actor(actor, as_vtk_type=False):
     """Access to vertices from actor.
 
     Parameters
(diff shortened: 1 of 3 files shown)


fury/utils.py
```diff
@@ -1008,41 +1008,70 @@ def fix_winding_order(vertices, triangles, clockwise=False):
...
-                                      GetPoints().GetData())
+    vtk_array = actor.GetMapper().GetInput().GetPoints().GetData()
+    if as_vtk_type:
+        return vtk_array
 
+    return numpy_support.vtk_to_numpy(vtk_array)
 
-def colors_from_actor(actor, array_name='colors'):
+
+def colors_from_actor(actor, array_name='colors', as_vtk_type=False):
```
Garyfallidis, line 1032: ``as_vtk`` or ``as_vtk_array``

fury/utils.py
```diff
@@ -1056,15 +1085,22 @@ def compute_bounds(actor):
...
 
-def update_actor(actor):
+def update_actor(actor, all_arrays=True):
     """Update actor.
 
     Parameters
     ----------
     actor : actor
+    all_arrays: bool, optional
+        if False, only vertices are updated
```
Garyfallidis, line 1095: if True ...

Garyfallidis:
Thank you @skoudoro. @filipinascimento please update your example to use ``update_actor``. You should not need to call modified that often from now on.

PR #389, opened 2021-03-06
ENH: peaks_slicer option for asymmetric peaks visualization

CHrlS98:
Hi,
This is a very small enhancement but it has been very useful to me for visualizing principal directions extracted from asymmetric fiber ODFs.

I added an argument `symmetric`, defaulting to `True`, to the `peak_slicer` actor. When set to `True`, it shows the peaks in directions `d` and `-d` whereas when set to `False` it only renders the peaks in the direction given by `d`.

Thank you!


skoudoro (commented):
Hi @CHrlS98,

Yes, I agree, can be useful, Thank you for that. 

It looks good to me, I need to test it locally, and then I will go ahead and merge it.

If someone wants to have a look, feel free to do it!

skoudoro:
Thanks @CHrlS98

Issue #355, opened 2021-01-08
[Vulnerability Bug] Used blacklisted dangerous function call that can lead to RCE


Aju100:
Hi Opensource enthusiast,

After looking through the whole codebase, I get to identify a few vulnerabilities on the fury repository. 

### Impact
- urlib not only opens http:// or https:// URLs, but also ftp:// and file://. With this it might be possible to open local files on the executing machine which might be a security risk if the URL to open can be manipulated by an external user.
- It can lead to self-RCE, and also general RCE if we can trigger it for the remote end-user.

### Vulnerable code
github_tools.py

```python
from urllib.request import urlopen, Request

... some code here
def fetch_url(url):
    req = Request(url)
    if GH_TOKEN:
        req.add_header('Authorization', 'token {0}'.format(GH_TOKEN))
    try:
        print("fetching %s" % url, file=sys.stderr)
        # url = Request(url,
        #               headers={'Accept': 'application/vnd.github.v3+json',
        #                        'User-agent': 'Defined'})
        f = urlopen(req)
```

### Patches
github_tools.py
```python
from urllib.request import Request
...... some codes
req = urllib.request.Request('urlhere.com')

def fetch_url(url):
    req = Request(url)
    if GH_TOKEN:
        req.add_header('Authorization', 'token {0}'.format(GH_TOKEN))
    try:
        print("fetching %s" % url, file=sys.stderr)
        # url = Request(url,
        #               headers={'Accept': 'application/vnd.github.v3+json',
        #                        'User-agent': 'Defined'})
        f = urlopen(req)
```

### Vulnerable code
fetcher.py
```python
from urllib.request import urlopen
.... some code

def _get_file_data(fname, url):
    with contextlib.closing(urlopen(url)) as opener:
        try:
            response_size = opener.headers['content-length']
        except KeyError:
            response_size = None

        with open(fname, 'wb') as data:
            if response_size is None:
                copyfileobj(opener, data)
            else:
                copyfileobj_withprogress(opener, data, response_size)

```
### Patches
fetcher.py
```python
import urllib.request
... some code 

url = urllib.request.Request('urlhere.com')
def _get_file_data(fname, url):
    with contextlib.closing(urlopen(url)) as opener:
        try:
            response_size = opener.headers['content-length']
        except KeyError:
            response_size = None

        with open(fname, 'wb') as data:
            if response_size is None:
                copyfileobj(opener, data)
            else:
                copyfileobj_withprogress(opener, data, response_size)

```
[...]

PR #399, opened 2021-03-26
Resolve warnings #317 & and Fix Issue: #355

amitchaudhari9121:
This commit will hopefully close issues #317 and #355.
---

I did not create tests for layout.py since it looks like a helper in actors.py.
I could say the same about convert.py but even then, I tried creating basic tests.
If these files grow in future, they might need their own tests.

As for #355, it's a really simple check just before we fetch the url.

Thank you for your time!
I'd appreciate any feedback...

### DIFF
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -42,7 +42,10 @@ GH_TOKEN = os.environ.get('GH_TOKEN', '')
 
 
 def fetch_url(url):
-    """This was pointed out as a Security issue in bandit.
+    """
+    Notes
+    -----
+    This was pointed out as a Security issue in bandit.
     please look at issue #355,
     we fixed it, but the bandit warning might remain,
     need to suppress it manually (just ignore it)
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -42,6 +42,10 @@ GH_TOKEN = os.environ.get('GH_TOKEN', '')
 
 
 def fetch_url(url):
+    # This was pointed out as a Security issue in bandit.
+    # please look at issue #355,
+    # we fixed it, but the bandit warning might remain,
+    # need to suppress it manually ¯\_(ツ)_/¯.
     req = Request(url)
     if GH_TOKEN:
         req.add_header('Authorization', 'token {0}'.format(GH_TOKEN))
@@ -50,7 +54,8 @@ def fetch_url(url):
         # url = Request(url,
         #               headers={'Accept': 'application/vnd.github.v3+json',
         #                        'User-agent': 'Defined'})
-        f = urlopen(req)
+        if url.lower().startswith('http'):
+            f = urlopen(req)
     except Exception as e:
         print(e)
         print("return Empty data", file=sys.stderr)
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -42,10 +42,10 @@ GH_TOKEN = os.environ.get('GH_TOKEN', '')
 
 
 def fetch_url(url):
-    # This was pointed out as a Security issue in bandit.
-    # please look at issue #355,
-    # we fixed it, but the bandit warning might remain,
-    # need to suppress it manually ¯\_(ツ)_/¯.
+    '''This was pointed out as a Security issue in bandit.
+    please look at issue #355,
+    we fixed it, but the bandit warning might remain,
+    need to suppress it manually (just ignore it)'''
     req = Request(url)
     if GH_TOKEN:
         req.add_header('Authorization', 'token {0}'.format(GH_TOKEN))
@@ -54,8 +54,10 @@ def fetch_url(url):
         # url = Request(url,
         #               headers={'Accept': 'application/vnd.github.v3+json',
         #                        'User-agent': 'Defined'})
-        if url.lower().startswith('http'):
-            f = urlopen(req)
+        if not url.lower().startswith('http'):
+            msg = "Please make sure you use http/https connection"
+            raise ValueError(msg)
+        f = urlopen(req)
     except Exception as e:
         print(e)
         print("return Empty data", file=sys.stderr)
--- a/docs/source/ext/github_tools.py
+++ b/docs/source/ext/github_tools.py
@@ -42,10 +42,11 @@ GH_TOKEN = os.environ.get('GH_TOKEN', '')
 
 
 def fetch_url(url):
-    '''This was pointed out as a Security issue in bandit.
+    """This was pointed out as a Security issue in bandit.
     please look at issue #355,
     we fixed it, but the bandit warning might remain,
-    need to suppress it manually (just ignore it)'''
+    need to suppress it manually (just ignore it)
+    """
     req = Request(url)
     if GH_TOKEN:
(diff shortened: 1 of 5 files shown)


amitchaudhari9121:
These Checks seem to be failing cause I assumed matplotlib is installed. It's not.
I'll push another commit that solves this, soon.

amitchaudhari9121:
Uwa!!! help me please, 
The failing CI checks are not giving me any good information...

CI is important, but in my case it takes me more time to notice and rectify my mistakes that CI catches, than it takes me to do the maintenance work...  ;-;

---

even on the Previous PR, I could not pass all the CI checks...

fury/tests/test_utils.py
```diff
@@ -130,9 +130,9 @@ def trilinear_interp_numpy(input_array, indices):
     y_indices = indices[:, 1]
     z_indices = indices[:, 2]
 
-    x0 = x_indices.astype(np.integer)
-    y0 = y_indices.astype(np.integer)
-    z0 = z_indices.astype(np.integer)
+    x0 = x_indices.astype(np.int64)
+    y0 = y_indices.astype(np.int64)
+    z0 = z_indices.astype(np.int64)
```
skoudoro, line 135: `int` instead of `np.int64`. it does not need to force 64bit

fury/tests/test_convert.py
```diff
@@ -0,0 +1,35 @@
...
+    # plt.plot(names, values)
+    # plt.suptitle('Categorical Plotting')
+    arr2 = matplotlib_figure_to_numpy(fig, transparent=True)
+
+    with TemporaryDirectory() as tmpdir:
+        fname = os.path.join(tmpdir, 'tmp.png')
+        dpi = 100
+        fig.savefig(fname, dpi=dpi, transparent=True)
+        arr1 = load_image(fname)
+        npt.assert_array_equal(arr1, arr2)
```
skoudoro, line 35: Nice test!

fury/tests/test_convert.py
```diff
@@ -0,0 +1,35 @@
...
+    values = [1, 10, 100]
+
+    fig = plt.figure(figsize=(9, 3))
+    # plt.subplot(131)
+    # plt.bar(names, values)
+    # plt.subplot(132)
+    # plt.scatter(names, values)
+    # plt.subplot(133)
+    # plt.plot(names, values)
+    # plt.suptitle('Categorical Plotting')
```
skoudoro, line 27: why is it commented? your figure seems empty

fury/tests/test_convert.py
```diff
@@ -0,0 +1,35 @@
...
+    arr2 = matplotlib_figure_to_numpy(fig, transparent=True)
```
skoudoro, line 28: for more control in your test, use `transparent=False`

docs/source/ext/github_tools.py
```diff
@@ -42,6 +42,10 @@
 
 
 def fetch_url(url):
+    # This was pointed out as a Security issue in bandit.
+    # please look at issue #355,
+    # we fixed it, but the bandit warning might remain,
+    # need to suppress it manually ¯\_(ツ)_/¯.
```
skoudoro, line 48: to convert to a complete docstring with a NOTE section

docs/source/ext/github_tools.py
```diff
@@ -50,7 +54,8 @@ def fetch_url(url):
         # url = Request(url,
         #               headers={'Accept': 'application/vnd.github.v3+json',
         #                        'User-agent': 'Defined'})
-        f = urlopen(req)
+        if url.lower().startswith('http'):
+            f = urlopen(req)
```
skoudoro, line 58: What happens if it does not start with http? here a proposition:

```python
if not url.lower().startswith('http'):
    msg = "custom message"
    raise ValueError(msg)
f= urlopen(req)
```

The `ValueError` will need to be tested with `npt.assert_raises`

skoudoro (commented):
Hi @amitchaudhari9121,

Thank you for fixing these 2 issues! overall, it looks good. It just needs some reshaping.

Thank you

skoudoro:
concerning the tests, no worries. We will need to make some of them more robust

fury/tests/test_convert.py
amitchaudhari9121: Thanks! 

I've (kindof) picked it up from the code you wrote 😅 
This just tests the exact way its written in test_actors.py 

But it's not redundant code. if something breaks in future. (or we add more conversion definitions) this will be useful...

fury/tests/test_convert.py
amitchaudhari9121: The test failed for some reason when I added these args,  

I think I could just test an empty figure, and confirm it works.
will that be the case? 

I'm not sure...

I don't understand enough about matplotlib, so I can't reliably say why it fails when I uncomment it... 
I'll keep researching & hopefully find a fix.

Any suggestions? any idea why this might be failing? any guess is appreciated...

docs/source/ext/github_tools.py
amitchaudhari9121: Nice!
It didn't even cross my mind !!!
I will change in next commit, thank you for this.

docs/source/ext/github_tools.py
```diff
@@ -42,6 +42,10 @@
 
 
 def fetch_url(url):
+    '''This was pointed out as a Security issue in bandit.
+    please look at issue #355,
+    we fixed it, but the bandit warning might remain,
+    need to suppress it manually (just ignore it)'''
```
skoudoro, line 48: The docstring is incomplete. Please, use double-quotes and follow the standard of our docstring. As an example, you can look [here](https://github.com/fury-gl/fury/blob/master/fury/colormap.py#L302-L324)

fury/tests/test_convert.py
skoudoro: See below the test updated. I admit that it was tricky., you had to update the `matplotlib_figure_to_numpy` option and the `fig.savefig` options to make sure that both images have the same size and orientation.

```python
    fig = plt.figure(figsize=(9, 3))
    plt.subplot(131)
    plt.bar(names, values)
    plt.subplot(132)
    plt.scatter(names, values)
    plt.subplot(133)
    plt.plot(names, values)
    plt.suptitle('Categorical Plotting')
    arr2 = matplotlib_figure_to_numpy(fig, transparent=False,
                                      flip_up_down=False)

    with TemporaryDirectory() as tmpdir:
        fname = os.path.join(tmpdir, 'tmp.png')
        fig.savefig(fname, transparent=False, bbox_inches='tight',
                    pad_inches=0)
        arr1 = load_image(fname)
        npt.assert_array_equal(arr1, arr2)
```

skoudoro (commented):
Hi @amitchaudhari9121,

See below for some suggestions to update your test. After those update, I think your PR will be ready to go.
Thanks!

docs/source/ext/github_tools.py
skoudoro: You did not address my comment above. Please, look at [the example](https://github.com/fury-gl/fury/blob/master/fury/colormap.py#L302-L324) cited above and complete the docstring. 
As suggested, the text should be in a `NOTE`

skoudoro:
Thank you for working on this @amitchaudhari9121. merging

PR #408, opened 2021-04-06
Updating the Missing Parenthesis

ganimtron-10:
For the Operating System and Version Info on Line 21, the command `print`  didn't had the closing parenthesis, so adding that as without that it throws an error while executing.


skoudoro:
good catch, thanks! merging

PR #410, opened 2021-04-07
added BulletList to UI

antrikshmisri:
This PR adds `BulletList` to the `UI` module. Below is an image for reference:-

card_ui

Thanks!


antrikshmisri:
@skoudoro I can't figure out why codecov is not counting tests for some functions. Is there something wrong with my tests?

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+        Parameters
+        ----------
+        node: str
+            Text to display after the bullet
+        bullet_radius: int
+            Radius of the bullet
+        indent: int
+            Indentation of the bullet point
+        draggable: Bool
+            If the card should be draggable
```
Nibba2018, line 5488: ~card~ `UI`

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+            Color of the text
+        """
+        self.content_size = size
+        self.bullet_radius = bullet_radius
+        self.indent = 2 * self.bullet_radius + indent
+        self.draggable = draggable
+        self.bg_color = bg_color
+        self.bg_opacity = bg_opacity
+        self.bullet_color = bullet_color
+        self.text_color = text_color
```
Nibba2018, line 5510: You should add these after calling the super constructor and implement the respective getter and setter methods

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+        self.nodes = []
+        self.dict = {}
+        self.content_size = size
+        self.indent = np.clip(indent, 10, int(self.content_size[0]/2))
+        self._last_child = None
+        self.draggable = draggable
+        self.bg_color = bg_color
+        self.bg_opacity = bg_opacity
+        self.text_color = text_color
+        self.bullet_bg_color = bullet_bg_color
```
Nibba2018, line 5291: add these after calling super constructor

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+        self.panel.position = _coords
+
+    def _get_size(self):
+        self.panel.size
+
+    def create_list(self, lst, head=None):
+        """Create a flat list with all nodes ordered sequentially
+        """
+        for ele in lst:
+            if isinstance(ele, list):
```
Nibba2018, line 5360: ~ele~ `element`

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+    def __init__(self, node="", bullet_radius=5,
+                 indent=5, draggable=False, size=(300, 20),
+                 position=(0, 0), bg_color=(1, 1, 1),
+                 bg_opacity=1, bullet_color=(0, 0, 0),
+                 text_color=(0, 0, 0)):
+        """
+        Parameters
+        ----------
+        node: str
+            Text to display after the bullet
```
Nibba2018, line 5482: I think you should rename this to something else or it will be confusing. Something like `bullet_label` or just `label` works.

docs/tutorials/02_ui/viz_bullet_list.py
```diff
@@ -0,0 +1,78 @@
...
+# manager.
+
+current_size = (1000, 1000)
+show_manager = window.ShowManager(size=current_size,
+                                  title="FURY Bullet List Example")
+
+show_manager.scene.add(panel)
+
+# To interact with the UI, set interactive = True
+interactive = True
```
Nibba2018, line 72: ~True~ `False`

fury/ui.py
```diff
@@ -5240,3 +5240,387 @@ def _set_position(self, coords):
...
+            UI component
+        size : (int, int)
+            Width and height of the pixels of this UI component.
+        bg_color: (float, float, float)
+            Background color of card
+        bg_opacity: float
+            Background opacity
+        """
+
+        self.points = points
```
Nibba2018, line 5281: You need to be extra careful here.
While assigning a list directly you are actually passing the reference and not the content, so any external changes to the original list will be reflected in your list as well.

I tried manipulating the original list externally and it seems your UI is unable to either throw an exception or update the UI.

Nibba2018 (commented):
Hey @antrikshmisri , Thank you for creating this PR. Here's a quick review. I will review it later in depth.

fury/ui.py
antrikshmisri: I'll pass the content of the original list instead of its reference, that's not a problem but updating the UI if original list is changed is a bit complicated. This is because how the UI element works. The UI has two lists, `self.points` (this list contains all the elements of the original list as is), `self.nodes` (this is a flat list that contains all the elements of the original list converted to `BulletPoint2D`). So, if I try to manipulate the original list this will not impact the UI as the UI element depends on `self.nodes` and not `self.points`. 
The only way of updating the UI when original list changes that I can think of is:-

1. check if original list and `self.points` are unequal right before adding the UI element to the scene
2. if they are unequal update `self.points` as well as `self.nodes`

Does this sound right?

fury/ui.py
antrikshmisri: Instead of these variables can I do something like this:-
```python
        super(BulletPoint2D, self).__init__()
        self.panel.color = bg_color
        self.panel.opacity = bg_opacity
        self.bullet.color = bullet_color
        self.bullet_textbox.color = text_color
        self.position = position
```

fury/ui.py
antrikshmisri: @Nibba2018 I think I need to add a custom event listener that listens for changes in any specific property in the UI element, as I couldn't find any information about this in vtk docs/examples. Do you think this is viable, if so where should it go, inside the interactor module or inside the base UI class as a `static` method?

skoudoro:
I suppose this PR can be close in favor of your future TreeUI PR, right ?  or you will just update this one @antrikshmisri ?

antrikshmisri:
> I suppose this PR can be close in favor of your future TreeUI PR, right ? or you will just update this one @antrikshmisri ?

Hey @skoudoro, I think once the Tree2D widget is merged, this widget can be updated to inherit from the Tree2D class as they share a lot of similarities. But the update would require me to remove all the existing logic/code from this PR. So, i think it would make sense to close this PR.

skoudoro:
Ok, thanks for the feedback, closing

PR #425, opened 2021-04-28
WIP: Cube Axes Actor.

amit-chaudhari1:
I wish to add an cube axis actor in vtk.
---

## What is the Problem?
The problem is the lack of an axes actor, with scaling.

## Why this solution?
It will help in Data Science Visualizations, where you need a visual "scale" to compare the data to.//
This will also be helpful in Engineering Visualizations. Where have a scale to compare against is crucial.//

## How is the problem solved?
I've used vtk's cube axis actor. [Here's the link to docs for cube axis actor](https://vtk.org/doc/nightly/html/classvtkCubeAxesActor.html)

WIP: There's Stuff left to do, so please hold on your review till I finish.
- [X] Implement a Basic CubeAxisActor.
- [X] Create Docs below the function.
- [ ] Improve the scale of the axis labels and the axis titles.
- [ ] Write tests? 
- [ ] Improvements to the implementation.
- [ ] CI fixes?

Example:
---
Screenshot 2021-04-28 at 9 19 05 AM

__Please let me know how can I improve this? am i missing some stuff that'll be needed to do?__

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -787,13 +787,73 @@ def scalar_bar(lookup_table=None, title=" "):
     return scalar_bar
 
 
-def cube_axis_actor(scene,bounds,label_X_color=[1,0,0], label_Y_color=[0,1,0], label_Z_color=[0,0,1],
-                    title_X="X-Axis",title_Y="Y-Axis",title_Z="Z-Axis",
-                    draw_grid_lines=True,minor_tick_visibility=False):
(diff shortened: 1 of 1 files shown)


amitchaudhari9121:
@skoudoro 
Should i create an example about its use case?
Also, I'd like to request a review from you, when you are free.

amitchaudhari9121:
Also, Should I write tests and  separate example tutorial for this?
It looks pretty Self explaining to me. 

I've even added a short example usage in the function description itself.

skoudoro (commented):
Thank you for doing this. I need to look deeper at the code. 

And yes, tests are always needed. for Documentation, it depends. in this case, it is not necessary

PR #426, opened 2021-04-28
Allows to define the priority of a shader_callback and obtain the vtkEventId

devmessias:
The AddObserver method of vtk returns an unsigned int (eventId) as discussed in
[https://vtk.org/doc/nightly/html/classvtkObject.html#aa9f4bcd0e0052a2ea4347f62ed0a5156
](https://vtk.org/doc/nightly/html/classvtkObject.html#aa9f4bcd0e0052a2ea4347f62ed0a5156)
This eventId tag can be used later to remove the observer associated with the vtk actor. 

With this pull-request, it'll be possible to remove an observer (created using the add_shader_callback method) associated with a vtk actor. In addition, it will also be possible to control the order of the shader's callbacks.

This behavior gives some improvements with uniform variables and avoids to create an infinitude of observers if we want to change the callback functions. This can be useful when we are dealing with animations using shader actors.

```python
mapper = nodes_actor.GetMapper()
id_observer = None
def timer_callback(obj, event):
    global  id_observer
    def callback(_caller, _event, calldata=None):
        program = calldata
        if program is not None:
            print(f'call observer number: {cnt}')
            program.SetUniformf('uniform_name',  cnt)

    if id_observer is not None:
        mapper.RemoveObserver(id_observer)
        
    id_observer = add_shader_callback(
            nodes_actor, callback)

    cnt = next(counter)

   ```
   Without removing the older vtkObservers, the number of shader callbacks will grow indefinitely

### CODE REVIEW

File: fury/shaders/base.py [hunk 1]
```diff
@@ -135,14 +135,41 @@ def add_shader_callback(actor, callback):
     callback : callable
         function or class that contains 3 parameters: caller, event, calldata.
         This callback will be trigger at each `UpdateShaderEvent` event.
+    priority : float
```
skoudoro, line 138: `float, optional` instead

fury/shaders/base.py
```diff
@@ -135,14 +135,41 @@ def add_shader_callback(actor, callback):
...
+         Commands with a higher priority are called first.
+
+    Returns:
+    --------
+        id_observer : int
+            An unsigned Int tag which can be used later to remove the event
+            or retrieve the vtkCommand used in the observer.
+            See more at: https://vtk.org/doc/nightly/html/classvtkObject.html
+
+    Example
```
skoudoro, line 148:
```
Examples
----------
```

fury/shaders/base.py
```diff
@@ -135,14 +135,41 @@ def add_shader_callback(actor, callback):
...
+
+    ```python
+    add_shader_callback(actor, func_call1)
```
skoudoro, line 151: Can you add the priority in your example?

fury/shaders/base.py
```diff
@@ -135,14 +135,41 @@ def add_shader_callback(actor, callback):
...
+    id_observer = add_shader_callback(actor, func_call2)
+    actor.GetMapper().RemoveObserver(id_observer)
+    ```
 
     """
     @vtk.calldata_type(vtk.VTK_OBJECT)
     def cbk(caller, event, calldata=None):
         callback(caller, event, calldata)
 
+    if isinstance(priority, (float, int)) is False:
```
skoudoro, line 161: `not isinstance(priority, (float, int))` instead

fury/shaders/base.py
```diff
@@ -135,14 +135,41 @@ def add_shader_callback(actor, callback):
...
+        # This avoid the strange error checking from vtk
+        # mapper.AddObserver(vtk.vtkCommand.UpdateShaderEvent, cbk, '12')
+        # TypeError: AddObserver argument 1: string or None required
```
skoudoro, line 164: you can remove the comment. the `if` above is explicit and the comment is more maintenance in case of a change 😄

fury/tests/test_shaders.py
```diff
@@ -16,3 +18,28 @@ def test_load():
...
+
+    def callback(_caller, _event, calldata=None):
+        program = calldata
+        if program is not None:
+            pass
+
+    id_observer = fs.add_shader_callback(
+            actor, callback, 42)
+
+    # print('\t...invalid priority type should create an exception')
```
skoudoro, line 39: can you remove this comment? Thanks!

fury/tests/test_shaders.py
```diff
@@ -16,3 +18,28 @@ def test_load():
...
+    with pytest.raises(Exception):
+        fs.add_shader_callback(actor, callback, priority='str')
+
+    mapper = actor.GetMapper()
+    if id_observer is not None:
+        mapper.RemoveObserver(id_observer)
```
skoudoro, line 45: Can you add a last test to check if the observer is really removed? 
Thanks!

skoudoro (commented):
Hi @devmessias,

Thank you for this useful PR. Overall, it looks good to me. See below some comments and then, it will be ready to go.

Thanks!

devmessias:
Thanks for all the suggestions @skoudoro!

skoudoro:
All the failing tests are not related so I am going ahead and merge this PR.

Thanks @devmessias

Issue #429, opened 2021-05-01
Colormap not working as intended with surface actor


SunTzunami:
## Description
 Colormap is not being mapped to the actor when generating a surface using surface actor. Colormap helps in visualizing elevation. 

## Code example 

```
import numpy as np
from fury import window, actor, colormap

## grid
x = np.linspace(-1, 1, 100)
y = np.linspace(-1, 1, 100)
X, Y = np.meshgrid(x, y)
X = X.reshape(-1)
Y = Y.reshape(-1)

## elevation
Z = 0.48*np.sin(2*np.pi*(X))*np.cos(2*np.pi*(Y))
xyz = np.vstack([X, Y, Z]).T

## colormap
v = np.copy(Z)
v /= np.max(np.abs(v),axis=0)
colors = colormap.create_colormap(v, name='spring')

surface_actor = actor.surface(xyz, colors=colors)

scene = window.Scene()
scene.background((1, 1, 1))
scene.zoom(10.5)
scene.set_camera(position=(-32, 50, 60), focal_point=(0.0, 0.0, 0.0), view_up=(0.34, -0.630, 0.68))
scene.add(surface_actor)
window.show(scene, size=(400, 400))
```
## Preview of the output 

![elevation](https://user-images.githubusercontent.com/65067354/116779575-aaa8c100-aa94-11eb-92a1-415d23c18ff6.png)

As can be seen in the output, the actor returned by the surface actor doesn't have the correct color.

amitchaudhari9121:
This should be fixed once the above PR gets merged. 

For future readers:
---
The method expected a 0-255  ranged values for color input.

but the `colors = colormap.create_colormap(v, name='spring')` would generate a 0-1 ranged values.
this resulted in the actor not displaying the color map properly.
Fix was to change the actor definition a bit.

PR #362, opened 2021-01-24
Animated Surfaces

SunTzunami:
Fixes #324 

An example to illustrate how one can render a time-varying 2D function using FURY.  I have added quite a few comments to make the code easy to interpret. 

Preview-

2d_func

PR #430, opened 2021-05-01
Surface actor colormap fix

amitchaudhari9121:
Hi! 
This is a fix for issue #429 .
@SunTzunami and I hopped on a call and fixed it together.

### Now

Screenshot 2021-05-01 at 6 42 31 PM

I think, this is exactly how It should look...
I've also changed the documentation to include the range of the Individual color values. 
In case someone else is confused due to this.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -298,7 +298,7 @@ def surface(vertices, faces=None, colors=None, smooth=None, subdivision=3):
 
     if colors is not None:
         triangle_poly_data.GetPointData().\
-            SetScalars(numpy_support.numpy_to_vtk_colors(colors))
+            SetScalars(numpy_support.numpy_to_vtk_colors(255 * colors))
 
     if faces is None:
         tri = Delaunay(vertices[:, [0, 1]])
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -298,7 +298,7 @@ def surface(vertices, faces=None, colors=None, smooth=None, subdivision=3):
 
     if colors is not None:
         triangle_poly_data.GetPointData().\
-            SetScalars(numpy_support.numpy_to_vtk_colors(255 * colors))
+            SetScalars(numpy_to_vtk_colors(colors))
 
     if faces is None:
         tri = Delaunay(vertices[:, [0, 1]])
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -15,7 +15,7 @@ from fury.utils import (lines_to_vtk_polydata, set_input, apply_affine,
                         set_polydata_vertices, set_polydata_triangles,
                         shallow_copy, rgb_to_vtk, numpy_to_vtk_matrix,
                         repeat_sources, get_actor_from_primitive,
-                        fix_winding_order)
+                        fix_winding_order, numpy_to_vtk_colors)
 from fury.io import load_image
 from fury.actors.odf_slicer import OdfSlicerActor
 import fury.primitive as fp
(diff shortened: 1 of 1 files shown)


SunTzunami:
@amitchaudhari9121 Thanks for the PR. 
And yes that's how it's supposed to look!!

skoudoro (commented):
Hi @amitchaudhari9121,

Thank you for this fix. As usual, it needs tests to avoid reproducing this issue.

Furthermore, we need to check what is the rules.  Are the other actors expecting 0-1 or 0-255. we should make sure that is uniform everywhere.

amitchaudhari9121:
>Furthermore, we need to check what is the rules. Are the other actors expecting 0-1 or 0-255. we should make sure that is uniform everywhere.

I've skimmed `actor.py` through all mentions of color, and yes. 
It is the norm in all of actor.py.

>As usual, it needs tests to avoid reproducing this issue.

I'm not sure on how do i go about testing it.
I'd like suggestions about this.

Do we need to test it? 
Like nothing stops working if you don't give it a proper input. ( I tried giving it 255, thinking it would crash, but it just rendered with a wrong colormap )
Even If I go above the limit of 1, It seems to work fine, Just not display color properly.

SunTzunami:
I think you could raise an error when the user inputs a color that's out of bounds. @skoudoro  will be able to guide better.

skoudoro:
We might let pass this PR and create a specific PR for all actors to raise an error if a color is out of bounds. Still thinking about it 🤔 ...

amitchaudhari9121:
Ok 👌
Thank you!

guaje:
@amitchaudhari9121, @skoudoro, and @SunTzunami.

This is blocking #362. Is any of you working on the "check out of bounds" PR or what are the steps to merge/close this and #362?

SunTzunami:
@guaje once this is done, #362 will work as intended and I'll update it to be merged. 
As for the out of bounds, I did a quick scan of [actor.py](https://github.com/fury-gl/fury/blob/master/fury/actor.py). The following functions seem to take input range from 0 to 255 -
1. [slicer](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L31)
2. [odf_slicer](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L852)

Unsure about - 

3. [tensor_slicer](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L915) as the range of acceptable colors is not clearly mentioned but the docstring mentions [RGB input](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L938).
4.  [_tensor_slicer_mapper](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1000) as the range of acceptable colors is not clearly mentioned but the docstring mentions [RGB input](https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1021).

@skoudoro and @amitchaudhari9121

amitchaudhari9121:
@guaje I can't find the time to work for fury for the moment, I'm busy with my intern at Linux foundation, 
Could you please explain how this is blocking #362 ?

As far as i remember (its been weeks since those commits), going out of bounds just messes up the colormap displayed. 
which is hard to notice at times. 

Personally, I don't think we need to add tests for it, it's mentioned clearly in the doc for that function, and any wrong input does not crash anything. We could merge this as is, and not have any problems.

@skoudoro  @SunTzunami  your thoughts?

SunTzunami:
@amitchaudhari9121 for #362 to work optimally, colormap need to work with surface actor (which was not the case before this PR). Once this PR is merged, colormaps will work with surface actor and #362 will work.

skoudoro:
Hi All, 

This PR can be merged so I will go ahead. However, we need to clarify the rules concerning the color ([0-1], vs [0-255]).

Let's create an issue to remember that we need to handle the color in all actors. I suppose, there is a function for that in utils.py or we need to create it.

skoudoro:
Thanks @amitchaudhari9121 for this. Merging!

PR #411, opened 2021-04-07
Added Layout test file

jhalak27:
For issue #403 

Added test file for fury/layout.py in fury/tests/test_layout.py

PR #434, opened 2021-05-27
WIP: added tests for layout module

antrikshmisri:
This PR adds tests for the layout module. This PR extends/modifies the PR opened by @jhalak27 ( #411 ).
Along with the tests a `get_default_cubes` method is also implemented that provides default parameter cube actors to test the grid with. The `GridLayout`  as well as the `Layout` classes are tested with these default cubes as it makes the visualization of the grid behavior easier.
 
daigonal_grid
Diagonal Grid
square/rect grid
Square/Rect Grid

### DIFF
--- a/fury/tests/test_layout.py
+++ b/fury/tests/test_layout.py
@@ -8,7 +8,7 @@ def get_default_cubes(centers=np.asarray([[[0, 0, 0]], [[5, 5, 5]]]),
                       directions=np.asarray([[[0, 0, 0]], [[0, 0, 0]]]),
                       colors=np.random.rand(2, 3), scales=[1, 1.5]):
     """Provides cube actors with default parameters
-    
+
     Parameters
     ----------
     centers: ndarray, shape (2, 3)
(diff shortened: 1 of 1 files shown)


skoudoro (commented):
Hi @antrikshmisri,

Thank you for doing this. Your tests look good. Could you also tests:

- [this line](https://github.com/fury-gl/fury/blob/master/fury/layout.py#L94) with `npt.assert_raises`
- Add a tests with some UI. add 2 sliders and check if it works as expected. Thank you!

skoudoro:
Also, there is one pep8. Look above and do not forget to fix it. 

Thanks!

antrikshmisri:
> Hi @antrikshmisri,
> 
> Thank you for doing this. Your tests look good. Could you also tests:
> 
> * [this line](https://github.com/fury-gl/fury/blob/master/fury/layout.py#L94) with `npt.assert_raises`
> * Add a tests with some UI. add 2 sliders and check if it works as expected. Thank you!



> Hi @antrikshmisri,
> 
> Thank you for doing this. Your tests look good. Could you also tests:
> 
> * [this line](https://github.com/fury-gl/fury/blob/master/fury/layout.py#L94) with `npt.assert_raises`
> * Add a tests with some UI. add 2 sliders and check if it works as expected. Thank you!

I tested the GridLayout with different UI's like `LineSlider2D`, `Panel2D` and `Rectangle2D` but there are a couple of problems. Firstly, not every UI element has an actor attribute that directly represents the specific UI element in terms of a `vtkActor2D`, like in `Button2D` or `Rectangle2D` as seen [here](https://github.com/fury-gl/fury/blob/2e7971a176c2540e10a9a6da861097583d08cb4a/fury/ui.py#L494). Hence, calculating the bounds of the UI element by  [get_bounding_box_sizes](https://github.com/fury-gl/fury/blob/2e7971a176c2540e10a9a6da861097583d08cb4a/fury/utils.py#L754) method is not possible. Secondly, even the UI elements that have the actor attribute return `None` with `GetBounds()` method for some reason. I guess bounds of a UI element need to be calculated differently than that of the regular actors.

skoudoro (approved):
ok, it makes sense. Thank you for the feedback. So this should be added to the general layout management discussion. 

Either we need to change the name if we have a layout for UI and another one for actors, or we need to improve the integrations of both worlds. 

Otherwise, this PR looks good to me.
 
Thank you, merging

Issue #435, opened 2021-05-27
Fury/VTK Streaming: webrtc/rtmp


devmessias:
As well know VTK provides a set of tools to perform the streaming of the rendered result. This streaming it's expensive, because uses jpg's images consumed as a base64 in the browser. In addition, sometimes we just want to stream the rendering result, with no interaction, to a large audience

@filipinascimento and I have been discussing how to perform this streaming which I need for my PhD and for the Summer of Code project. 

We already have some results to achieve this through WebRTC and opengl buffers. But, we need to define how to organize the code. 

I think we can create an object inside the fury/window.py which will be responsible to get the buffer (FuryStreamClient). 
Otherwise, maybe the Fury Stream Server it's more suitable for a different repo, because it has no dependency with fury.

My proposal it's to have three ways to perform the communication between FuryStreamClient and FuryStreamServer
## Communication using pipe
```bash
$ python viz_example.py | furystream --webrtc --youtube --youtube_key=.... --webrtc_port=8888
````

```python
# example.py
from fury.window import FuryStreamClient
...
ms = 16 # ms refresh rate, 0 means a renderevent observer isntead of timer
stream = FuryStreamClient(
    showm, True image_buffer=None, info_buffer=None, broker_url=None)
stream.init(ms,)  
....
showm.start()
```
### Pros
- Simple
- good performance
- ???
### Cons
- non-agnostic
- works only in the same machine
- ???
## Communication using a broker
```bash
$ furystream  --broker --webrtc --youtube --youtube_key=.... --webrtc_port=8888 --broker_port=5555
$ python viz_example.py
````
```python
# example.py
 from fury.window import FuryStreamClient
 ...
ms = 16 # ms refresh rate, 0 means a renderevent observer isntead of timer
stream = FuryStreamClient(
    showm, False, image_buffer=None, info_buffer=None, broker_url=broker_url)
stream.init(ms,)  
....
showm.start()
```
### Pros
- Simple
- Can distribute the computation results into different machines. 
- Agnostic (different from pipe)
-  ???
### Cons
- Depends of another protocol and lib (zmq)
- Slow than  pipe and shared memory approaches
- ???

## 3  Communication using shared memory and forks 
```bash
$ python viz_example.py
````
```python
# example.py
from fury.window import FuryStreamClient
from fury_stream_server import furry_webrtc, furry_youtube
...
process_wrtc = multiprocessing.Process(target=furry_webrtc, args=(image_buffer, info_buffer))
[...]

PR #437, opened 2021-05-27
FURY Streaming System Proposal

devmessias:
related with #435


fury/stream/README.md
```diff
@@ -0,0 +1,38 @@
+
+## TODO
```
Garyfallidis, line 2: This should go in the description of the pull request.

docs/tutorials/04_stream/viz_interaction.py
```diff
@@ -0,0 +1,135 @@
...
+        showm.start()
+    p.kill()
+    stream.stop()
+    stream_interaction.stop()
+    stream.cleanup()
+    stream_interaction.cleanup()
+    # open a browser using the following the url
+    # http://localhost:8000/
+
+window.record(
```
skoudoro, line 134: why this indentation as the end?

docs/tutorials/04_stream/viz_no_interaction.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    p.start()
+
+    stream.start(ms,)
+    if interactive:
+        showm.start()
+    stream.stop()
+    stream.cleanup()
+
+window.record(
```
skoudoro, line 110: why this indentation as the end?

docs/tutorials/04_stream/viz_widget.py
```diff
@@ -0,0 +1,45 @@
+################################################
+#
+"""
+=======================================================
+Streaming FURY with WebRTC/MJPEG using the Widget Object
+========================================================
+
```
skoudoro, line 7: the tutorial needs more explanation

docs/tutorials/04_stream/viz_no_interaction.py
```diff
@@ -0,0 +1,111 @@
+"""
+====================================
+Streaming FURY with WebRTC/MJPEG
+====================================
+
```
skoudoro, line 5: the tutorial needs more explanation

docs/tutorials/04_stream/viz_interaction.py
```diff
@@ -0,0 +1,135 @@
+"""
+====================================
+Streaming FURY with user interaction
+====================================
+
```
skoudoro, line 5: the tutorial needs more explanation

fury/stream/client.py
```diff
@@ -0,0 +1,362 @@
...
+
+        w, h, _ = vtk_image.GetDimensions()
+        np_arr = np.frombuffer(vtk_array, dtype='uint8')
+        if np_arr is not None:
+            stream_client.img_manager.write_into(w, h, np_arr)
+            # time.sleep(kwargs['ms']/1000)
+            stream_client._in_request = False
+            if kwargs['force_render']:
+                stream_client.showm.window.Render()
+                stream_client.showm.iren.Render()
```
skoudoro, line 47: in python, to avoid unnecessary indentation, you should inverse the logic,  see below:
```python
if stream_client._in_request:
    return
stream_client._in_request = True
stream_client.window2image_filter.Update()
.....
if np_arr is None:
   return
stream_client.img_manager.write_into(w, h, np_arr)
....
```

fury/stream/client.py
```diff
@@ -0,0 +1,362 @@
...
+
+class FuryStreamClient:
+    def __init__(
+            self, showm,
+            max_window_size=None,
+            use_raw_array=True,
+            whithout_iren_start=False,
+            num_buffers=2,
+    ):
+        '''This obj is responsible to create a StreamClient.
```
skoudoro, line 58: double quote instead of single quote

fury/stream/client.py
```diff
@@ -0,0 +1,362 @@
...
+                print(f'Shared Memory {self.img_manager.info_buffer_name}\
+                        (info_buffer) File not found')
+            for buffer, name in zip(
+                    self.img_manager.image_buffers,
+                    self.img_manager.image_buffer_names):
+                buffer.close()
+                try:
+                    buffer.unlink()
+                except FileNotFoundError:
+                    print(f'Shared Memory {name}(buffer image) File not found')
```
skoudoro, line 195: same comment as above

fury/stream/client.py
```diff
@@ -0,0 +1,362 @@
...
+    showm : ShowmManager
+    iren : vtkInteractor
+    render_after : bool, optional
+        If the render method should be called after an
+        dequeue
+
+    """
+    ts = time.time()*1000
+    data = circular_queue.dequeue()
+    if data is not None:
```
skoudoro, line 214: same comment as above

requirements/default.txt
```diff
@@ -2,4 +2,6 @@
 numpy>=1.7.1
 vtk>=8.1.2,!=9.0.0
 scipy>=1.2.0
-pillow>=5.4.1
+Pillow>=8.0.1
+aiohttp>=3.7.4
+aiortc
```
skoudoro, line 7: should we make this module optional to the user?

skoudoro (commented):
Hi @devmessias,

Thank you for this excellent new module. Here are some of my issues:

- The documentation generation does not work anymore.
- it would be good to provide more textual information/description in each tutorial
- There are no tests. Too much code not tested. 
- Many functions are not documented.

See below for some other comments.

Please, can you do a sprint on this issue. thank you.
@filipinascimento, Can you help @devmessias to finalize this module. Thank you

docs/tutorials/04_stream/viz_interaction.py
devmessias: > Hi @devmessias,
> 
> Thank you for this excellent new module. Here are some of my issues:
> 
> * The documentation generation does not work anymore.
> * it would be good to provide more textual information/description in each tutorial
> * There are no tests. Too much code not tested.
> * Many functions are not documented.
> 
> See below for some other comments.
> 
> Please, can you do a sprint on this issue. thank you.
> @filipinascimento, Can you help @devmessias to finalize this module. Thank you

I'll try to cover more cases inside of widget.py and server.py. However, I couldn't find a good way to cover some specific parts of the streaming system

### 1 Optional modules and python version
image

opencv and Cython are not mandatory. Because these libs just provides a better performance for some specific tasks. However, this approach has an issue: pytest will cover the methods based on the installed packages.

For example, this part of the code.
image

Maybe this is a solution: https://coverage.readthedocs.io/en/latest/cmd.html#combining-data-files-coverage-combine
However we need to change the github action to perform multiple tests with different requirements (and python versions) and then use *coverage combine* to join all the reports

### 2 aiohttp endpoints and webserver
I need to discuss that with @filipinascimento
image

We have a problem because the usual flow for the streaming system is that the aiohttp server need to be created in a unique process. Therefore, we will not capture anything running pytest. In the first version of #437  was possible to create the webserver in the main thread. However, @filipinascimento and I have agreed to remove this behavior. Maybe we need to get this behavior back.

### 3 Python core bug
https://bugs.python.org/issue38119 
image

image

This is more problematic. This `try exception` was not caught by pytest because the exception is called only if something goes wrong with the  semaphores and shared memory. Maybe I can force a memory error to test this. But I don't know if it's a good idea. I can also add  `# pragma: no cover` to  avoid those lines being added in the coverage report.

For example,  I've used  `# pragma: no cover`  inside of the abstractmethods 
image

devmessias:
> Hi @devmessias,
> 
> Thank you for this excellent new module. Here are some of my issues:
> 
> * The documentation generation does not work anymore.
> * it would be good to provide more textual information/description in each tutorial
> * There are no tests. Too much code not tested.
> * Many functions are not documented.
> 
> See below for some other comments.
> 
> Please, can you do a sprint on this issue. thank you.
> @filipinascimento, Can you help @devmessias to finalize this module. Thank you

> Hi @devmessias,
> 
> Thank you for this excellent new module. Here are some of my issues:
> 
> * The documentation generation does not work anymore.
> * it would be good to provide more textual information/description in each tutorial
> * There are no tests. Too much code not tested.
> * Many functions are not documented.
> 
> See below for some other comments.
> 
> Please, can you do a sprint on this issue. thank you.
> @filipinascimento, Can you help @devmessias to finalize this module. Thank you

I'll try to cover more cases inside of widget.py and server.py. However, I couldn't find a good way to cover some specific parts of the streaming system

### 1 Optional modules and python version
image

opencv and Cython are not mandatory. Because these libs just provides a better performance for some specific tasks. However, this approach has an issue: pytest will cover the methods based on the installed packages.

For example, this part of the code.
image

Maybe this is a solution: https://coverage.readthedocs.io/en/latest/cmd.html#combining-data-files-coverage-combine
However we need to change the github action to perform multiple tests with different requirements (and python versions) and then use *coverage combine* to join all the reports

### 2 aiohttp endpoints and webserver
I need to discuss that with @filipinascimento
image

We have a problem because the usual flow for the streaming system is that the aiohttp server need to be created in a unique process. Therefore, we will not capture anything running pytest. In the first version of #437  was possible to create the webserver in the main thread. However, @filipinascimento and I have agreed to remove this behavior. Maybe we need to get this behavior back.

### 3 Python core bug
https://bugs.python.org/issue38119 
image

image

This is more problematic. This `try exception` was not caught by pytest because the exception is called only if something goes wrong with the  semaphores and shared memory. Maybe I can force a memory error to test this. But I don't know if it's a good idea. I can also add  `# pragma: no cover` to  avoid those lines being added in the coverage report.

For example,  I've used  `# pragma: no cover`  inside of the abstractmethods 
image

requirements/default.txt
devmessias: Maybe just webrtc protocol which depends on aiortc and ffmpeg. Although  the WebRTC streaming is the best way (or the only way) to share the streaming in a low-budget scenario like using ngrok.

The MJPEG streaming doesn't depend on anything special except aiohttp which is a very stable library. Thus I don't see why aiohttp and the MJPEG should be an optional module.

devmessias:
percentage covered by tests until now. 
image

skoudoro:
Hi @devmessias, 

Thank you very much for addressing most of the comment and providing explanation. I will get back to you ASAP.

skoudoro:
Also, it would be good if @filipinascimento could provide a review. Let us know. Thanks !

filipinascimento (changes requested):
Widget demo do not work on windows. May also need fixes for the timers.

docs/tutorials/04_stream/viz_interaction.py
filipinascimento: If we run this tutorial without aiortc installed, it runs but the link displayed is for webrtc and not mjpeg. If webrtc is not available the link should point out to mjpeg encoding instead.

devmessias:
> Widget demo do not work on windows. May also need fixes for the timers.

Fixed by the following commits: 36ed049,  183badd and 91f3256

docs/tutorials/04_stream/viz_interaction.py
```diff
@@ -0,0 +1,175 @@
...
+You can choose between two differnet encodings: WebRTC or MJPEG.
+WebRTC is a more robust option and can be used to perform
+a live streaming with a low-latency connection for exampel using 
+ngrok. However, to use webRTC you need to install the aiortc library.
+
+.. code-block:: bash
+    pip install aiortc
+
+In addition, if you don't have ffmpeg installed, you need
```
Garyfallidis, line 17: Do you need ffmpeg on Linux too?

docs/tutorials/04_stream/viz_interaction.py
```diff
@@ -0,0 +1,175 @@
...
+.. code-block:: bash
+    pip install aiortc
+
+In addition, if you don't have ffmpeg installed, you need 
+to install it.
+
+Linux
+
+
+`apt install libavdevice-dev libavfilter-dev libopus-dev libvpx-dev pkg-config`
```
Garyfallidis, line 23: I do not see ffmpeg here. Is that okay?

docs/tutorials/04_stream/viz_interaction.py
```diff
@@ -0,0 +1,175 @@
...
+
+    if use_raw_array:
+        p = multiprocessing.Process(
+            target=web_server,
+            args=(
+                stream.img_manager.image_buffers,
+                None,
+                stream.img_manager.info_buffer,
+                None,
+                stream_interaction.circular_queue.head_tail_buffer,
```
Garyfallidis, line 122: The args can be simplified very much by just providing as input ``stream`` and ``stream_interaction``. Then you can grab the other properties from the inside your function.

docs/tutorials/04_stream/viz_widget.py
```diff
@@ -0,0 +1,88 @@
+"""
+========================================================
+Streaming FURY with WebRTC/MJPEG using the Widget Object
+========================================================
+
+The Widget Object simplifies the process of streaming
+your data vizualization with WebRTC or MJPEG. Enconding.
```
Garyfallidis, line 7: In which cases MJPEG is preferable?

fury/stream/widget.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+class Widget:
+    def __init__(
+            self, showm, ms_stream=33,
+            ms_interaction=33,
+            host='localhost', port=None,
+            encoding='mjpeg', ms_jpeg=33,
+            queue_size=20):
+        """Thi Obj it's able execute the fury streaming system
```
Garyfallidis, line 53: Please add more information about this.

Garyfallidis: About what this class actually does.

Garyfallidis: It doesn't only execute the fury streaming system. Correct?

fury/stream/widget.py
```diff
@@ -0,0 +1,254 @@
...
+            int(self.window_size[0]*(1+0.1)),
+            int(self.window_size[1]*(1+0.1))
+        )
+        self.max_window_size = max_window_size
+        self.ms_stream = ms_stream
+        self.ms_interaction = ms_interaction
+        self.ms_jpeg = ms_jpeg
+        self._host = host
+        if port is None:
+            port = np.random.randint(7000, 8888)
```
Garyfallidis, line 90: Is this safe to do?

docs/tutorials/04_stream/viz_interaction.py
filipinascimento: Maybe we need a section in the documentation describing better the architecture and usage of the streaming system. A glossary of terms (WebRTC, h264, mjpeg, etc etc) and how the different optional packages improve performance.

Garyfallidis (commented):
This is the first round of my reviews. More asap.

docs/tutorials/04_stream/viz_interaction.py
devmessias: Yes. But just for WebRTC which is now optional

fury/stream/client.py
```diff
@@ -0,0 +1,412 @@
...
+    stream_client._in_request = True
+    stream_client.window2image_filter.Update()
+    stream_client.window2image_filter.Modified()
+    vtk_image = stream_client.window2image_filter.GetOutput()
+    vtk_array = vtk_image.GetPointData().GetScalars()
+    # num_components = vtk_array.GetNumberOfComponents()
+
+    w, h, _ = vtk_image.GetDimensions()
+    np_arr = np.frombuffer(vtk_array, dtype='uint8')
+    if np_arr is None:
```
skoudoro, line 44: What happened if `np_arr = np.array([])`?

fury/stream/client.py
```diff
@@ -0,0 +1,412 @@
...
+        if self._interval_timer is not None:
+            self._interval_timer.stop()
+            self._interval_timer = None
+
+        self._started = False
+
+    def cleanup(self):
+        """Relsese the shared memory resources if necessary.
+        """
+        self.circular_queue.cleanup()
```
skoudoro, line 412: It looks like a lot `FuryStreamClient start stop and cleanup`. 

Maybe there is a way to simplify this. Need some thinking

skoudoro (commented):
Hi @devmessias,

See below for some comments, I still need to spend time on `fury.stream.tools`

fury/stream/widget.py
devmessias: Yes, I think so. @filipinascimento  do you have any considerations?

fury/stream/widget.py
devmessias: No, this also creates a process to start the webrtc enconding.

fury/stream/client.py
devmessias: It seems that vtk_array can be just two things: None (empty buffer) of a non-empty array (buffer). But I can change this condition to
```python
if np_arr:
...
```

fury/stream/client.py
```diff
@@ -0,0 +1,412 @@
+import vtk
+import sys
+import logging
+import time
+if sys.version_info.minor >= 8:
+    PY_VERSION_8 = True
+else:
+    shared_memory = None
+    PY_VERSION_8 = False
```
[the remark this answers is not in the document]
devmessias: solved in c074f01

docs/tutorials/04_stream/viz_interaction.py
Garyfallidis: Otherwise, create two classes one for shared memory and one for raw array.

fury/stream/widget.py
filipinascimento: This do not check for existing open ports, it is safe, but can fail if the selected port is already open.

devmessias: We can check this. But this operation takes some time.

fury/stream/client.py
```diff
@@ -0,0 +1,412 @@
...
+    the ImageManager instance
+
+    Parameters
+    ----------
+    args : args
+    kwargs : kwargs
+
+    """
+    stream_client = kwargs['stream_client']
+    if stream_client._in_request:
```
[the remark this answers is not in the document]
devmessias: Solved

fury/stream/tools.py
```diff
@@ -0,0 +1,1000 @@
...
+    Notes
+    -----
+        More details at: https://bugs.python.org/issue38119
+
+    """
+
+    def fix_register(name, rtype):
+        if rtype == "shared_memory":
+            return
+        return resource_tracker._resource_tracker.register(
```
filipinascimento, line 51: `self` is undefined here. Results in a silent error when the server is running in another process.

devmessias: This is how we can do monkey patching on python. This is an open bug in CPython implementation regarding shared memory

devmessias: Alto this is silent ERROR or silent Warning?

filipinascimento: Indeed this results in a warning caused by an unraisable exception, which I'm not sure if it is unraisable because it is running during garbage collector, destroy or because it runs in another process.

devmessias: What you can do is call `ls -ls /dev/shm` and see if some resource was not properly removed from.

filipinascimento: Ok, I've tested this, and the patch works on mac. The problem is that, for some reason, pytest checks this self and gives an NameError. On macs, /dev/shm does not exist, but I've checked for memory usage and everything seems to work nicely on Python 3.8.

filipinascimento: Adding a NameError try catch, as you suggested, gets rid of the warning for when running pytest. Memory is reclaimed. so everything should be fine. The problem does not happen on Windows.

docs/tutorials/04_stream/viz_interaction.py
filipinascimento: it seems that for mac, ffmpeg, opus or libvpx are not needed. I've just pipinstall aiortc

fury/stream/client.py
```diff
@@ -0,0 +1,386 @@
...
+            self._id_observer = None
+
+        if self._interval_timer is not None:
+            self._interval_timer.stop()
+            self._interval_timer = None
+
+        self._started = False
+
+    def cleanup(self):
+        """Relsese the shared memory resources if necessary.
```
filipinascimento, line 384: Relsese->Release

filipinascimento (approved):
I think the PR is ready (may need a pass from Serge and Elef to make sure everything is ok with coding style). We can improve the documentation incrementally later on.

fury/stream/client.py
```diff
@@ -0,0 +1,386 @@
...
+            whithout_iren_start=False,
+            num_buffers=2,
+    ):
+        """
+        A StreamClient extracts a framebuffer from the OpenGL context
+        and writes into a shared memory resource.
+
+        Parameters
+        ----------
+        showm : FuryShowmManager
```
Garyfallidis, line 58: Just say ShowManager

fury/stream/widget.py
Garyfallidis: This object is able to execute the FURY streaming system ...

fury/stream/widget.py
```diff
@@ -0,0 +1,252 @@
...
+
+class Widget:
+    def __init__(
+            self, showm, ms_stream=33,
+            ms_interaction=33,
+            host='localhost', port=None,
+            encoding='mjpeg', ms_jpeg=33,
+            queue_size=20):
+        """Thi Obj it's able execute the fury streaming system
+        using the SharedMemory from python multiprocessing.
```
Garyfallidis, line 52: using the SharedMemory object from Python multiprocessing.

requirements/optional.txt
```diff
@@ -1,2 +1,4 @@
 matplotlib>=2.0.0
 dipy>=1.0.0
+opencv-python
```
Garyfallidis, line 3: Do we still need opencv?

Garyfallidis (changes requested):
Hi @devmessias Please see comments and also address pep 8 issues. Keep pushing this is getting super close to be merged! FURY streaming is on the way! Go go team!

requirements/optional.txt
devmessias: Ok,  It seems that there are no significant improvements. I will remove and leave only the option with the Pillow lib.

```python
import cv2
import io
from PIL import Image 
import numpy as np
import pytest


def convert_pil(image):
    image_encoded = Image.fromarray(image, mode="RGB")
    bytes_img_data = io.BytesIO()
    image_encoded.save(bytes_img_data, format='jpeg')
    bytes_img = bytes_img_data.getvalue()

def convert_cv2(image):
    # preserve width and height, flip B->R
    image = image[:, :, ::-1]
    image_encoded = cv2.imencode('.jpg', image)[1]
    bytes_img = image_encoded.tobytes()

@pytest.mark.parametrize("what", ["cv2", "pil"])
@pytest.mark.benchmark(
    min_rounds=5,
    disable_gc=True,
    warmup=True
)
def test_cv_vs_pil(benchmark, what):
    image = np.random.normal(size=(1024, 2048, 3)).astype(np.uint8)
    if what=='cv2':
        benchmark(convert_cv2, image)
    else:
        benchmark(convert_pil, image)
```

devmessias: Done at 609a439a5e1ae42108ccbf4d49eb9a635f539f29

fury/stream/tools.py
devmessias: Done at https://github.com/fury-gl/fury/pull/437/commits/25b39792d69c212eb02c8bd5f6957c50cee3c95e
Maybe is good to describe this issue on cypthon repo. https://github.com/python/cpython/issues/82300

Garyfallidis:
Congratulations @devmessias This important PR is finally merged. Great job!

Garyfallidis:
Next step @devmessias is the creation of furyspeed package that will accelerate streaming speed @skoudoro

PR #440, opened 2021-06-10
added support for URL image in ImageContainer2D

antrikshmisri:
This adds support for fetching the image from a URL. The image is stored in the directory temporarily until the image has been processed, then it is removed from the directory. I'll try to look at other approaches that may be more flexible than this in the meantime and update this if necessary.

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -25,10 +25,15 @@ def load_image(filename, as_vtktype=False, use_pillow=True):
         desired image array
 
     """
-    is_url = filename.startswith('http://') or filename.startswith('https://')
+    is_url = filename.lower().startswith('http://') \
+        or filename.lower().startswith('https://')
 
     if is_url:
         image_name = os.path.basename(filename)
+
+        if len(image_name.split('.')) < 2:
+            raise IOError(f'{filename} is not a valid image URL')
+
         urlretrieve(filename, image_name)
         filename = image_name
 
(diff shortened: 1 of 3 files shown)


skoudoro (commented):
Hi @antrikshmisri,

Thank you for this. 

I recommend you:
- to remove the parameter `is_url` everywhere.
- To check if the filename starts with `http://` or `https://`
- Be aware of your `os.remove` because it seems to do it every time.

So in your `load_image` function you can do: `is_url = filename.startswith("http://") or filename.startswith("https://")`

Thank you

fury/io.py
```diff
@@ -24,6 +25,13 @@ def load_image(filename, as_vtktype=False, use_pillow=True):
         desired image array
 
     """
+    is_url = filename.startswith('http://') or filename.startswith('https://')
```
skoudoro, line 28: `filename.lower().startswith('http://')` in case you have `HTTP://`

skoudoro (commented):
Thank you for the update.

Please, Can you add a test @antrikshmisri?

After that it is ready to be merged

fury/io.py
antrikshmisri: Added the tests/requested changes.

skoudoro (approved):
Looks good to me.

Can you have a quick look @Nibba2018? 

I will merge this PR by the end of the day

skoudoro:
Thank you @antrikshmisri! merging

PR #398, opened 2021-03-25
feat: added a Card2D widget to UI

antrikshmisri:
This adds a card2d widget as mentioned in issue #397.
Here is a screenshot for reference:-
Screenshot (176)

PR #442, opened 2021-06-16
added method to wrap overflowing text

antrikshmisri:
This PR adds a method `wrap_overflow` to wrap the overflowing text in `TextBlock2D`.


fury/ui.py
```diff
@@ -69,6 +69,53 @@ def clip_overflow(textblock, width, side='right'):
...
+        Wrapped version of the text.
+    """
+    original_str = textblock.message
+    start_ptr = 0
+    end_ptr = len(original_str)
+
+    if wrap_width <= 0:
+        raise ValueError(
+            f'Value of wrap_width cannot be less that or equal to 0'
+        )
```
Nibba2018, line 94: Instead of throwing an error it would be better to handle it silently. (Make the algo robust enough to handle those cases)

fury/ui.py
```diff
@@ -69,6 +69,53 @@ def clip_overflow(textblock, width, side='right'):
...
+    while start_ptr < end_ptr:
+        mid_ptr = (start_ptr + end_ptr)//2
+        textblock.message = original_str[:mid_ptr]
+        if textblock.size[0] < wrap_width:
+            start_ptr = mid_ptr
+        elif textblock.size[0] > wrap_width:
+            end_ptr = mid_ptr
+
+        if mid_ptr == (start_ptr + end_ptr)//2 or\
+           textblock.size[0] == wrap_width:
```
Nibba2018, line 108: You need to rework this, There's a a lot of code duplication here with respect to `clip_overflow`. 
That's why asked you to implement a separate method to check for overflow and then call it in `clip_overflow` and `wrap_overflow` as necessary.

fury/ui.py
```diff
@@ -69,6 +69,53 @@ def clip_overflow(textblock, width, side='right'):
...
+
+            for idx, _ in enumerate(original_str):
+                if idx % mid_ptr == 0 and idx != 0:
+                    original_str = original_str[:idx]\
+                     + '\n' + original_str[idx:]
+
+            textblock.message = original_str
+            return textblock.message
```
Nibba2018, line 116: `enumerate` in general is expensive and you are discarding the characters anyway so its better to use `range` instead.

Also you haven't considered the case of textblocks with backgrounds as the size calculation for them is different.

Nibba2018 (changes requested):
Hello @antrikshmisri , there are some things that needs to be addressed that I have mentioned below.

Thank you.

fury/ui.py
antrikshmisri: Hey @Nibba2018, sure this makes sense. Will update the PR accordingly.

fury/ui.py
antrikshmisri: Sure, makes sense. Also, will update w.r.t textblocks with background.

skoudoro:
What is the status of this PR @antrikshmisri?

@Nibba2018, Can you review it after the last update?

This PR is almost ready to go I think. Thank you for the feedback

Nibba2018:
Hello @skoudoro , It's not done yet as it still has code duplicacy and some inefficient operations as I mentioned in my review.

skoudoro:
ok, Thank you for the feedback! Let's finish this PR before the next online meeting

antrikshmisri:
> What is the status of this PR @antrikshmisri?

Hey @skoudoro, this PR needs a few updates. Those will be added hopefully by tomorrow. Thanks!

antrikshmisri:
Hey @Nibba2018, I tried to add the is_overflowing method but only a little duplication. I think it would make sense to create an umbrella class and implement all the methods in that class. That way a lot of code and variables can be reused. Do you think this is viable?

skoudoro:
> Do you think this is viable?

Personnally, I recommend the solution without the class and you @Nibba2018?

Nibba2018:
Yes, I think the same @skoudoro . For helper methods I don't think classes are necessary.

antrikshmisri:
Sure, then I'll continue with function instead of class.

antrikshmisri:
Hey @Nibba2018, I have addressed all the comments. PTAL

fury/ui/helpers.py
```diff
@@ -44,14 +44,91 @@ def clip_overflow(textblock, width, side='right'):
...
+    if wrap_width <= 0:
+        wrap_width = textblock.size[0] - abs(wrap_width) % textblock.size[0]
+
+    if textblock.size[0] <= wrap_width:
+        textblock.have_bg = prev_bg
+        return original_str
+
+    while start_ptr < end_ptr:
+        mid_ptr = (start_ptr + end_ptr)//2
+        textblock.message = original_str[:mid_ptr]
```
Nibba2018, line 88: This entire block can be reused.

Nibba2018 (commented):
Hello @antrikshmisri , there can be further optimisation as I have highlighted below.

fury/ui/helpers.py
antrikshmisri: Hey @Nibba2018, this has been fixed. Thanks!

fury/ui/helpers.py
```diff
@@ -23,6 +22,74 @@ def clip_overflow(textblock, width, side='right'):
...
+
+    is_overflowing, mid_ptr = check_overflow(textblock, width, '...', side)
+
+    if mid_ptr == 0:
+        return original_str
+
+    if not is_overflowing:
+        textblock.have_bg = prev_bg
+        return textblock.message
+
```
skoudoro, line 37: What happens in all the other cases? for example `is_overflowing= True` and `mid_ptr !=0`?  

Please, can you add a test for this case?

fury/ui/helpers.py
```diff
@@ -23,6 +22,74 @@ def clip_overflow(textblock, width, side='right'):
...
+    if not is_overflowing:
+        for i in range(len(original_str)):
+            if i % mid_ptr == 0 and i != 0:
+                original_str = original_str[:i]\
+                    + '\n' + original_str[i:]
+
+        textblock.have_bg = prev_bg
+        textblock.message = original_str
+        return textblock.message
+
```
skoudoro, line 74: Same questions as above? what happens if we do not fill both conditions?

skoudoro (commented):
Hi @antrikshmisri,

Can manage the corner case. After that, it should be good to go

skoudoro:
Do you have an update concerning my comment above?

fury/ui/helpers.py
antrikshmisri: Hey @skoudoro, with the lastest commit this check has been removed as the value will always be `False`.

fury/ui/helpers.py
antrikshmisri: Same as above, the checks were not necessary.

antrikshmisri:
> Do you have an update concerning my comment above?

Hey @skoudoro, the comments have been addressed with the latest commit. Also the bugs discussed in the meeting have also been fixed. Thanks!

skoudoro:
Great! Thank you for the feedback.  I will play with it and tell you if all is good.

@Nibba2018, Can you have a last look?

Nibba2018:
Is it me or the methods dont work anymore?
image

antrikshmisri:
> Is it me or the methods dont work anymore?
> image
> 

Hey @Nibba2018, you need to call wrap overflow after you add the text to the the scene.

Nibba2018:
Yeah, sorry my bad :P

Nibba2018:
Here are some of the points that we discussed in the meeting:

- [ ] Add support for hyphens when words are wrapped in between.
    In my opinion it would be better to not wrap words in between and instead move them to the next line.
- [ ] I currently dont remember the use case that @Garyfallidis proposed which caused the the wrap overflow to not function properly.

fury/ui/helpers.py
```diff
@@ -23,6 +22,81 @@ def clip_overflow(textblock, width, side='right'):
...
+    wrapped text : str
+        Wrapped version of the text.
+    """
+    original_str = textblock.message
+    str_copy = textblock.message
+    prev_bg = textblock.have_bg
+    wrap_idx = []
+
+    _, mid_ptr = check_overflow(textblock, wrap_width, '', side)
+
```
Nibba2018, line 62: I would recommend to change the name of `mid_ptr`. Maybe `wrap_idx`? and `wrap_idx` to `wrap_idxs` or something similar which is plural.

fury/ui/helpers.py
```diff
@@ -23,6 +22,81 @@ def clip_overflow(textblock, width, side='right'):
...
+        return original_str
+
+    wrap_idx.append(mid_ptr)
+
+    while mid_ptr != 0:
+        str_copy = str_copy[mid_ptr:]
+        textblock.message = str_copy
+        _, mid_ptr = check_overflow(textblock, wrap_width, '', side)
+        if mid_ptr != 0:
+            wrap_idx.append(wrap_idx[-1]+mid_ptr+1)
```
Nibba2018, line 73: Same as above.

fury/ui/helpers.py
```diff
@@ -23,6 +22,81 @@ def clip_overflow(textblock, width, side='right'):
...
+        The textblock object whose text is to be checked.
+    width: int
+        Required width of the text.
+    overflow_postfix: str, optional
+        Postfix to be added to the text if it is overflowing.
+
+    Returns
+    -------
+    tuple(bool, int)
+
```
Nibba2018, line 99: You need to refactor the signature of this method as you never return `True` for this method.

fury/ui/tests/test_helpers.py
```diff
@@ -47,3 +47,46 @@ def test_clip_overflow():
...
+    rectangle = ui.Rectangle2D(position=(50, 50), size=(100, 50))
+
+    sm = window.ShowManager()
+    sm.scene.add(rectangle, text)
+
+    text.message = "A very very long message to clip text overflow"
+
+    is_overflowing, _ = check_overflow(text, rectangle.size[0], '~')
+
+    npt.assert_equal(False, is_overflowing)
```
Nibba2018, line 91: This should return `True` not `False` as it indeed is overflowing.

Nibba2018 (commented):
Some minor changes apart from that it should be good to go.

Also there's a case where if the string is too long it would overflow vertically from the specified space which may not be a desired behaviour for #398 . It could be fixed later in a different PR.

skoudoro:
> Add support for hyphens when words are wrapped in between.

Concerning this point, I think we said that we keep it for another PR. Let's not spend too much time on this PR.

> In my opinion it would be better to not wrap words in between and instead move them to the next line.

When the rectangle is too small, you can not even do that. Some words are too long. Let's keep the behavior simple for now. 
It is good as it is for a second version.

fury/ui/helpers.py
antrikshmisri: This has been updated to avoid confusion.

fury/ui/helpers.py
antrikshmisri: Done.

fury/ui/helpers.py
antrikshmisri: The method has been updated, there wasn't any purpose for the bool, now it returns just the overflow idx.

fury/ui/tests/test_helpers.py
antrikshmisri: Test have been updated accordingly.

antrikshmisri:
Hey @Nibba2018, all the comments have been updated, PTAL. Thanks!

fury/ui/helpers.py
```diff
@@ -23,6 +22,81 @@ def clip_overflow(textblock, width, side='right'):
...
+        The textblock object whose text is to be checked.
+    width: int
+        Required width of the text.
+    overflow_postfix: str, optional
+        Postfix to be added to the text if it is overflowing.
+
+    Returns
+    -------
+    int
+
```
Nibba2018, line 99: You need to explain what this integer represents. @antrikshmisri

antrikshmisri: Done.

Nibba2018 (approved):

skoudoro (approved):
+1 for merging

PR #445, opened 2021-06-18
[WIP] Example to show how to render multiple bonds

SunTzunami:
A small example to show how to render multiple bonds.
Molecules rendered - 
1. Ethane (single bond)
2. Ethene (double bond)
3. Ethyne (triple bond)

Preview - 

viz_multiple_bonds

 @nasimanousheh PTAL


SunTzunami:
updated preview: 
viz_multiple_bonds

ghutchis:
Using the length is really tricky, because C=O, C#N (triple bond to nitrogen) etc. all have different expected bond lengths.

You can use the sum of the covalent atomic radii vs. the bond length as an estimate.. For example in Avogadro, we use something like this as a guess:
https://github.com/OpenChemistry/avogadrolibs/blob/master/avogadro/qtplugins/editor/editor.cpp#L374

SunTzunami:
Thank you @ghutchis ! That is indeed an elegant way of estimating bond order.

ghutchis:
Understand that it doesn't work that well. If you read a file format with bond orders, use those. The real code to perceive bond orders is much more complicated--you need to look at bond angles, handle cases where the bond orders exceed limits, etc.

But it's a heuristic that's an okay starting point.

SunTzunami:
Insightful info. Thanks @ghutchis !

docs/examples/viz_multiple_bonds.py
```diff
@@ -0,0 +1,280 @@
...
+    return c1_l, c1_u, c2_l, c2_u
+
+###############################################################################
+# Here, we define the function which will create the ball and stick model
+# It takes in two numpy arrays as input-
+# atom_coords: array of atomic coordinates
+# elem_names: array of names of elements corresponding to atom_coords
+
+
+def ball_and_stick(atom_coords, elem_names):
```
skoudoro, line 67: Do you plan to move this function to the molecular module?

docs/examples/viz_multiple_bonds.py
```diff
@@ -0,0 +1,280 @@
...
+from fury import disable_warnings, window, actor
+import vtk
+disable_warnings()
+
+###############################################################################
+# Offsets for double and triple bonds (generate coordinates of the streamtubes
+# which will represent bonds)
+
+
+def offsets(atom_coords, c1_coord, c2_coord, bond_type):
```
skoudoro, line 24: Do you plan to move this function to the molecular module?

skoudoro (commented):
Hi @SunTzunami,

I was wondering if those functions will go to another module. Also, you were not using the `cpkr` anymore (`vtkperiodictable` instead) so is this code up to date?
What are the next steps?

SunTzunami:
Hi @skoudoro!
 I think I'll close this PR. This was before the vtkOpenGLMoleculeMapper. Now that we're making a proper molecular module, this will be part of the demo that I make using molecular module.

skoudoro:
make sense 👍🏾

PR #462, opened 2021-07-17
Updated helical, brownian, emwave animations

SunTzunami:
1. Updated the demo of helical motion to stop using multiple line actors as discussed in the meeting.
2. Updated the demo of brownian motion to make it more scientifically useful (removed unnecessary rotation of camera during animation and box actor)
3. Added simulation data to both animations (number of simulated steps for brownian motion and velocity of the particle for helical motion)

Preview - 

1. Helical motion 

viz_helical_motion

2. Brownian motion 

viz_brownian_motion


docs/examples/viz_helical_motion.py
```diff
@@ -35,21 +34,25 @@
...
 phase_angle = 0.002
+radius_path = 1.5
 
 
 ###############################################################################
 # Creating a scene object and configuring the camera's position
 
 scene = window.Scene()
+scene.background((1, 0, 0))
+scene.GradientBackgroundOn()
```
Garyfallidis, line 55: @SunTzunami Add extra options in ``scene.background`` to enable the gradient. Do not call a vtk method here like that. Thx.

docs/examples/viz_helical_motion.py
```diff
@@ -68,94 +71,130 @@
 directions = np.array([[1, 0, 0]])
 heights = np.array([8])
 arrow_actor = actor.arrow(centers, directions, color_arrow, heights,
-                          resolution=20, tip_length=0.06, tip_radius=0.012,
+                          resolution=50, tip_length=0.06, tip_radius=0.012,
                           shaft_radius=0.005)
+arrow_actor.GetProperty().SetAmbient(1)
```
Garyfallidis, line 76: Same here @SunTzunami. Please make a function in ``material.py`` that takes an input an actor and changes its ambience or use directly the function ``manifest_standard`` from the same file to chose the options that you need.

docs/examples/viz_helical_motion.py
```diff
@@ -68,94 +71,130 @@
...
+
+path_actor = actor.line([position], colors, linewidth=3)
+path_actor.position = position
+path_actor.vertices = utils.vertices_from_actor(path_actor)
+path_actor.vcolors = utils.colors_from_actor(path_actor)
+print(len(path_actor.vcolors))
+path_actor.no_vertices_per_point = \
+    len(path_actor.vertices) / num_total_steps
+path_actor.initial_vertices = path_actor.vertices.copy() - \
+    np.repeat(position, path_actor.no_vertices_per_point, axis=0)
```
Garyfallidis, line 108: @SunTzunami please make a utility function called ``attributes_to_actor`` in ``fury.actor`` and move these operations (lines 101-105) there.

Garyfallidis: The function should take as input an actor and return the actor with the new attributes as you did above.

docs/examples/viz_helical_motion.py
```diff
@@ -68,94 +71,130 @@
...
+path_actor.GetProperty().SetAmbient(1)
```
Garyfallidis, line 109: See my other comment on the same issue.

docs/examples/viz_helical_motion.py
SunTzunami: Done.

docs/examples/viz_helical_motion.py
SunTzunami: Used manifest_standard. Done.

docs/examples/viz_helical_motion.py
SunTzunami: Created attributes_to_actor.

docs/examples/viz_helical_motion.py
SunTzunami: Done

SunTzunami:
Hi @Garyfallidis ! Thank you for the feedback. I've addressed all your comments. PTAL.

docs/examples/viz_helical_motion.py
```diff
@@ -68,94 +70,123 @@
...
+###############################################################################
+# Initializing point actor which will represent the charged particle
+
+color_particle = window.colors.yellow  # color of particle can be manipulated
+charge_actor = actor.point(pts, color_particle, point_radius=radius_particle,
+                           phi=32, theta=32)
+actor.attributes_to_actor(charge_actor, pts)
+
+# manipulating shading for aesthetics
+material.manifest_standard(charge_actor, ambient_level=1, diffuse_level=1,
```
Garyfallidis, line 116: Is the effect exactly how it was before? If not then create an additional ``manifest_ambience`` function and only enable ambient light.

SunTzunami: The effect was a bit different, I created `manifest_ambience` accordingly.

SunTzunami:
Thanks for the feedback @Garyfallidis ! I've addressed all the comments.

docs/examples/viz_emwave_animation.py
```diff
@@ -171,4 +160,5 @@ def timer_callback(_obj, _event):
 interactive = False
 if interactive:
     showm.start()
+print(scene.get_camera())
```
skoudoro, line 163: Can you remove this `print`? Thanks !

skoudoro (commented):
Hi @SunTzunami,

Overall, it looks good, this is a nice update. I have some doubts concerning the function name (`attributes_to_actor`). I believe we should rename it. This is a point to discuss.

docs/examples/viz_emwave_animation.py
SunTzunami: Removed. Thanks!

SunTzunami:
Hi @skoudoro ! Thanks for the feedback! I agree that `attributes_to_actor` can be misleading as `fury.shaders` has a function called `attribute_to_actor`. Any suggestions for a new function name?

SunTzunami:
Hi @Garyfallidis @skoudoro 
I've rebased and updated the PR, please let me know an appropriate name instead of  ```attributes_to_actor ```

saguileran:
Hi.

I modified the files viz_brownian_motion and viz_helical_motion to add time step and velocities text to the animation, the 2nd and 3rd requests made by SunTzunam but I do not understand the first request, what is it about?

I uploaded the python scripts in a zip file, is this the best way to do it or there is a more appropriate way? 

[viz.zip](https://github.com/fury-gl/fury/files/8165268/viz.zip)i

PR #466, opened 2021-07-20
two more small bib changes

danielskatz:


Garyfallidis:
Got it. Done and merged. Apologies we missed that. Should be good now.

danielskatz:
@whedon recommend-accept

Issue #457, opened 2021-07-08
Improve loading speed using partial imports


Garyfallidis:
We can try to make our scripts load faster by importing only the functions we need from VTK. This is supposed to be supported from VTK 9 and on. Let's try it.

The issue was reported during our JOSS review https://github.com/openjournals/joss-reviews/issues/3384 by N.P. Rougier.

skoudoro:
This issue has been addressed by all the PR's above. We can close it for now. We still need to compare the performance. This can be done on a new issue.

PR #470, opened 2021-07-26
Update the way we import external libraries by using only the necessary modules

skoudoro:
This PR is the beginning of #457. 

The goal is to make our scripts load faster by importing only the functions/modules we need from VTK.

### DIFF
--- a/fury/__init__.py
+++ b/fury/__init__.py
@@ -25,7 +25,7 @@ def get_info(verbose=False):
     import sys
     import numpy
     import scipy
-    import vtk
+    import vtkmodules.vtkCommonCore as ccvtk
 
     mpl, have_mpl, _ = optional_package('matplotlib')
     dipy, have_dipy, _ = optional_package('dipy')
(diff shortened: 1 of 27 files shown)


skoudoro:
This PR is ready to go. Can you have a look @Garyfallidis and @guaje? 

Thank you

Garyfallidis:
@skoudoro there is currently a conflict. 
I wonder if we should have one file that does all the vtk imports and then just load functions from that file.
For example,
```
from fury.vtk import vtkPolyDataMapper
```

Garyfallidis:
And then in fvtk.py we add the 

```
import vtkmodules.vtkCommonCore as ccvtk
Polydata = ccvtk.vtkPolyDataMapper
or
vtkPolydata = ccvtk.vtkPolyDataMapper

```
In this way in the other scripts you can just do
```
from fury.fvtk import vtkPolyData
or 
from fury.fvtk import Polydata
```
Also what speedup do you see?

Issue #463, opened 2021-07-17
`GridUI` throws error when captions are `None`


antrikshmisri:
## Description
When the `captions` parameter is set to `None` the `GridUI` throws the following error:-
```python
'vtkmodules.vtkRenderingOpenGL2.vtkOpenGLActor' object has no attribute '_items'
  File "fury/ui/containers.py", line 774, in init
    self._actors.append(item._items[0])
  File "grid_testing.py", line 15, in <module>
    grid = ui.GridUI(actors)
```
This is being caused because with captions the actors are converted to a `Container` that has the actor and the text but when captions are set to None the actor is passed as is, thus throwing `AttributeError`.

## Way to reproduce
1. Run the following code (captions set to None), to see the error.
```python
from fury import window, actor, ui
import numpy as np


scene = window.Scene()
scene.zoom(12.5)
scene.set_camera(position=(60, 40, 30), focal_point=(0.0, 0.0, 0.0),
                 view_up=(0.0, 0.0, 1.0))
showm = window.ShowManager(scene, size=(600, 600))

sp1 = actor.sphere(np.zeros((1, 3)), colors=np.random.rand(3), radii=1)
sp2 = actor.sphere(np.zeros((1, 3)), colors=np.random.rand(3), radii=1)

actors = [sp1, sp2]
sgrid = ui.GridUI(actors)  # gives an error\
```
A patch needs to be made somewhere over [here](https://github.com/antrikshmisri/fury/blob/d498b6d1cf7b9594f55a80cfcbc047a9c8fd5bd1/fury/ui/containers.py#L773-L775).

skoudoro:
Good catch! thank you for pointing that

aadityasinha-dotcom:
Is this issue available?

ganimtron-10:
Hello @aadityasinha-dotcom ,
I have worked on this issue and created a PR regarding the same!!
Feel free to look around for another issues.
Thanks!

aadityasinha-dotcom:
Okay

skoudoro:
fixed by https://github.com/fury-gl/fury/pull/478. closing

PR #478, opened 2021-07-30
Resolving GridUI caption error

ganimtron-10:
Fixing Bug mentioned at #463.


skoudoro (commented):
Hi @ganimtron-10,

Thank you for doing that! Could you add a test also? Thank you

@antrikshmisri, Can you test this PR and check if it fix the reported issue. Thanks!

ganimtron-10:
I have added tests for the changes.

skoudoro:
Thank for the update @ganimtron-10!

ping @antrikshmisri

fury/tests/test_actors.py
```diff
@@ -1176,6 +1176,7 @@ def timer_callback(_obj, _event):
     arr[arr < 20] = 0
     report = window.analyze_snapshot(arr)
     npt.assert_equal(report.objects, 6)
+    npt.assert_equal(type(container._items[0]), actor.Container)
```
antrikshmisri, line 1179: Please avoid using private members, instead use the property.

fury/tests/test_actors.py
```diff
@@ -1196,6 +1197,7 @@ def timer_callback(_obj, _event):
     arr = window.snapshot(scene)
     report = window.analyze_snapshot(arr)
     npt.assert_equal(report.objects > 6, True)
+    npt.assert_equal(type(container._items[0]), actor.Container)
```
antrikshmisri, line 1200: Same comment as above.

antrikshmisri (commented):
Hey @ganimtron-10, please see my comments below. Rest everything seems to be working fine. Thanks!

ganimtron-10:
Thanks @antrikshmisri for the review!
I have made changes according to the comments.

fury/tests/test_actors.py
ganimtron-10: Done!

antrikshmisri:
Hey @ganimtron-10, thanks for the update. LGTM.

antrikshmisri (approved):

ganimtron-10:
Hii @antrikshmisri !! 
Is there any thing left to be changed in this pr??

skoudoro (commented):
Thanks @antrikshmisri  for the review. 

I need to test this PR and we can go ahead and merge it. It should be done by the end of this week.

ganimtron-10:
Hello @skoudoro ,
Any updates on this?

skoudoro (changes requested):
Hi @ganimtron-10,

I do not think this is the right approach. 

It should be fixed at `GridUI` class level and not at the `grid`function level. 

Currently, you are adding complexity which is not necessary. to be more precise, you are doing:  `Container(Container(actor), Container(actor), etc...)`  and each container has only one actor so useless.

Please, let me know if you have time to update this PR, otherwise, I will close it. 

Thanks!

ganimtron-10:
Hello @skoudoro ,
Thanks for the update!

Last time, I just got confused between the return type of the `grid` function and the type of the `items` as I thought the `grid` function is inconsistent (ie. which returns either `Container` or `Actor`) due to which I implemented that approach, but this isn't the case.

Now I have updated the PR and made changes into the `GridUI` itself.
PTAL.

skoudoro (commented):
Much better! Thank you for the quick update! 

Can you just update the test of `gridUI`?  Then, it will be ready to merge. You could start with the initial issue #463 and make sure that we have the expected output in the test. 

Thanks !

skoudoro:
@ganimtron-10, this PR is almost ready to be merged. 

See my comment above. Can you update the tests? Thank you!

ganimtron-10:
Hello @skoudoro ,
I tried creating a similar test and as per the older conversation used try and except block to test whether it throw error or not. Please let me know if its correct or not.

ganimtron-10:
Hey @skoudoro ,
I have updated the tests and they are now passing on `macOS`.

skoudoro:
Thank you for this @ganimtron-10, looks good!

Issue #483, opened 2021-08-04
Fix: ShowManager Repeating Timer behavior and window size attribute


devmessias:
Hi everyone. I noticed some bugs on the ShowManager object

## Timer Event
ShowManager `destroy_timer` method destroys the timer  without removing the callback related with the observer. Thus, 
we lost the  reference for the callback. Because of that,  another timer event  created after that will keep calling the 
callback function with the missed reference.   See the code bellow,

```python
from fury.window import ShowManager
from fury.actor import sphere
import numpy as np

repeat = False # bug appears with both options: False or True
centers = np.array([[1, 0,0]])
showm = ShowManager(size=(400, 400))
i1 = 0
i2 = 0

def t1(_, __):
    global i1
    i1 += 1
    print(f'timer 1 event {i1}')

def t2(_, __):
    global i2
    i2 += 1
    print(f'timer 2 event {i2}')

showm.initialize()
id1 = showm.add_timer_callback(repeat, 1, t1)
id2 = showm.add_timer_callback(repeat, 2, t2)

showm.destroy_timer(showm.timers[0])
color = np.array([0, 1, 0])
actor = sphere(centers ,color)
showm.scene.add(actor)

showm.start()
#showm.render()
assert i1 == 0
```

## Size Window

This is more trickly and usually happens on Windows.  The vtk  creates a window instance with a slightly different size.
Therefore,  the following test can fail:
```python
size = (300, 400)
showm = ShowManager(size=size)
# code goes here
assert size[0] = showm.size[0] and size[1] == showm.size[1]
```
I noticed  that bug running the  tests of PR #437  on windows github actions.

OS and pkgs

```
{'fury_version': '0.7.1', 'pkg_path': '/home/devmessias/phd/fury/fury',
'commit_hash': '426c15b9e4895bdba449be3c63168b5179a34202', 'sys_version': '3.8.5
(default, Sep  4 2020, 07:30:14) \n[GCC 7.3.0]', 'sys_executable':
'/home/devmessias/anaconda3/bin/python', 'sys_platform': 'linux',
'numpy_version': '1.20.3', 'scipy_version': '1.5.2', 'vtk_version': '9.0.3',
'matplotlib_version': '3.3.2', 'dipy_version': '1.4.1'}
```

skoudoro:
Hi @devmessias, 

Good catch for the timer event issue !!

I can not reproduce the error on window size. Are you sure about this one?

filipinascimento:
[...]

PR #484, opened 2021-08-04
Fixes some bugs from ShowManager Object.

devmessias:
Related with #483 
This PR adress two minor issues with the ShowManager object

## 1.  destroy timer behavior

ShowManager destroy_timer method destroys the timer without removing the callback related with the observer. Thus,
we lost the reference for the callback. Because of that, another timer event created after that will keep calling the
callback function with the missed reference.

## 2.  Size Window attribute
This is more trickly and usually happens on Windows. The vtk creates a window instance with a slightly different size.
Therefore, the following test can fail:
```python
size = (300, 400)
showm = ShowManager(size=size)
# code goes here
assert size[0] = showm.size[0] and size[1] == showm.size[1]
I noticed that bug running the tests of PR #437 on windows github actions.
```

This is an example which shows the bug solved by this PR. 

```python
from fury.window import ShowManager
from fury.actor import sphere
import numpy as np

centers = np.array([[1, 0,0]])
showm = ShowManager(size=(400, 400))
i1 = 0
i2 = 0

#check_bug = False  

def t1(_, __):
    global i1
    i1 += 1
    print(f'timer 1 event {i1}')

def t2(_, __):
    global i2
    i2 += 1
    print(f'timer 2 event {i2}')

id1 = showm.add_timer_callback(False, 1, t1)
print(id1, showm._timers)
id2 = showm.add_timer_callback(False, 2, t2)

print(id2, showm._timers)
showm.destroy_timer(id1)

showm.initialize()

color = np.array([0, 1, 0])
actor = sphere(centers, color)
showm.scene.add(actor)

showm.start()
#showm.render()
assert i1 == 0 # the code will fail here
```
After this PR, the above code should work whithout problems

### DIFF
--- a/fury/tests/test_window.py
+++ b/fury/tests/test_window.py
@@ -335,15 +335,15 @@ def test_record():
             assert_less_equal(arr.shape[1], 5000)
 
 
-def test_timers():
+def test_timers(interactive=False):
     centers = np.array([[1, 0, 0]])
     showm = window.ShowManager(size=(400, 400))
     # number of times that  each time call is called
-    counts_by_interval = {0: 0, 1: 0, 2: 0}
+    counts_by_interval = {0: 0, 1: 0, 2: 0, 3:0}
     # interval of each timer in miliseconds
-    ms_1 = 1
-    ms_2 = 100
-    ms_3 = 1000
+    ms_1 = 10
+    ms_2 = 10
+    ms_3 = 100
     state = {
         'first_call': True,
         # amount of milleseconds passe since the timer was started
@@ -364,7 +364,9 @@ def test_timers():
     def t3(iren, event_name_str):
         if state['current_time'] >= ms_3:
             counts_by_interval[2] += 1
-            showm.iren.TerminateApp()
+
+            showm.destroy_timers()
+            showm.exit()
 
     id1 = showm.add_timer_callback(False, ms_1, t1)
     showm.add_timer_callback(True, ms_2, t2)
@@ -378,7 +380,9 @@ def test_timers():
     showm.scene.add(sphere)
 
     state['first_time'] = time.time()
-    showm.start()
-    # callback 1 was not called
-    npt.assert_equal(counts_by_interval[0], 0)
-    npt.assert_equal(counts_by_interval[1] >= ms_3//ms_2, True)
+    if interactive:
+        showm.start()
+        # callback 1 was not called
+        npt.assert_equal(counts_by_interval[0], 0)
+        npt.assert_equal(counts_by_interval[1] >= ms_3//ms_2, True)
+
--- a/fury/tests/test_window.py
+++ b/fury/tests/test_window.py
@@ -337,8 +337,41 @@ def test_record():
             assert_less_equal(arr.shape[1], 5000)
 
 
+def test_timers_basic():
+    centers = np.array([[1, 0, 0]])
+    showm = window.ShowManager(size=(400, 400))
+    # interval of each timer in miliseconds
+    ms_1 = 10
+    ms_2 = 10
+    state = {
+        'first_call': True,
+        # amount of milleseconds passe since the timer was started
+        'current_time': 0,
+    }
 
-def test_timers():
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Hi @devmessias,

Looks good to me. Can you add some tests to make sure that the detected bug will not appear again?

After that, we should merge this PR. Thanks!

devmessias:
> Hi @devmessias,
> 
> Looks good to me. Can you add some tests to make sure that the detected bug will not appear again?
> 
> After that, we should merge this PR. Thanks!

Hi @skoudoro. I'm trying to figure out how to test this whithout using showm.start(). I couldn't find a method inside of  vtkRenderWindowInteractor  which returns the list of observers and timers. In a vtkActor instance we have Mapper().GetObservers()

fury/tests/test_window.py
```diff
@@ -336,6 +337,60 @@ def test_content(filename='fury.png', colors_found=(True, True)):
...
+    color = np.array([0, 1, 0])
+    sphere = actor.sphere(centers, color)
+    showm.scene.add(sphere)
+
+    state['first_time'] = time.time()
+    if interactive:
+        showm.start()
+        # callback 1 was not called
+        npt.assert_equal(counts_by_interval[0], 0)
+        npt.assert_equal(counts_by_interval[1] >= ms_3//ms_2, True)
```
skoudoro, line 391: interactive is `False` so I do not know what you are testing in this test. Can you clarify?

fury/window.py
```diff
@@ -512,39 +511,107 @@ def play_events_from_file(self, filename):
...
-        self.iren.AddObserver("TimerEvent", timer_callback)
+        """Add a timer callback.
+
+        Parameters
+        ----------
+        repeat : bool
+            If True, the timer will repeat.
+        duration : int
+            Timer duration in milliseconds.
+        timer_callback : function
```
skoudoro, line 544: `callable` instead of `function`. (can be a class for example)

fury/window.py
```diff
@@ -512,39 +511,107 @@ def play_events_from_file(self, filename):
 
     def add_window_callback(self, win_callback,
                             event=vtk.vtkCommand.ModifiedEvent):
-        """Add window callbacks."""
-        self.window.AddObserver(event, win_callback)
+        """Add window callbacks.
+
+        Parameters
+        ----------
+        win_callback : function
```
skoudoro, line 518: `callable` instead of `function`. (can be a class for example)

skoudoro (commented):
Hi @devmessias,

This PR is almost ready to go, thank you! 

Can you address the pep8 and the comments below?

Garyfallidis:
@devmessias this PR needs more work. First of all, you need to explain in the description which exact bugs you are fixing. Referencing another issue is not enough. Also you have to explain a bit on the test what you are testing. Currently it is not clear.
At least in Windows I do not see the expected behavior and when I enable interactive = True I get this error:

```
  File "test_window.py", line 490, in <module>
    test_timers()
  File "test_window.py", line 391, in test_timers
    npt.assert_equal(counts_by_interval[1] >= ms_3//ms_2, True)
  File "C:\Users\elef\AppData\Local\Continuum\anaconda3\lib\site-packages\numpy\testing\_private\utils.py", line 427, in assert_equal
    raise AssertionError(msg)
AssertionError:
Items are not equal:
 ACTUAL: False
 DESIRED: True
```
Also there are still some free flying comments and pep 8 issues.
Please address the issues reported here and fix the title too. The title is very generic and does not help reviewing.
Thanks in advance!

Garyfallidis:
Ping @devmessias!

fury/tests/test_window.py
devmessias: I need to fix that.  The reason for  this is because I seems it's impossible check the `counts_by_interval` without start the ShowManager.

skoudoro:
Hi @devmessias,

Thank you for your update! Could you also rebase to resolve those conflicts?

PR #485, opened 2021-08-09
Adding ProgressUI to the UI

ganimtron-10:
Creating a ProgressUI Widget which can be used as Loading bars.

https://user-images.githubusercontent.com/64432063/128671950-849c19b5-17ca-47a2-92a5-3f6e357538c5.mp4

https://user-images.githubusercontent.com/64432063/131996694-e7cf3e34-b010-499d-a7d7-832c2348859e.mp4


ganimtron-10:
Hi @skoudoro @antrikshmisri !!
Please review whether the logic for this UI seems correct or not.

docs/tutorials/02_ui/viz_progress.py
```diff
@@ -0,0 +1,94 @@
...
+===========
+Progress UI
+===========
+
+This example shows how to use the Progress UI. We will demonstrate how to
+create a whole ProgressBar and show the progress of adding Cubes to the scene.
+
+First, some imports.
+"""
+from fury import ui, window, actor
```
antrikshmisri, line 12: Could you update the import order.

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+    """
+
+    def __init__(self, position=(0, 0), size=(300, 100), padding=20,
+                 initial_value=50, min_value=0, max_value=100,
+                 bg_color=(1, 1, 1), progress_color=(0.5, 0.5, 0.5)):
+        """Init this UI element.
+
+        Parameters
+        ----------
+        size : (float, float)
```
antrikshmisri, line 3068: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Center of the slider's center.
+        initial_value : float
```
antrikshmisri, line 3070: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Initial value of the slider.
+        min_value : float
```
antrikshmisri, line 3072: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Minimum value of the slider.
+        max_value : float
```
antrikshmisri, line 3074: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Maximum value of the slider.
+        padding : float
```
antrikshmisri, line 3076: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Distance between loader and background.
+        bg_color : (float, float, float)
```
antrikshmisri, line 3078: `, optional`

fury/ui/elements.py
```diff
@@ -3051,3 +3052,114 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+            Background color of progress bar.
+        progress_color : (float, float, float)
```
antrikshmisri, line 3080: `, optional`

antrikshmisri (commented):
Hey @ganimtron-10, thanks for the PR, please see my comments below. I think the logic is fine, I just can't think of many asynchronous processes in FURY where a progress bar will be helpful. Can you mention some use cases where this might be helpful. Also, can you add a user hook which will be triggered when all the processes are complete for eg.  removing the progress bar from the scene or some other effect. Thanks!

ganimtron-10:
Thank You @antrikshmisri for the review!!

> Can you mention some use cases where this might be helpful.

Below examples are my understanding about the process, I don't have a good idea about how it works!
- Recently "Creation of Molecules and Proteins" was under work so there if we change any single parameter while interacting with it, we would need to modify the whole molecule/protein again, while this time we could keep user hooked by showing the progress of the work completed using this.
- If in future Fury tries too implement transitioning between the scenes then there also we can use this.
- While we store some data into buffer before presenting it ie. online images or physics simulations data then also this can be used

I have also addressed all the comments and will add a hook as soon as possible.

ganimtron-10:
Hi @antrikshmisri !!
I added an on_complete hook which is called when the progressbar completes all the process. PTAL.
Thanks!

fury/ui/elements.py
```diff
@@ -3051,3 +3052,118 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+        elif progress_width < 0:
+            progress_width = 0
+        self.progress.width = progress_width
+        self.update()
+
+    def update(self):
+        self._ratio = self.progress.width/self.progress_total_size[0]
+
+        value_range = self.max_value - self.min_value
+        self._value = self.min_value + self.ratio * value_range
```
antrikshmisri, line 3169: value is being set to a float instead of an integer. Hence, the value is never equal to its intended value.

fury/ui/elements.py
```diff
@@ -3051,3 +3052,118 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+        self.background.position = coords
+        self.progress.position = coords + self.padding
+
+    @property
+    def value(self):
+        return self._value
+
+    @value.setter
+    def value(self, value):
+        value_range = self.max_value - self.min_value
```
antrikshmisri, line 3147: set `self._value` here instead of `update` and set it to the `value` parameter.

fury/ui/elements.py
```diff
@@ -3051,3 +3052,118 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+        self.progress_total_size = (self.bg_size[0]-2*self.padding,
+                                    self.bg_size[1]-2*self.padding)
+
+        self.background = Rectangle2D(size=self.bg_size, color=self.bg_color)
+        self.progress = Rectangle2D(size=self.progress_total_size,
+                                    color=self.progress_color)
+
+    def _get_actors(self):
+        """Get the actors composing this UI component."""
+        return self.background + self.loader
```
antrikshmisri, line 3114: `self.background.actors` instead of `self.background`. Also, there is no attribute `loader`.

antrikshmisri (changes requested):
Hey @ganimtron-10 , thanks for the update. See my comments below. Your UI is missing the `resize` method. Also, can you demonstrate the use of `on_complete` hook in the tutorial.

ganimtron-10:
Hi @antrikshmisri , I have updated the code as per the comments and also added the `resize` method and `on_complete` hook tutorial.

antrikshmisri (commented):
Hey @ganimtron-10 , thanks for the update. Please see my comments below. Also, could you add a flag to make the UI draggable?

ganimtron-10:
Hi @antrikshmisri ,
All the comments are been addressed and to make `UI` `draggable` and for more flexibility, I changed the background to `Panel2D`.
PTAL.
Thanks!

PR #499, opened 2021-09-14
Adding `SpinBoxUI` to the `UI` module

ganimtron-10:
Creating a SpinBoxUI with tutorials and tests.
Fury-SpinBoxUI


skoudoro (commented):
Hi @ganimtron-10,

Thank you for doing this! Overall, it looks good to me. I need to play a bit with these new UI elements because I am not sure that is draggable. if not, I recommend you add an option for this.

The logic for `set_value` is strange to me.  If I do  `set_value(10)` I would expect my value to be 10 and not in/decremented by 10.  `value` should be a property where you can set whatever you want.  `increment(self)` and `decrement(self)` function should be added and used in your `de/increment_callback`

ganimtron-10:
Hi @skoudoro !
Thanks for the review.

>The logic for set_value is strange to me. If I do set_value(10) I would expect my value to be 10 and not in/decremented by 10. value should be a property where you can set whatever you want. increment(self) and decrement(self) function should be added and used in your de/increment_callback

For this, At first I too used the same approach of creating two different functions but both the functions were replication of each other just differing at in/decrement so I created a single function, but I should have given it more descriptive name!

I will update the code according to the comments! ASAP.
Thanks!

ganimtron-10:
I have updated the callbacks and made `value` as a property. PTAL. Thanks!

skoudoro (commented):
Hi @ganimtron-10,

See below for some comments to improve this UI. it is close to being merged.

- When you change the SpinBox size, there are many alignment issues.
- When I `min_val=10` and we reach it, the cone is still rotating. is it normal? same with `max_val`
- `TextBox2D` is editable. when we edit it on the scene, it would be good to catch this event and rotate the cone. 

Thanks for the future update!

ganimtron-10:
Hi @skoudoro ,
I have updated the code according to comments and also created a new PR for catching the textbox editing complete event.
PTAL.
Thanks!

ganimtron-10:
Hey @skoudoro , 
Let me know if there are any updates on this PR. 
Thanks!

ganimtron-10:
Hey @skoudoro ,
Are there any works still pending on the PR??

skoudoro:
Hi @ganimtron-10,

I will let you know a bit later, we are currently focus on our CI's which are failing too often. After that, I will be back on your PR and all the other.

ganimtron-10:
Hello @m-agour ,
Can you test this PR and let me know if there are any other features that you will require in this `SpinBoxUI` for integrating it into the `Playback Panel`?

dodicono:
> Hello @m-agour , Can you test this PR and let me know if there are any other features that you will require in this `SpinBoxUI` for integrating it into the `Playback Panel`?

Yes, I will test it and let you know.

m-agour:
Hello @ganimtron-10,
Thanks for investing the time to complete this PR. I will test it in my timeline and let you know the results.

skoudoro:
Hi @ganimtron-10,

Can you fix all the conflicts?

@m-agour and @ganimtron-10, Did you have time to look at this PR? 

Thanks for the feedback

ganimtron-10:
Hello @skoudoro ,
All the conflicts are fixed! PTAL.
Thanks!

skoudoro:
Thanks! @ganimtron-10 !

ganimtron-10:
Hello @skoudoro,
I have rebased this PR with the current changes. PTAL.

ganimtron-10:
I have updated this branch!
This will be ready to go once #790 is merged!

skoudoro:
Thank you @ganimtron-10. 

I will be able to look at #790 later this week. From next week, the review rhythm will accelerate and be faster

skoudoro:
What is the update on this PR @ganimtron-10 ? Since, `Textblock` is fixed and merged, this PR should done, finalized and merged this week, right ?

ganimtron-10:
Hey @skoudoro,
This PR is ready for review!
Most of the tests have passed and some are still in progress which would also get done.

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+    def _setup(self):
+        """Setup this UI component.
+
+        Create the SpinBoxUI with Background (Panel2D) and InputBox (TextBox2D)
+        and Increment,Decrement Button (Button2D).
+        """
+        self.panel = Panel2D(size=self.panel_size, color=self.panel_color)
+
+        self.textbox = TextBox2D(width=self.textbox_width,
+                                 height=self.textbox_height)
```
skoudoro, line 4433: you can use a percentage of `self.panel_size` to define width and height  here

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+        if value > self.max_val:
+            self._value = self.max_val
+        elif value < self.min_val:
+            self._value = self.min_val
+        else:
+            self._value = value
+
+    def increment(self):
+        """Increment the current value by the step."""
+        current_val = int(self.textbox.message)
```
skoudoro, line 4533: since textblock is editable, it failed if message is empty. a check need to be added here. 

here the error:
```python
  File "/Users/skoudoro/devel/fury/fury/interactor.py", line 401, in _callback
    callback(self, prop, *args)
  File "/Users/skoudoro/devel/fury/fury/ui/core.py", line 288, in left_button_click_callback
    self.on_left_mouse_button_pressed(i_ren, obj, self)
  File "/Users/skoudoro/devel/fury/fury/ui/elements.py", line 4509, in increment_callback
    self.increment()
  File "/Users/skoudoro/devel/fury/fury/ui/elements.py", line 4533, in increment
    current_val = int(self.textbox.message)
ValueError: invalid literal for int() with base 10: ''
```

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+        if current_val == self.max_val:
+            return
+        self.value = current_val + self.step
+
+        self.textbox.set_message(str(self.value))
+        self.on_change(self)
+
+    def decrement(self):
+        """Decrement the current value by the step."""
+        current_val = int(self.textbox.message)
```
skoudoro, line 4543: same as above: since textblock is editable, it failed if message is empty. a check need to be added here. 

here the error:
```python
  File "/Users/skoudoro/devel/fury/fury/interactor.py", line 401, in _callback
    callback(self, prop, *args)
  File "/Users/skoudoro/devel/fury/fury/ui/core.py", line 288, in left_button_click_callback
    self.on_left_mouse_button_pressed(i_ren, obj, self)
  File "/Users/skoudoro/devel/fury/fury/ui/elements.py", line 4509, in increment_callback
    self.increment()
  File "/Users/skoudoro/devel/fury/fury/ui/elements.py", line 4533, in increment
    current_val = int(self.textbox.message)
ValueError: invalid literal for int() with base 10: ''
```

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+        if current_val == self.min_val:
+            return
+        self.value = current_val - self.step
+
+        self.textbox.set_message(str(self.value))
+        self.on_change(self)
+
+    def textbox_update_value(self, textbox):
```
skoudoro, line 4551: where is it called ?  do you need it ?  you already have `value` setter?

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+        """
+        self.panel_size = size
+        self.padding = padding
+        self.panel_color = panel_color
+        self.min_val = min_val
+        self.max_val = max_val
+        self.value = initial_val
+        self.step = step
+        self.textbox_width = textbox_width
+        self.textbox_height = textbox_height
```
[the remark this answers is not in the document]
ganimtron-10: The name might be ambiguous but these parameter sets how many characters should be there on a line (width) and how many lines (height) the textbox would have. Also being a required parameter for textbox maybe we can't skip.

fury/ui/elements.py
ganimtron-10: same as above

>The name might be ambiguous but these parameter sets how many characters should be there on a line (width) and how many lines (height) the textbox would have. Also being a required parameter for textbox maybe we can't skip.

fury/ui/elements.py
ganimtron-10: when we press the `return` key or get off focused from the `TextBox2D` we used this as a hook!

fury/ui/elements.py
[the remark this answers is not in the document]
skoudoro: I still believe that the user should not have to define that for a spinbox. it should be automatically computed from `self.panel` like you do in `resize` function. Those values are skipped just after the `init` so the user never see this change.

fury/ui/elements.py
skoudoro: Same as above

> I still believe that the user should not have to define that for a spinbox. it should be automatically computed from `self.panel` like you do in `resize` function. Those values are skipped just after the `init` so the user never see this change.

fury/ui/elements.py
```diff
@@ -4368,3 +4370,186 @@ def left_button_dragged(self, i_ren, _obj, _sub_component):
...
+        Create the SpinBoxUI with Background (Panel2D) and InputBox (TextBox2D)
+        and Increment,Decrement Button (Button2D).
+        """
+        self.panel = Panel2D(size=self.panel_size, color=self.panel_color)
+
+        self.textbox = TextBox2D(width=self.textbox_width,
+                                 height=self.textbox_height)
+        self.textbox.text.dynamic_bbox = False
+        self.textbox.text.auto_font_scale = True
+        self.textbox.set_message(str(self.value))
```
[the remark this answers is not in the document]
ganimtron-10: - This textbox uses `auto_font_scale` which is why having smaller value might make font look bigger at start but if you put a big number it automatically scales itself to fit inside.

- Yes Textbox is editable! Having taken reference from other UI libraries, I made the textbox editable as it helps in directly inserting the required value.
Let me know what I should do with it! (personally according to me I feel it should be editable by default for increased accessibility and generalization of UI)

- Yes will work on the integer validating part!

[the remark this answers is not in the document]
skoudoro: > This textbox uses auto_font_scale which is why having smaller value might make font look bigger at start but if you put a big number it automatically scales itself to fit inside.

I mean, the font touch the gray area on the top, but plenty of white space in the bottom. it should  be centered and not touching any gray area.

> Yes Textbox is editable! Having taken reference from other UI libraries, I made the textbox editable as it helps in directly inserting the required value.
Let me know what I should do with it! (personally according to me I feel it should be editable by default for increased accessibility and generalization of UI)

ok, fine. we let it as it is. Just make sure that some value or character are impossible. Now, we can write word and make it crash.

fury/ui/elements.py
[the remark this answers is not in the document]
ganimtron-10: Maybe I was not clear in the last comment!

See below I have attached two pictures
1. textwidth = 15 (ie.15 character can fit in a line which is A - O) && textheight = 5 (ie. 5 lines of character can accomodate)
image

2. textwidth = 10 (ie.10 character can fit in a line which is H - Q) && textheight = 2 (ie. only 2 lines of character)
image






I think I will rename it to textheight and textwidth it seems more appropriate now!

ganimtron-10:
Hey @skoudoro , I have updated the validating part for the textbox input and the tests completed successfully.

Regarding the text touching the borders, I checked the positions for both the text actor and the background, and they are at the same position. 
I am not sure how we should proceed with this, but an approach that comes to my mind is 
- to move the background a bit up so the text aligns with the background 
- or maybe add a new parameter called padding which defines the padding around the text so that the background is always bigger than the text actor.

I checked other elements too and this issue is not specific to `SpinBoxUI`, it is a general issue with the `TextBlock2D` and `TextBox2D` which was visible because of the larger size of the `SpinBoxUI`. Maybe we need to handle this separately to keep this PR contained to the specified scope.

skoudoro:
>  I have updated the validating part for the textbox input and the tests completed successfully.

Thank you for the update. I will look into it in a 1h

> I checked other elements too and this issue is not specific to SpinBoxUI, it is a general issue with the TextBlock2D and TextBox2D which was visible because of the larger size of the SpinBoxUI. Maybe we need to handle this separately to keep this PR contained to the specified scope.

Ok, I see, can you start a PR to fix that. Let's try to fix this issue this week. we can do a pair coding session on Friday to finalize what you will start today.

skoudoro (commented):
Hi @ganimtron-10,

All good! Just missing one thing:

can you add your example inside the ui section of [_valid_examples.toml](https://github.com/fury-gl/fury/blob/master/docs/examples/_valid_examples.toml)

Then we can merge the PR and create a new PR concerning issue with the TextBlock2D and TextBox2D

PR #501, opened 2021-09-18
Creating an `off_focus` hook in `TextBox2D`

ganimtron-10:
To know when the user has completed editing in the `TextBox2D`, creating an `on_blur` hook.
Required in #499 .
Also changing the condition sequence in the `handle_character` function,  because when we press "return" key instead of removing focus from the UI it adds the "\r" character to the message.


Garyfallidis:
Please use ``focus_in`` or ``focus_out`` instead of ``on_blur``. Alternatively, ``off_focus/on_focus``.

Garyfallidis:
A test is missing too.

ganimtron-10:
Hey @skoudoro ,
I have updated the hook name and also added the tests for the same.
PTAL.
Thanks!

skoudoro (approved):
Thank you for this PR @ganimtron-10. 

Looks good to me. merging

PR #489, opened 2021-08-11
This PR creates an actor to draw labels using texture maps

devmessias:
- I've created  the actor and tests in different files just for help me to develop this PR. But latter I'll put everything inside of  actors.py and test_actors.py. 

To create pre-built texture atlas for FURY you just need to put the ttf files inside of `data/files` and run the following command
```bash
python -c 'from fury import text_tools;text_tools._create_fury_system_atlas_fonts()'
```

The example 
[docs/examples/viz_billboard_labels.py](https://github.com/fury-gl/fury/blob/c05b1c06e84be97a32abf84565359dd4dfda3c13/docs/examples/viz_billboard_labels.py)
shows almost all the behaviors available for this PR

Result with bitmap texture
Screenshot_20210815_134435

Result with SDF
image

PR #507, opened 2021-10-01
Deprecate and rename label to vector_text

skoudoro:
This is a follow-up of [this comment](https://github.com/fury-gl/fury/pull/489#discussion_r719525377) from #489.

The goal is to:
- rename `label(...)` to `vector_text(...)`
- deprecate `label` and warn the users of the change

### DIFF
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -226,7 +226,7 @@ def test_contour_from_roi(interactive=False):
 
     npt.assert_equal(report.objects, 1)
     npt.assert_equal(report2.objects, 2)
-   
+
 
 @pytest.mark.skipif(skip_osx, reason="This test does not work on macOS + "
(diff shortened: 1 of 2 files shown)


fury/tests/test_actors.py
```diff
@@ -804,7 +804,8 @@ def test_points(interactive=False):
 
 
 def test_labels(interactive=False):
-    text_actor = actor.label("Hello")
+    npt.assert_warns(DeprecationWarning, actor.label, "hello")
```
Garyfallidis, line 807: Hello?

Garyfallidis (commented):
See above.

Garyfallidis:
One of the tests is failing in relation to slicer I think. Can you check? Is that a random error?

Garyfallidis:
All green! Let's go! Merging ...

PR #511, opened 2021-11-28
Flocking-simulation using boid rules

RajPaliwal10:
Hi here is the Flocking Simulation project using boid rules in python please leave any suggestions or issues in the project. It will definitely help me to learn from my mistakes.


RajPaliwal10:
Thanks for the quick reply I have to work a lot on this.

PR #514, opened 2021-12-15
[FIX] Radio button and checkbox tests

skoudoro:
This PR update some tests that were failing often:
- radio button (tests needed to be recorded again)
- checkbox (tests needed to be recorded again)
- *.tif format on IO

It should fix Windows and MacOSx CIs


Garyfallidis:
Windows and Mac okay. Linux is next!

PR #515, opened 2021-12-16
Changing how we do things with our test suite.

Garyfallidis:
This PR aims to resolve multiple issues:

a) The circular import issue with DIPY. See also https://github.com/dipy/dipy/pull/2493
b) Fixing the bad window issues we get from time to time during testing.
c) Solves an issue with windows staying open and not closing as fast as they should during tests.
d) Finally, this PR is also dealing with a segmentation fault issue that pytest has when running multiple files at once.
For more information see here https://github.com/pytest-dev/pytest/issues/3672
Our current solution for the later is to run pytest on individual files. Pytest claims that this is an issue with software opening threads but this is to be seen.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -925,6 +925,110 @@ def _makeNd(array, ndim):
     return array.reshape(new_shape)
 
 
+def _roll_evals(evals, axis=-1):
+    """Check evals shape.
+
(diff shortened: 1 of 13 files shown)


skoudoro:
Thanks @Garyfallidis

PR #516, opened 2021-12-31
Adding Rhombicuboctahedron actor

PrayasJ:


PrayasJ:
@skoudoro I have added the actor and primitive for rhombicuboctahedron and tests for it. Could you give it a look? thanks.

skoudoro:
Hi @PrayasJ,

The rhombicuboctahedron was already existing apparently. What is the difference between the previous version and your update? Why did you choose to update the rhombicuboctahedron instead of adding a new shape?

Thank you for this and your future feedback

PrayasJ:
@skoudoro I decided to go off with this mostly because I tried creating an actor for a primitive that wasn't already there and the primitive for rhombicuboctahedron was not functioning as it was supposed to so I decided to rewrite that. That was all my reason for it. The previous version wasn't really rendering properly (probably due to the rotation issues on some faces).

SunTzunami:
Hi @PrayasJ ! Thanks for this PR. 
It'll be helpful if you could attach some images for reference which show how the previous version looked and how does your modified, new version look.  
Thanks in advance!

PrayasJ:
Hi @SunTzunami, the rhombicuboctahedron primitive looked like 
Screenshot (111) if I attach it to the actor function I wrote and the the primitive I wrote looked like 
Screenshot (110)

fury/tests/test_primitive.py
```diff
@@ -77,6 +77,16 @@ def test_vertices_primitives_triangularprism():
...
 
+def test_vertices_primitives_rhombicuboctahedron():
+    # Testing the default vertices of the primitive rhombicuboctahedron.
+    vertices, _ = fp.prim_rhombicuboctahedron()
+    shape = (24, 3)
+    npt.assert_equal(vertices.shape, shape)
+    npt.assert_equal(np.mean(vertices, axis=0), 0)
+    npt.assert_equal(vertices.min(), -0.5)
+    npt.assert_equal(vertices.max(), 0.5)
+
```
skoudoro, line 88: You can remove this. this is a duplicate. This is already done line 13

skoudoro (commented):
Thank you @PrayasJ,

Overall, looks good to me. See below for the comment. After that, it should be ready to be merged

fury/tests/test_primitive.py
PrayasJ: Thanks for pointing this out, removed that. @skoudoro

skoudoro:
Thanks for the update, merging!

Issue #423, opened 2021-04-26
The axes actor needs improvements.


amitchaudhari9121:
### **Is your feature request related to a problem? Please describe.**
I've found that **it is difficult to distinguish the X Y and Z using the color coded axes actor**, I am pretty sure others must have been confused by this as well.

Apart from this. there's also the issue that it conveys very little information i.e only orientation information. I'll make another comment detailing this.

### **Describe the solution you'd like**
I propose that we annotate the azes with the respective letters. easy enough.
Example:
![Screenshot 2021-04-26 at 12 53 10 PM](https://user-images.githubusercontent.com/21289530/116044281-5cf41900-a68e-11eb-8255-89d42c18631e.png)

### **Describe alternatives you've considered**
Other alternatives could be a blender style Axes actor,(personally, I think our's is better)

![Screenshot 2021-04-26 at 12 57 03 PM](https://user-images.githubusercontent.com/21289530/116044738-e7d51380-a68e-11eb-9c16-5673d16afa84.png)

or a 6-axis alternative like this, but with annotations.
![Screenshot 2021-04-26 at 12 58 40 PM](https://user-images.githubusercontent.com/21289530/116044936-24087400-a68f-11eb-8433-6fcd4b9ac5da.png)

### **Additional context**
I'll be submitting a PR really soon that brings these changes, and also for the ones i mention below.

Cheers!

amitchaudhari9121:
>Apart from this. there's also the issue that it conveys very little information i.e only orientation information.
>I'll make another comment detailing this.

I expect reference axes to give more information, 
- Title of that axis,
- Labels at certain points,
- scale or "ticks" at regular distance intervals.

PR #517, opened 2022-01-05
Added annotations to axes as requested in #423 

PrayasJ:
The axes seemed to be confusing to me as someone new since I was confused which direction I was facing, so felt like adding X, Y and Z labels to the axes might help out a bit while looking at the models. Images attached for changes:

without labels:
Screenshot (117)

with labels:
Screenshot (116)

Also added a function in utils to combine an array of vtkActors into a vtkAssembly.


skoudoro (commented):
Thanks @PrayasJ,

I think labels should be optional. Can you add it (boolean option)?
We need to test it a bit. 
Could you also add a test for your function `combine_actors`.
Thanks

PrayasJ:
@skoudoro added the test and made the labels optional.

skoudoro (commented):
Hi @PrayasJ,

It would be great if the label follow de camera. We do not start a tutorial / Application with always the same camera angle and in this case, the labels are not always visible.

Can you update it?

Thanks!

PrayasJ:
@skoudoro I was actually trying to do that previously, the label actor is supposed to follow the camera right? I tried using it but it wasn't doing that while I was testing it. I think I would need to fix that up too. (or maybe I am using them wrong)

PrayasJ:
Screenshot (152)

It follows the camera now, forgot to upload the screenshot before.

PR #518, opened 2022-01-05
Pytest patch

skoudoro:
Run Pytest on individual files instead of package


Garyfallidis:
Coverage will be corrected in subsequent PRs.

Issue #467, opened 2021-07-20
Window snapshot inverts the displayed scene


jhlegarreta:
## Description

FURY's `window.snapshot` method seems to save a given scene with the wrong orientation. The scene is displayed with a given orientation when calling `window.show` (and this matches the appropriate anatomical orientation in this case), but the scene is inverted upside down when saving to a file.

## Way to reproduce

A self-contained script to reproduce the issue has been posted in:
https://gist.github.com/jhlegarreta/f13446716c1cff8df86c504f0b71a7d9

Relevant images below.

Environment:

```
'fury_version': '0.7.0.post51+g2e7971a', 
commit_hash': '2e7971a176c2540e10a9a6da861097583d08cb4a'
'sys_version': '3.8.10 (default, Jun  2 2021, 10:49:15) \n[GCC 9.4.0]'
'sys_platform': 'linux'
'numpy_version': '1.21.1'
'scipy_version': '1.7.0'
'vtk_version': '9.0.3'
'matplotlib_version': '3.4.2'
'dipy_version': '1.4.1'
```

Happens as well with FURY 0.6.1 using the same environment. Did not happen with previous versions, but might have to do with some change in VTK as well. I do not have an environment that is not showing the issue now, since I recently upgraded my workstation and e.g. VTK 8.x have no wheels for Python 3.8.

In order to avoid these issues, it would be useful to add one of the DIPY structural MRI test volumes to FURY's test data, and once the issue has been fixed, to generate a baseline image that is compared against every time tests are run.

- [X] Code example
- [X] Relevant images (if any)
- [X] Operating system and versions (run `python -c "from fury import get_info; print(get_info())"`)

Relevant images below (sizes have been reduced to avoid taking an excessive space):

- A manual screenshot of the displayed scene:
![fury_shown_scene](https://user-images.githubusercontent.com/5576557/126388710-c90bb7ce-4ba6-40cb-872c-835b20031d07.png)

- The `PNG` file saved by the `window.snapshot` method:
![scene_snapshot](https://user-images.githubusercontent.com/5576557/126389165-5cca0a18-916a-479f-9a33-952141f7c5d0.png)

skoudoro:
Thank you for the report @jhlegarreta

jhlegarreta:
I confirm that #521 fixed the issue. Thanks for the effort @Garyfallidis.

PR #521, opened 2022-01-13
Snapshot flipping bug fix

Garyfallidis:
This PR fixes the issue reported in https://github.com/fury-gl/fury/issues/467

Basically, we were not flipping the images correctly in ``save_image`` and ``load_image``. ``save_image`` was called by ``snapshot``.

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -77,12 +77,11 @@ def load_image(filename, as_vtktype=False, use_pillow=True):
                                 0, 0)
             vtk_image.SetSpacing(1.0, 1.0, 1.0)
             vtk_image.SetOrigin(0.0, 0.0, 0.0)
-            arr_tmp = np.flipud(image)
-            arr_tmp = image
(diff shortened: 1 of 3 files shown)


fury/io.py
```diff
@@ -77,6 +78,7 @@ def load_image(filename, as_vtktype=False, use_pillow=True):
             vtk_image.SetSpacing(1.0, 1.0, 1.0)
             vtk_image.SetOrigin(0.0, 0.0, 0.0)
             arr_tmp = np.flipud(image)
```
skoudoro, line 80: Can you remove this line?  It is not used since you do `arr_tmp = image` in the line below (81)

skoudoro (commented):
LGTM! I will merge it as soon as you fix my comment below. Thanks!

fury/io.py
Garyfallidis: Done!

skoudoro:
Failing CI's not related, thank @Garyfallidis. merging

PR #527, opened 2022-01-19
[FIX] Allow sphere actor to use faces/vertices without casting issues. In addition, update versioning system (versioneer).

### LINKED ISSUE #400: Sphere actor does not appear when vertices and faces are used
## Description
When I use 'theta' and 'phi' parameters in sphere actor, I can see the spheres in the window. But I don't see any spheres when I use 'vertices' and 'faces' parameters. 

Here is the code sample to verify the issue:
```
from fury import actor, primitive, window
import fury
print(fury.__version__)
print(fury.__file__)
import numpy as np

scene = window.Scene()
sphere_res = 'symmetric362'
vertices, faces = primitive.prim_sphere(name=sphere_res,
                                        gen_faces=False)
pos = 10 * np.random.rand(100, 3)
colors = 100 * np.ones((100, 3), dtype='uint8')
radii_spheres = 2+np.random.rand(100)
sphere_actor = actor.sphere(centers=pos,
                                colors=colors,
                                radii=radii_spheres,
                                vertices=vertices,
                                faces=faces)
scene.add(sphere_actor)
window.show(scene)
```
Thank you all!

skoudoro:
This PR fixes #400. The sphere actor was not working with the primitives' sphere due to casting. Indeed, all the sphere vertices are float values between [-1;1]. They were converted to zero due to the casting. I suppose the casting was there to save memories. 

Thanks @SunTzunami for the introspection. 

I use this opportunity to:
- [x] Update versioneer
- [x] remove deprecated `distutils` 
- [x] add packaging as a dependency


Garyfallidis:
Please next time separate this PR in two PRs. 
Also after the release we should make sphere, cylinder and cone support repeat_primitives and not repeat_sources. This will allow FURY to have the same picking/selection experience.

PR #533, opened 2022-01-22
Sphere actor uses repeat_primitive by default

### LINKED ISSUE #528: Sphere actor needs to use repeat_primitives by default
This will allow sphere to have the same behavior when picking vertices or edges as with other primitives. 

However, we should also keep repeat_sources as a secondary option just in case we need to use some of the other VTK filters.

See example of box actor (using repeat_primitives) here https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1696

See current implementation of sphere here https://github.com/fury-gl/fury/blob/master/fury/actor.py#L1511

xtanion:
Resolves #528


fury/actor.py
```diff
@@ -1545,16 +1545,27 @@ def sphere(centers, colors, radii=1., phi=16, theta=16,
...
+
+    else:
+        scales = np.multiply(radii, [1, 1, 1])
+        directions = (1, 0, 0)
+        verts, faces = fp.prim_sphere('symmetric724', False)
+        res = fp.repeat_primitive(verts, faces,
+                                  directions=directions, centers=centers,
+                                  colors=colors, scales=scales)
+        big_verts, big_faces, big_colors, _ = res
+        actor = get_actor_from_primitive(big_verts, big_faces, big_colors)
```
skoudoro, line 1568: In this case, `phi` and `theta` and `opacity` are not used which is not good. 

- You will have to update the `prim_sphere` function to generate the point of a sphere following `phi` and `theta`
- You will have to manage the opacity also.

fury/tests/test_actors.py
```diff
@@ -818,14 +818,15 @@ def test_labels(interactive=False):
     npt.assert_equal(scene.GetActors().GetNumberOfItems(), 1)
 
 
-def test_spheres(interactive=False):
+@pytest.mark.parametrize('interactive, prim', [(False, True), (False, False)])
```
skoudoro, line 821: We prefer to avoid using `pytest.mark.parametrize` so you can remove it and use a more traditional way

skoudoro (commented):
Hi @xtanion,

Thank you for this PR. See below for the comments.

fury/actor.py
xtanion: Thanks @skoudoro, I will be changing the `prim_sphere` function to use `phi` and `theta`. However, the opacity of the sphere is managed by the following line :
```py
actor.GetProperty().SetOpacity(opacity)
```
And opacity can also be set through colors parameter by passing (R,G,B,A) values

skoudoro: Indeed,  good catch, so forgets the opacity and focus just on the updating `prim_shere`.

xtanion: In the `prim_sphere` method, it gets `vertices` and `faces` from`.npz` file and akes `name` and `gen_faces` parameters, should I replace it or make it optional to use `name` and `gen_faces` parameter.

skoudoro: you should keep it, it is really important

xtanion: Okay

skoudoro: > I noticed faces of the sphere are quadrilateral, Is it fine to go with?

and yes, it is fine. As an option, we should be able to increase the number of points (density) and the quadrilateral will be smaller.

xtanion: Hi @skoudoro, made the required changes to the `prim_sphere`. Can you please have a look?

skoudoro:
Thanks @xtanion, I will look into it sometime this week and get back to you.

skoudoro (commented):
Hi @xtanion,

See below for a quick and initial review. 

Thanks!

fury/primitive.py
```diff
@@ -290,15 +290,42 @@ def prim_sphere(name='symmetric362', gen_faces=False):
...
-    fname = SPHERE_FILES.get(name)
-    if fname is None:
-        raise ValueError('No sphere called "%s"' % name)
-    res = np.load(fname)
-
-    verts = res['vertices'].copy()
-    faces = faces_from_sphere_vertices(verts) if gen_faces else res['faces']
-    faces = fix_winding_order(res['vertices'], faces, clockwise=True)
-    return res['vertices'], faces
+    if phi is None:
```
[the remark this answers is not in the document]
xtanion: Thanks @skoudoro for the review, I'll be changing that to`if not phi or not theta`

fury/actor.py
```diff
@@ -1515,7 +1515,7 @@ def point(points, colors, point_radius=0.1, phi=8, theta=8, opacity=1.):
 
 
 def sphere(centers, colors, radii=1., phi=16, theta=16,
-           vertices=None, faces=None, opacity=1):
+           vertices=None, faces=None, opacity=1, prim=True):
```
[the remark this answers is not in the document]
xtanion: Okay, I'll change that.

fury/actor.py
```diff
@@ -1551,16 +1551,28 @@ def sphere(centers, colors, radii=1., phi=16, theta=16,
...
-        src.SetThetaResolution(theta)
-        src.SetPhiResolution(phi)
+        if src is not None:
+            src.SetRadius(1)
+            src.SetThetaResolution(theta)
+            src.SetPhiResolution(phi)
+
+        sphere_actor = repeat_sources(centers=centers, colors=colors,
+                               active_scalars=radii, source=src,
+                               vertices=vertices, faces=faces)
```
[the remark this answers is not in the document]
xtanion: Okay, I'll make the changes.

fury/primitive.py
```diff
@@ -290,15 +290,42 @@ def prim_sphere(name='symmetric362', gen_faces=False):
...
+        if fname is None:
+            raise ValueError('No sphere called "%s"' % name)
+        res = np.load(fname)
+
+        verts = res['vertices'].copy()
+        faces = faces_from_sphere_vertices(
+            verts) if gen_faces else res['faces']
+        faces = fix_winding_order(res['vertices'], faces, clockwise=True)
+    else:
+        phi_indices, theta_indices = np.arange(0, phi), np.arange(1, theta-1)
```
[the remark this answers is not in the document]
xtanion: I tried using `(0, theta)` but in that case, the top and bottom vertices were created multiple times, Here I'm adding the top and bottom vertices (`_points`) after all of the other vertices are created.

The minimum value of `theta` can be `3`. At `theta <=2` won't be a 3D shape.
In `sphere_actor` (using SphereSource) if we decrease `theta` or `phi` to  less than 3, it returns the actor with `theta=3` or `phi=3`.

xtanion:
Hii @skoudoro , made the following changes, PTAL
- Modified conditions of `prim_sphere` to `if not phi or not theta`
- `use_primitive` instead of `prim`
- return `sphere_actor` and `verts, faces` before the else statement

fury/primitive.py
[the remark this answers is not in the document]
xtanion: Hi @skoudoro, I've added the checks for `phi` and `theta`. Compared with `sphere_actor`, this is how it looks at `theta` and `phi` equal to 0.

https://user-images.githubusercontent.com/74976752/153140745-2bcc33cc-0e11-4623-a7b4-d57f37b792a0.mp4

Is it the correct way of implementing this?

skoudoro:
Hi @xtanion,

Can you rebase this PR? Also, my previous comments on the cone PR can be applied to this PR also. Can you update this PR with my previous comments.

Thank you in advance

xtanion:
Thanks for the review @skoudoro, I made the necessary changes and added a demo.
PTAL, Thanks

skoudoro (commented):
Hi @xtanion,

Thank you for the update. Overall, it looks good. I will be able to merge your PR after you address the last comment below. 

Thanks!

docs/tutorials/01_introductory/viz_sphere.py
```diff
@@ -0,0 +1,46 @@
+"""
+======================================================================
+Fury sphere Actor
+======================================================================
+This example shows how to use both primitive & vtkSource sphere actor.
```
skoudoro, line 5: `and` instead of `&`

docs/tutorials/01_introductory/viz_sphere.py
```diff
@@ -0,0 +1,46 @@
...
+# First thing, you have to specify centers and colors of the sphere
+
+centers = np.zeros([1, 3])
+colors = np.random.rand(1, 3)
+
+############################################################################
+# The below sphere actor is generated by repeating the sphere primitive.
+
+prim_sphere_actor = actor.sphere(centers, colors=colors, radii=5,
+                                 phi=30, theta=30)
```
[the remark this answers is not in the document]
xtanion: There's no specific reason for that, removed it.

skoudoro (approved):
Thank you for this @xtanion, merging

PR #540, opened 2022-02-02
updating methods to use `numpy_to_vtk_image_data` from utils

ganimtron-10:
Updating the load_image method to directly convert the image into vtk_imagedata using the utils function

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -1152,15 +1152,8 @@ def rgb_to_vtk(data):
     vtkImageData
 
     """
-    grid = ImageData()
-    grid.SetDimensions(data.shape[1], data.shape[0], 1)
-    nd = data.shape[-1]
-    vtkarr = numpy_support.numpy_to_vtk(
-        np.flip(data.swapaxes(0, 1), axis=1).reshape((-1, nd), order='F')
-    )
-    vtkarr.SetName('Image')
-    grid.GetPointData().AddArray(vtkarr)
-    grid.GetPointData().SetActiveScalars('Image')
+    grid = numpy_to_vtk_image_data(data)
+    grid.GetPointData().GetScalars().SetName('Image')
     grid.GetPointData().Update()
     return grid
 
--- a/fury/io.py
+++ b/fury/io.py
@@ -125,27 +125,7 @@ def load_image(filename, as_vtktype=False, use_pillow=True):
                 image = np.asarray(pil_image)
 
         if as_vtktype:
-            if image.ndim not in [2, 3]:
-                raise IOError('only 2D (L, RGB, RGBA) or 3D image available')
-
-            vtk_image = ImageData()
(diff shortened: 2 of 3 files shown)


skoudoro (commented):
Thank you @ganimtron-10 for updating this.

Can you update all the other functions also?

`actor.slicer`, `actor.contour_from_roi`,  `io.save_image`, `io.load_image`, `utils.rgb_to_vtk`,
`test_windows.test_scene`, `test_window.test_skybox`.

ganimtron-10:
Hi @skoudoro , 

This pr isn't yet ready for review, still some changes are remaining! 
I will complete it ASAP and let you know.

Thanks!

ganimtron-10:
Hey @skoudoro ,

I have updated the functions mention above.
`io.save_image`, `io.load_image`, `actor.slicer`, `actor.contour_from_roi` methods are working fine according to the previous tests, but I was unable to test the `rgb_to_vtk` as it doesn't have any test and the `test_window.*` method was throwing the following error
```
npt.assert_equal(scene.GetSphericalHarmonics(), None)
E       AttributeError: 'Scene' object has no attribute 'GetSphericalHarmonics'
```
PTAL

Thanks!

fury/utils.py
```diff
@@ -170,23 +172,17 @@ def numpy_to_vtk_image_data(array, spacing=(1.0, 1.0, 1.0),
...
-    temp_arr = np.flipud(array)
-    temp_arr = temp_arr.reshape(array.shape[1] * array.shape[0], depth)
-    temp_arr = np.ascontiguousarray(temp_arr, dtype=array.dtype)
+
     vtk_array_type = numpy_support.get_vtk_array_type(array.dtype)
-    uchar_array = numpy_support.numpy_to_vtk(temp_arr, deep=deep,
+    uchar_array = numpy_support.numpy_to_vtk(array, deep=deep,
                                              array_type=vtk_array_type)
     vtk_image.GetPointData().SetScalars(uchar_array)
     return vtk_image
```
skoudoro, line 188: I am not sure what you are doing here. Why only 2D image?

the previous code is good so I do not understand why you modify it and add dimension...

skoudoro (commented):
Hi @ganimtron-10,

Can you rebase your PR and address my comment below. There is one step I do not understand in this PR.

fury/utils.py
ganimtron-10: When working with the `actor.slicer` and `actor.contour_from_roi` we are getting 1D array and in other cases we are getting 2D arrays so to make the function more generic I kept all the array conversion logic outside the function due to which I changed the if condition.
Similarly each method was setting different dimension for the ImageData so made that as a property.

skoudoro: I am not sure about this approach, still thinking and checking some stuff. Thank you for your feedback

PR #544, opened 2022-02-04
Improve setuptools

skoudoro:
We were using setuptools in half so this PR goal is to migrate completely to setuptools


Garyfallidis:
Great! Merging.

Issue #553, opened 2022-03-09
Refresh code of all tutorials and demos


skoudoro:
Many tutorials and demos are outdated. We need to update by using more `fury.utils` functions (1 tutorials, demos / Pull Request)

skoudoro:
This is a too generic issue. Closing for now

PR #556, opened 2022-03-16
Updated code of `viz_network_animated` to use `fury.utils`

xtanion:
Refreshed code of demo `viz_network_animated` as per #553 to use `fury.utils` function.


docs/examples/viz_network_animated.py
```diff
@@ -122,8 +122,8 @@ def new_layout_timer(showm, edges_list, vertices_count,
     b = 1.0
     deltaT = 1.0
 
-    sphere_geometry = np.array(numpy_support.vtk_to_numpy(
-        sphere_actor.GetMapper().GetInput().GetPoints().GetData()))
+    sphere_geometry = np.array(get_polydata_vertices(
+        sphere_actor.GetMapper().GetInput()))
```
skoudoro, line 126: `vertices_from_actor(sphere_actor).copy()` instead. you will have to import the function from `utils`

docs/examples/viz_network_animated.py
```diff
@@ -188,20 +188,20 @@ def _timer(_obj, _event):
             iterate(1)
         else:
             pos[:] += (np.random.random(pos.shape) - 0.5) * 1.5
-        spheres_positions = numpy_support.vtk_to_numpy(
-            sphere_actor.GetMapper().GetInput().GetPoints().GetData())
+        spheres_positions = get_polydata_vertices(
+            sphere_actor.GetMapper().GetInput())
```
skoudoro, line 192: same comment as above

docs/examples/viz_network_animated.py
```diff
@@ -188,20 +188,20 @@ def _timer(_obj, _event):
...
         spheres_positions[:] = sphere_geometry + \
             np.repeat(pos, geometry_length, axis=0)
 
-        edges_positions = numpy_support.vtk_to_numpy(
-            lines_actor.GetMapper().GetInput().GetPoints().GetData())
+        edges_positions = get_polydata_vertices(
+            lines_actor.GetMapper().GetInput())
```
skoudoro, line 197: same comment as above

skoudoro (commented):
Hi @xtanion,

Thank you very much for this first update.

1. Can you remove all parenthesis in all the `if` statement.
2. replace the `ComputeBounds()` with the appropriate function in `fury.utils``
3. Replace `ResetCameraClippingRange()`
4 test that we have the exact same result.

Thanks

docs/examples/viz_network_animated.py
xtanion: done @skoudoro, PTAL
thanks

Issue #548, opened 2022-02-25
Black window on screen on "window.record".


brunodireito:
## Description
Black window on screen on "window.record". Every time I run the window.record command, a black screen opens on screen (a black patch with the size of the image).

## Way to reproduce
- [ e.g. Tutorial #1 ] 
- ubuntu20.04, jupyter-lab
- [ ] Operating system and versions (run `python -c "from fury import get_info; print(get_info())"`)
{'fury_version': '0.7.1', 'pkg_path': '/home/xxxx/anaconda3/envs/scikit_py3/lib/python3.9/site-packages/fury', 'commit_hash': '426c15b9e4895bdba449be3c63168b5179a34202', 'sys_version': '3.9.7 (default, Sep 16 2021, 13:09:58) \n[GCC 7.5.0]', 'sys_executable': '/home/xxx/anaconda3/envs/scikit_py3/bin/python', 'sys_platform': 'linux', 'numpy_version': '1.21.2', 'scipy_version': '1.7.3', 'vtk_version': '9.0.2', 'matplotlib_version': '3.3.4', 'dipy_version': '1.4.1'}
vtkDebugLeaks has found no leaks.

skoudoro:
Solved by #564, and should be present in the next release.  closing

Issue #196, opened 2020-03-11
window.snapshot() memory leak?


nmnaughton:
I am new to fury and am trying to use it to make a video of streamlines that are changing in time. To do this I am saving a .png for every frame and then later stitching them together using ffmpeg. 

When I save the images by looping through streamlines for each timestep, it seems like fury does not release the memory associated with saved images. This leads to memory accumulation as the loop runs, eventually crashing my computer. 

Not sure if this is expected behavior or a bug. If expected behavior, how can I free the memory so I can run the loop for an arbitrary number of iterations?

The MWE code below ends up using ~7 Gb of memory by the time it finishes and the memory usage is proportional to the value of the time loop. This behavior occurs for both window.snapshot() and window.record().

```python
from fury import window, actor
import numpy as np

t = np.linspace(-10, 10, 100)
bundle = []
for i in np.linspace(3, 5, 1000):
    pts = np.vstack((np.cos(2 * t/np.pi), np.zeros(t.shape) + i*10, t )).T
    bundle.append(pts)

ren = window.Scene()
for time in range(200):#len(plot_params["rod0"]["time"])):
    ren.clear()
    bundle_actor = actor.streamtube(bundle, window.colors.red, linewidth=0.01)
    ren.add(bundle_actor)
    ren.SetBackground(*window.colors.white)
    image_array = window.snapshot(
        ren, fname='tmp/file%02d.png' % time, size=(1000, 1000))
```

Thanks!

skoudoro:
Hi @nmnaughton,

Thank you for this nice report and your example. We will look into it and let you know as soon as possible.

Can you give us your FURY version?

We know that a list of Numpy array creates a memory leak. This is why we created the `ArraySequence` object on Nibabel package or the sister `Streamlines` object on DIPY.

So it could be the main reason for your problem but we need to check to be sure.

nmnaughton:
Thanks for your help on this. 

I am using version 0.4.0

I get similar behavior if I use `bundle = dipy.tracking.streamline.Streamlines()` instead of `bundle = []` so I am not sure if that is it. 

Also, the memory size is substantially larger than 200x the size of my streamline data or the size of the saved images (18 MB total), so it seems some intermediary state is accumulating the memory. 

My current work around is to just manually run the loop in batches with a small enough loop size that the memory size doesn't get too large.  I could also probably write a work around where I run the batches with a child process, which should then release the memory, but neither are very fast or scalable options.

Garyfallidis:
[...]

PR #564, opened 2022-03-31
Record should not make the window appear

Garyfallidis:
This PR allows ``fury.window.record`` to go offscreen as it should, and not create any windows while saving images. 

This is resolving the memory issue when repeatedly calling record. And the fact that you see a black window for a few seconds.

See relevant issues here:
https://github.com/fury-gl/fury/issues/548
https://github.com/fury-gl/fury/issues/196

```
import numpy as np

t = np.linspace(-10, 10, 100)
bundle = []
for i in np.linspace(3, 5, 1000):
    pts = np.vstack((np.cos(2 * t/np.pi), np.zeros(t.shape) + i*10, t )).T
    bundle.append(pts)

ren = window.Scene()
for time in range(200):#len(plot_params["rod0"]["time"])):
    ren.clear()
    bundle_actor = actor.streamtube(bundle, window.colors.red, linewidth=0.01)
    ren.add(bundle_actor)
    ren.SetBackground(*window.colors.white)
    # image_array = window.snapshot(
    #     ren, fname='/tmp/file%02d.png' % time, size=(1000, 1000))

    window.record(ren, out_path='/tmp/file%02d.png' % time, size=(1000, 1000))
```

A note here to say that we do not see anymore the memory leak on either snapshot or record (with this PR merged).

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -701,7 +701,10 @@ def record(scene=None, cam_pos=None, cam_focal=None, cam_view=None,
     az_ang : float, optional
         Azimuthal angle of camera rotation.
     magnification : int, optional
-        How much to magnify the saved frame. Default is 1.
+        How much to magnify the saved frame. Default is 1. A value greater
+        than 1 increases the quality of the image. However, the output
+        size will be larger. For example, 200x200 image with magnification
+        of 2 will be a 400x400 image.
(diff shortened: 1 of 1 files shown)


skoudoro:
Thanks! merging

PR #571, opened 2022-04-12
Fix vertex order in prim_tetrahedron

rkharsan:
Currently, 2 of the faces returned by `primitive.prim_tetrahedron()` have the vertices in incorrect order. This leads to incorrect normals being calculated, and the two faces in question appear invisible from the outside.

Current version: Note the missing faces!
Screenshot from 2022-04-12 20-58-20

Fixed version: All faces are now visible from the outside.
Screenshot from 2022-04-12 20-57-33

Minimal reproducible example:
```python
from fury import utils, window, primitive

v, f = primitive.prim_tetrahedron()

# Uncomment to see the fix
# f[1][1], f[1][2] = f[1][2], f[1][1]
# f[3][1], f[3][2] = f[3][2], f[3][1]

a = utils.get_actor_from_primitive(v, f)

scene = window.Scene()
scene.add(a)
window.show(scene)
```

### DIFF
--- a/fury/primitive.py
+++ b/fury/primitive.py
@@ -374,9 +374,9 @@ def prim_tetrahedron():
                              [-0.5, -0.5, 0.5]])
 
     pyramid_triag = np.array([[2, 0, 1],
-                              [0, 3, 2],
+                              [0, 2, 3],
                               [0, 3, 1],
-                              [1, 2, 3]], dtype='i8')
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Thank you and good catch @rkharsan,

That is strange... I wonder when we introduce this issue and we did not see it. 

Anyway, thanks and merging

PR #575, opened 2022-04-15
Button and footer changes in docs

sreekarchigurupati:


docs/source/_static/vendor/sphinx-footer.css
```diff
@@ -0,0 +1,749 @@
+.footer-h5 {
```
skoudoro, line 1: I do not think you need such a big css footer. All this code is more maintenance. Please can you reduce it to the essential? thank you

skoudoro (commented):
Hi @sreekarchigurupati,

Thank you for this PR.

there some design issue as you can see below:
image

The download button is better but we are missing a hover effect.

`sphinx-footer.css` should be simplify

this PR is missing the pbr gif as we talked previously. Can you add it?

skoudoro:
Closed in favor of #580

Issue #562, opened 2022-03-29
drop_down_menu icon flaw in ComboBox2D 


lej0hn:
## Description
When a choice is made from the combobox, the icon changes not once but a few times, creating a flickering effect. This happens only when the combobox is added on a tabUI.

## Way to reproduce
The following example is taken from the Tab_UI tutorial.
- Run the following code
- Click on the tab to open dropdown menu
- Choose a color from the list 

```
from fury import ui, window, actor
import numpy as np
from fury.data import fetch_viz_icons


fetch_viz_icons()

colors = {
    "Violet": (0.6, 0, 0.8),
    "Indigo": (0.3, 0, 0.5),
    "Blue": (0, 0, 1),
    "Green": (0, 1, 0),
    "Yellow": (1, 1, 0),
    "Orange": (1, 0.5, 0),
    "Red": (1, 0, 0)
}

tab_ui = ui.TabUI(position=(49, 94), size=(300, 300), nb_tabs=1 , draggable=True)
color_combobox = ui.ComboBox2D(items=list(colors.keys()),
                               placeholder="Choose Text Color",
                               size=(250, 150), draggable=True)

tab_ui.add_element(0, color_combobox, (0.1, 0.3))
label = ui.TextBlock2D(
    position=(600, 300), font_size=40, color=(1, 0.5, 0),
    justification="center", vertical_justification="top",
    text="FURY rocks!!!")                               

def change_color(combobox):
    label.color = colors[combobox.selected_text]

tab_ui.tabs[0].title = "Colors"
color_combobox.on_change = change_color
sm = window.ShowManager(size=(800, 500), title="Viz Tab")
sm.scene.add(tab_ui,label)
interactive = True

if interactive:
    sm.start()
```

### Operating system and versions
[...]

PR #576, opened 2022-04-18
Resolving icon flaw in comboBox2D

lej0hn:
Fixing bug mentioned in #562


skoudoro (commented):
Hi @lej0hn,

Thank you for doing this! 

Can you address all the pep8 issues and add a test? 

Thank you

skoudoro (commented):
Hi @lej0hn,

This workaround works but I think it would have been great to see why all callbacks are called so many times instead of adding a flag.

Also, Can you add a test? Thanks

skoudoro:
superseded by #768, closing

PR #581, opened 2022-04-27
Keep original dtype for offsets in vtk format

frheault:
Allows saving VTK using offsets as int32 or int64 by keeping the initial datatype rather than casting it to int64.
Since the ArraySequence as offsets as int64 for default, this does not change the current behavior of the stateful_tractogram in DiPy.

@arokem (this PR match the one I made today in Dipy, to solve limitation with VTK for Chris' Benchmarking

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -117,6 +117,7 @@ def numpy_to_vtk_cells(data, is_coords=True):
         connectivity + offset information
 
     """
+    offset_dtype = data._offsets.dtype
     data = np.array(data, dtype=object)
     nb_cells = len(data)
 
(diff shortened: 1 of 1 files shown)


skoudoro (changes requested):
Hi @frheault,

Thank you for this. All tests are failing after your change. Can you look at it and add a test? thanks in advance.

fury/utils.py
```diff
@@ -117,6 +117,10 @@ def numpy_to_vtk_cells(data, is_coords=True):
         connectivity + offset information
 
     """
+    if isinstance(data, list) or isinstance(data, np.ndarray):
```
skoudoro, line 120: `if isinstance(data, (list, np.ndarray))` instead

skoudoro (changes requested):
Hi @frheault,

With this change, We encounter many times and at different places the following error:

```python

2022-05-25 16:41:51.973 (   1.931s) [        69E6D740]       vtkCellArray.cxx:908    ERR| vtkCellArray (0x5566eb83e3c0): Invalid array types passed to SetData: offsets=vtkUnsignedLongLongArray, connectivity=vtkUnsignedLongLongArray
```
Can you look at it and run the tests locally.

Thanks !

frheault:
@skoudoro can you activate tests on both PR, I think they should run now.
They run locally and I do not see the warnings anymore

skoudoro:
Thank you @frheault, merging. 

We plan to do a release by the end of this month (June)

PR #582, opened 2022-04-29
Deprecation of the function shaders.load

guaje:
This PR deprecates the function `shaders.load` in favor of the more explicit function `shaders.import_fury_shader`.

### DIFF
--- a/fury/shaders/base.py
+++ b/fury/shaders/base.py
@@ -2,6 +2,7 @@ import os
 
 from functools import partial
 from fury import enable_warnings
+from fury.deprecator import deprecate_with_version
 from fury.io import load_text
 from fury.lib import (VTK_OBJECT, Command, DataObject, Shader, calldata_type,
                       numpy_support)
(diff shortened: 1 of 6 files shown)


fury/shaders/tests/test_base.py
```diff
@@ -350,6 +351,7 @@ def test_load():
     npt.assert_string_equal(load(dummy_file_name), dummy_file_contents)
 
     os.remove(os.path.join(SHADERS_DIR, dummy_file_name))
+"""
```
skoudoro, line 354: remove the double quote, quit the test, and check if `assert_warns` works as expected

fury/shaders/base.py
```diff
@@ -126,6 +127,7 @@ def load_shader(shader_file):
     return load_text(shader_file)
 
 
+'''
 def load(filename):
```
skoudoro, line 131: you should remove the quote and use the decorator instead:

```python
@deprecate_with_version(since='0.8.1', until='0.9.0',
                        message='Load function has been replaced by import_fury_shader.')
def load(filename):
    ....
```

skoudoro (commented):
Thanks @guaje.

Overall look good, see below for comments

fury/shaders/base.py
guaje: Done.

fury/shaders/tests/test_base.py
guaje: By quitting the test, do you mean remove the current test and leave just the `assert_warns`?

skoudoro (commented):
LGTM, thanks @guaje , merging

PR #584, opened 2022-05-03
Changing dot actor

tvcastillod:
Hello, I made this PR in order to improve flexibility of the current dot actor. I changed `dots` actor to allow usage of multiple colors, so now it has the option to set a different color for each point created. For it to support both ndarray (N,3  or 4) and tuple (3 or 4,) I created a function called `__color_check` to manage `colors` dot's parameter. Additionally, as the actor includes an `opacity` parameter If a value is given, each dot will have the same opacity otherwise opacity is set to 1 by default, or is defined by Alpha parameter in `colors` if given.


fury/tests/test_actors.py
```diff
@@ -756,27 +760,45 @@ def test_dots(interactive=False):
 
     npt.assert_equal(scene.GetActors().GetNumberOfItems(), 1)
 
-    extent = scene.GetActors().GetLastActor().GetBounds()
-    npt.assert_equal(extent, (0.0, 1.0, 0.0, 1.0, 0.0, 0.0))
+    arr = window.snapshot(scene)
+    report = window.analyze_snapshot(arr, colors=colors)
+    npt.assert_equal(report.objects, 3)
+    npt.assert_equal(report.colors_found, [False]*3)
```
guaje, line 766: Since you are testing for colors, this test should return true for the colors in the original array. However, the range of the colors in the snapshot is between [0, 255]. Additionally, opacity acts like a percentage of the color, so you could use the opacity column to calculate the corresponding percentage of 255, like this:

```python
expected_colors = np.floor(colors[:, 3] * 255) * colors[:, :3]
report = window.analyze_snapshot(arr, colors=expected_colors)
npt.assert_equal(report.colors_found, [True, True, True])
```

fury/tests/test_actors.py
```diff
@@ -756,27 +760,45 @@ def test_dots(interactive=False):
...
+
+    if interactive:
+        window.show(scene, reset_camera=False)
 
     arr = window.snapshot(scene)
-    report = window.analyze_snapshot(arr,
-                                     colors=(0, 255, 0))
+    report = window.analyze_snapshot(arr, colors=colors)
     npt.assert_equal(report.objects, 3)
+    npt.assert_equal(report.colors_found, [False]*1)
```
guaje, line 784: Same as above, make sure these tests always return true. Use opacity to calculate the appropriate percentage of 255.

fury/tests/test_actors.py
```diff
@@ -756,27 +760,45 @@ def test_dots(interactive=False):
...
 
+    if interactive:
+        window.show(scene, reset_camera=False)
+
     arr = window.snapshot(scene)
-    report = window.analyze_snapshot(arr,
-                                     colors=(0, 0, 255))
+    report = window.analyze_snapshot(arr, colors=(1, 1, 1))
     npt.assert_equal(report.objects, 1)
+    npt.assert_equal(report.colors_found, [False]*1)
```
guaje, line 801: Make sure this returns true.

fury/actor.py
```diff
@@ -1440,42 +1446,92 @@ def dots(points, color=(1, 0, 0), opacity=1, dot_size=5):
...
 
-    aPolyVertexGrid.SetPoints(polyVertexPoints)
-    aPolyVertexMapper = DataSetMapper()
-    aPolyVertexMapper.SetInputData(aPolyVertexGrid)
-    aPolyVertexActor = Actor()
-    aPolyVertexActor.SetMapper(aPolyVertexMapper)
+    return poly_actor
+
+
+def __color_check(dots, colors=None):
```
guaje, line 1495: Can you add a test for this function? And it seems it will be better if we move it to `fury.utils `.

fury/actor.py
```diff
@@ -1440,42 +1446,92 @@ def dots(points, color=(1, 0, 0), opacity=1, dot_size=5):
...
+
+    Returns
+    -------
+    color_array : vtkDataArray
+        vtk scalar array with name 'colors'.
+    global_opacity : float
+        returns 1 if the colors array doesn't contain opacity otherwise -1.
+
+    """
+    num_dots = len(dots)
```
guaje, line 1518: Since you only need the length of the array, you could just pass that as an integer.

fury/actor.py
```diff
@@ -1440,42 +1446,92 @@ def dots(points, color=(1, 0, 0), opacity=1, dot_size=5):
...
-    aPolyVertexMapper.SetInputData(aPolyVertexGrid)
-    aPolyVertexActor = Actor()
-    aPolyVertexActor.SetMapper(aPolyVertexMapper)
+    return poly_actor
+
+
+def __color_check(dots, colors=None):
+    """
+    Returns a VTK scalar array containing colors information for each one of
+    the dots according to the policy defined by the parameter colors.
```
guaje, line 1498: This is a great generic function. Many actors could benefit from this, so, let's not limit it to the new dot actor only.

guaje (changes requested):
@tvcastillod Thanks for this PR. The new dot actor looks promising. And the `color_check` function can benefit many actors. Here are some comments and sugestions.

fury/utils.py
```diff
@@ -1369,3 +1369,46 @@ def update_surface_actor_colors(actor, colors):
...
+    global_opacity = 1
+    if colors is None:
+        # Automatic RGB colors
+        colors = np.asarray((1, 1, 1))
+        color_array = numpy_to_vtk_colors(np.tile(255 * colors, (len(actors), 1)))
+    elif type(colors) is tuple:
+        global_opacity = 1 if len(colors) == 3 else -1
+        colors = np.asarray(colors)
+        color_array = numpy_to_vtk_colors(np.tile(255 * colors, (len(actors), 1)))
+    else:
```
guaje, line 1406: To avoid future problems, let's check for ndarrays. You can do that in the following way: `elif isinstance(colors, np.ndarray):`

fury/utils.py
```diff
@@ -1369,3 +1369,46 @@ def update_surface_actor_colors(actor, colors):
...
+        colors = np.asarray(colors)
+        if colors.shape[1] == 4:
+            opacities = np.unique(colors[:, 3])
+            global_opacity = opacities[0] if len(opacities) == 1 else -1
```
guaje, line 1410: Thanks for the addition!

fury/utils.py
```diff
@@ -1369,3 +1369,46 @@ def update_surface_actor_colors(actor, colors):
...
+        If one tuple of color is used. Then all actors will have the same color.
+        If an array (N, 3 or 4) is given, where N is equal to the number of
+        points. Then every point is colored with a different RGB(A) color.
+
+    Returns
+    -------
+    color_array : vtkDataArray
+        vtk scalar array with name 'colors'.
+    global_opacity : float
+        returns 1 if the colors array doesn't contain opacity otherwise -1.
```
guaje, line 1394: With the addition of the unique check, this description should be updated.

fury/utils.py
```diff
@@ -1369,3 +1369,46 @@ def update_surface_actor_colors(actor, colors):
...
+
+    """
+    global_opacity = 1
+    if colors is None:
+        # Automatic RGB colors
+        colors = np.asarray((1, 1, 1))
+        color_array = numpy_to_vtk_colors(np.tile(255 * colors, (len(actors), 1)))
+    elif type(colors) is tuple:
+        global_opacity = 1 if len(colors) == 3 else -1
```
guaje, line 1403: If len(colors) == 4, shouldn't global_opacity be equal to colors[3]?

guaje (changes requested):
@tvcastillod thank you for applying the previously requested changes, the addition of the `test_color_check` function, and the unique check.

I've added some comments regarding the latest changes.

Additionally, you need to rebase the PR.

PR #588, opened 2022-05-23
Fixed Sphere Creation Error on viz_pbr_interactive Tutorial

saran820:
Substituted a python list with numpy array. Additionally, **use_primitive = False** was added during sphere creation to fix error.


skoudoro (commented):
Hi @saran820,

Thank you, for this PR. 

The sphere issue needs to be fixed directly on the sphere function and not at the level of the tutorial. After doing this fix in `actor.py`, some tests need to be added.

Can you explain why you need a `np.array` here ? thank you.

PR #591, opened 2022-05-27
Replaced a deprecated method.

m-agour:
Replaced deprecated method 'label' with 'vector_text'.
Also eliminated the "true division" runtime warning.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1593,11 +1593,7 @@ def sphere(centers, colors, radii=1., phi=16, theta=16,
         sphere_actor.GetProperty().SetOpacity(opacity)
         return sphere_actor
 
-    if type(radii) in [float, int]:
-        scales = np.array([1, 1, 1]) * radii
-    else:
-        scales = np.array([1, 1, 1]) * radii[:, None]
-
+    scales = radii
     directions = (1, 0, 0)
 
     if faces is None and vertices is None:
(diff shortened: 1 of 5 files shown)


fury/actor.py
```diff
@@ -1593,7 +1593,11 @@ def sphere(centers, colors, radii=1., phi=16, theta=16,
         sphere_actor.GetProperty().SetOpacity(opacity)
         return sphere_actor
 
-    scales = np.multiply(radii, [1, 1, 1])
+    if type(radii) in [float, int]:
+        scales = np.array([1, 1, 1]) * radii
+    else:
+        scales = np.array([1, 1, 1]) * radii[:, None]
```
xtanion, line 1599: This works perfectly fine @m-agour , 
You may also use `scales = np.array(radii)` as `repeat_primitive` handles the reshaping of the scales array.

xtanion (commented):
Thanks, @m-agour for this PR. 
Please take a look at my comment below. Thanks

ganimtron-10 (commented):
Hey @m-agour ,
All your changes are perfectly fine!!

There is just an issue with the sphere texture in `viz_earth_animation` which is inverted horizontally as shown below!!

earth_animation_issue
earth_animation_issue_2

fury/actor.py
m-agour: Hello @xtanion,
Thank you for reviewing my PR.
The compact solution you mentioned works just fine. Thanks for mentioning it.

m-agour:
Hello @ganimtron-10,
Thank you for pointing out this issue. I'm looking into it.

skoudoro (approved):
Thank you for this @m-agour,

Overall, looks good to me. Good catch for the reverted earth @ganimtron-10! I think, this should be fixed in a new PR otherwise the title will not correspond to the change.

So, I will go ahead and merge it. Thanks all

PR #593, opened 2022-05-27
Support empty ArraySequence in saving (for empty vtk)

frheault:
Added a simple check to avoid the np.stack if an array is empty.

The diff looks weird, but it's simply a new if to avoid the coloring part if there is no streamlines

### DIFF
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -259,7 +259,7 @@ def lines_to_vtk_polydata(lines, colors=None):
 
     """
     # Get the 3d points_array
-    if isinstance(lines, ArraySequence):
+    if not isinstance(lines, list):
         points_array = lines._data
     else:
         points_array = np.vstack(lines)
--- a/fury/utils.py
+++ b/fury/utils.py
@@ -1,3 +1,4 @@
+from nibabel.streamlines.array_sequence import ArraySequence
 import numpy as np
 from scipy.ndimage import map_coordinates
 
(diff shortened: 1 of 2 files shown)


fury/utils.py
```diff
@@ -1,3 +1,4 @@
+from nibabel.streamlines.array_sequence import ArraySequence
```
skoudoro, line 1: FURY does not depend on Nibabel. Other fields do not need `nibabel` and we want to stay the most generic as possible for FURY project.

skoudoro (commented):
Hi @frheault,

Thank you for this. We will not add `nibabel` as a dependency.  Unfortunately, you need to change the way you are doing it here.

fury/utils.py
frheault: How do I check if it's an ArraySeqence then?
I test if not a list?

skoudoro (changes requested):
Hi @frheault,

Thank you for the update. Can you look at this last error:

```python
        # Get the 3d points_array
        if not isinstance(lines, list):
>           points_array = lines._data
E           AttributeError: 'numpy.ndarray' object has no attribute '_data'
../fury/utils.py:262: AttributeError
=========================== short test summary info ============================
FAILED ../fury/tests/test_molecular.py::test_bounding_box - AttributeError: '...
========================= 1 failed, 12 passed in 4.03s =========================

```

skoudoro (commented):
Hi @frheault,

Can you do this small change and add a small test with different types of empty streamlines.
After, it should be good to go. I might have merged your other PR a bit fast.  

Thank you. 

ps: since `dipy` is optional, we will have to use the decorator `skipif` and optional package for `dipy`.

skoudoro (approved):
Thank you for this @frheault, merging

PR #597, opened 2022-06-08
Added an accurate way to get the FPS for the showManager

m-agour:
## What is wrong with the old way?
Since it basically depends on the time needed to render a single frame using the VTK' method "**GetLastRenderTimeInSeconds()**", it will be accurate with both of the two conditions below are true:
 1.  If the scene is all about rendering and no side calculations.
 2. The render method is called with an offset (delay) less than (not guaranteed) or equals the time required to render a single frame.

**For example**:
If the render method is called 32 times every second, this should give a 32 FPS as calculated by [Fraps](https://fraps.com/download.php) in yellow. This makes sense since it only depends on the render time.
```
showm.add_timer_callback(True, 1000//32 , timer_callback)
```
 image

The proposed way is to actually measure the time from rendering a frame until it renders the next frame.

## The old way vs. the new way:
For a scene with the conditions above, they are both accurate:

https://user-images.githubusercontent.com/63170874/172708989-982fff98-462e-4bf7-a1ef-781e2d42521f.mp4

### But for the other conditions, the old way is not accurate as shown below:
 
 https://user-images.githubusercontent.com/63170874/172709571-c3c6bfd8-a349-4f69-9ee5-be3e0f7ec5a7.mp4

### Code used for the last test:
```
import numpy as np  
from fury import window, actor, ui  
import itertools  
  
xyz = 1000 * np.random.rand(10, 3)  
colors = np.random.rand(10, 4)  
radii = np.random.rand(10) * 50 + 0.5  
  
scene = window.Scene()  
  
sphere_actor = actor.sphere(centers=xyz,  
  colors=colors,  
  radii=radii)  
  
scene.add(sphere_actor)  
  
showm = window.ShowManager(scene,  
  size=(900, 768), reset_camera=False,  
  order_transparent=True)  
  
showm.initialize()  
  
tb = ui.TextBlock2D(bold=True, position=(20, 720))  
  
counter = itertools.count()  
  
fps_scene = []  
fps_showm = []  
  
def timer_callback(_obj, _event):  
    global counter, fps_scene, fps_showm  
  
    cnt = next(counter)  
  
    if not (cnt % 5):  
        fps_scene.append(scene.frame_rate)  
        fps_showm.append(showm.frame_rate)  
        tb.message = f"Avg. FPS (old): {sum(fps_scene) // len(fps_scene)}" \  
                     f"\nAvg. FPS (new): {sum(fps_showm) // len(fps_showm)}"  
  fps_scene = []  
        fps_showm = []  
  
    showm.scene.azimuth(0.05 * cnt)  
    sphere_actor.GetProperty().SetOpacity(cnt / 100.)  
    showm.render()  
  
scene.add(tb)  
  
showm.add_timer_callback(True, 1000//32, timer_callback)  
showm.start()
```

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -412,7 +412,6 @@ class ShowManager(object):
 
     def start(self):
         """Start interaction."""
-        # self._last_render_time = time.perf_counter()
         try:
             self.render()
             if self.title.upper() == "FURY":
--- a/fury/window.py
+++ b/fury/window.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import gzip
+import time
 from tempfile import TemporaryDirectory as InTemporaryDirectory
 from warnings import warn
 
@@ -362,6 +363,8 @@ class ShowManager(object):
         self.interactor_style = interactor_style
         self.stereo = stereo
         self.timers = []
+        self.fps = 0
+        self._last_render_time = 0
 
         if self.reset_camera:
             self.scene.ResetCamera()
@@ -403,10 +406,13 @@ class ShowManager(object):
 
     def render(self):
         """Render only once."""
+        self.fps = 1.0 / (time.perf_counter() - self._last_render_time)
+        self._last_render_time = time.perf_counter()
         self.window.Render()
 
     def start(self):
         """Start interaction."""
+        # self._last_render_time = time.perf_counter()
         try:
             self.render()
             if self.title.upper() == "FURY":
@@ -433,6 +439,11 @@ class ShowManager(object):
         self.window.Finalize()
         del self.iren
         del self.window
+
+    @property
+    def frame_rate(self):
+        """Returns FPS."""
+        return self.fps
 
     def record_events(self):
         """Record events during the interaction.
--- a/fury/window.py
+++ b/fury/window.py
@@ -261,10 +261,10 @@ class Scene(OpenGLRenderer):
         return self.GetActiveCamera().GetDirectionOfProjection()
 
     @property
-    def frame_rate(self):
-        rtis = self.GetLastRenderTimeInSeconds()
-        fps = 1.0 / rtis
-        return fps
+    def last_render_time(self):
+        """Returns the last render time in seconds."""
+
+        return self.GetLastRenderTimeInSeconds()
 
     def fxaa_on(self):
         self.SetUseFXAA(True)
--- a/fury/window.py
+++ b/fury/window.py
@@ -406,7 +406,7 @@ class ShowManager(object):
 
     def render(self):
         """Render only once."""
-        self.fps = 1.0 / (time.perf_counter() - self._last_render_time)
+        self._fps = 1.0 / (time.perf_counter() - self._last_render_time)
         self._last_render_time = time.perf_counter()
         self.window.Render()
 
--- a/fury/window.py
+++ b/fury/window.py
@@ -363,7 +363,7 @@ class ShowManager(object):
         self.interactor_style = interactor_style
         self.stereo = stereo
         self.timers = []
-        self.fps = 0
+        self._fps = 0
         self._last_render_time = 0
 
         if self.reset_camera:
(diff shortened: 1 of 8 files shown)


ganimtron-10 (commented):
Hello @m-agour ,
Thank you for this PR.
Take a look at below comments.
Thanks!

fury/window.py
```diff
@@ -403,10 +406,13 @@ def initialize(self):
 
     def render(self):
         """Render only once."""
+        self.fps = 1.0 / (time.perf_counter() - self._last_render_time)
+        self._last_render_time = time.perf_counter()
         self.window.Render()
 
     def start(self):
         """Start interaction."""
+        # self._last_render_time = time.perf_counter()
```
ganimtron-10, line 415: It seems this comment isn't required here.

m-agour: Yes, indeed. I was thinking about uncommenting it. But it won't do anything useful. So I'm going to remove it.

skoudoro (commented):
Hi @m-agour,

Nice description, it clarifies well the problem. This PR looks good and it is almost ready to go. I just 2 questions:

- Can you add a test? e.g look/update [here](https://github.com/fury-gl/fury/blob/master/fury/ui/tests/test_elements_callback.py#L20) 
- there is a `frame_rate` property in the `Scene` class, can you check it and tell us if it is coherent with your function `frame_rate` added in `ShowManager`?

skoudoro:
Note: Your code used above can be used as a test in `test_window.py`

fury/window.py
```diff
@@ -403,6 +406,8 @@ def initialize(self):
 
     def render(self):
         """Render only once."""
+        self.fps = 1.0 / (time.perf_counter() - self._last_render_time)
```
xtanion, line 409: It should be `self._fps`?

xtanion (commented):
Hi @m-agour, Nice explanation on the old vs new frame rate counter.
However, while running the code snippet you provided for the last test, `showm.frame_rate` is always returning 0. PTAL, Thanks.

m-agour:
> Hi @m-agour,
> 
> Nice description, it clarifies well the problem. This PR looks good and it is almost ready to go. I just 2 questions:
> 
> * Can you add a test? e.g look/update [here](https://github.com/fury-gl/fury/blob/master/fury/ui/tests/test_elements_callback.py#L20)
> * there is a `frame_rate` property in the `Scene` class, can you check it and tell us if it is coherent with your function `frame_rate` added in `ShowManager`?



Hello @skoudoro,

- I saw [this](https://github.com/fury-gl/fury/blob/master/fury/ui/tests/test_elements_callback.py#L20) test and already tested the new `fram_rate` before making the PR. But the result was the same (FPS for the anti-aliased scene is the same as the non-anti-aliased one) which is strange! it might not be if FXAA is used but MSAA show reflect a decrease in FPS!
I will investigate it again and update it or add a new test today. Please don't merge just yet.

- The `frame_rate` property in the `Scene` is actually what I meant by "the old way". It basically calculates the time required to render a single frame. And does not reflect the actual FPS. But since it's already getting calculated by VTK, and it might be useful to benchmark the rendering process itself, I suggest we keep it (maybe rename it and make it return the time, not the frequency).


Thank you

fury/window.py
m-agour: How did I miss that! Thanks for pointing that out @xtanion .

fury/window.py
```diff
@@ -434,6 +440,11 @@ def start(self):
         del self.iren
         del self.window
 
+    @property
+    def frame_rate(self):
```
filipinascimento, line 444: Old definition of frame_rate based on GetLastRenderTimeInSeconds was not removed.

filipinascimento (changes requested):

fury/window.py
m-agour: Since it's being used in some tests, and it's already getting calculated anyway, we may just change its name and return what "GetLastRenderTimeInSeconds" returns. It would be good to have a benchmark for the rendering itself.

skoudoro:
What is the status of this PR? Can you look at why the tests are failing @m-agour? Thanks!

skoudoro:
Since the failures are not related to this PR, I will go ahead and merge it. 

Thank you for this work @m-agour!

PR #598, opened 2022-06-09
Timer id is returned after creating the timer.

m-agour:
There was no way to get the id of a timer to destroy it. So I altered the "add_timer_callback" method to return the timer id so that we can use it to destroy the timer.
Also updated the tutorial viz_timers.py to show how to create/destroy timers during runtime.


ganimtron-10 (commented):
Hey @m-agour ,
This PR LGTM.
Thanks!

skoudoro (approved):
LGTM! thanks @m-agour and thanks @ganimtron-10 for the review. merging

PR #508, opened 2021-10-18
[WIP] point_cloud actor and tutorial

guaje:
This PR adds an actor to visualize point cloud datasets and a tutorial explaining how to visualize different parts of a point cloud analysis pipeline.

Here are some of the figures included in the tutorial.

| Colors | Normals | Reoriented Normals |
|---|---|---|
|  |  |  |

PR #604, opened 2022-06-15
Repeating normals of primitives and applying it to sphere actor.

m-agour:
# Smooth shading sphere actor

The sphere actor generated using repeat_primitive was not shaded the same as the vtk sphere due to the lack of normals.
VTK already applied this but was not applied using `repeat_primitives`.
It doesn't need any additional calculations since the mean normal vector for each vertex is the same vector from the center to this vertex and is also the same as the vertex itself (v - origin = v)
The `repeat_primitve` function was not prepared to handle repeating normals, so this had to be implemented as well and will be useful when applying the same to any round shape like a cylinder and arrow
Some actors' code is edited as well since the `repeat_primitive` now returns 5 vars. 

### Before using normals (blue sphere):
VTK sphere

### After applying normals (blue sphere):
image


xtanion (commented):
Thanks for the PR @m-agour. Everything's working fine. I found no issues in my testing.

m-agour:
> Thanks for the PR @m-agour. Everything's working fine. I found no issues in my testing.

Thank you shivam for the testing. I really appreciate it.

ganimtron-10 (commented):
Hello @m-agour ,
Thank You for the PR, LGTM.
I will play around with it to test more.

m-agour:
Hello @guaje,
Thank you for your review. 
I will need to investigate [this](https://github.com/fury-gl/fury/pull/508/files#diff-6929c99074db37dea3bb6f30958ecca518e6e685d7a178907568e63624241e66) PR Couldn't run it now without XYZ file. I'll obtain one and try it later. 
The method I used, as you know, has to be done manually for each actor (if needed) and I think there is no algorithm to do it (I think it's doable but it depends mainly on the actor and one's preferences). for instance, In OpenGL, there is this function `glShadeModel(GL_SMOOTH)`  that basically does what I did with the sphere. But for other actors like cylinders, it would look like a capsule and the arrow would lose its details, but at the same time for complex objects, it makes them very good looking (I can't help but care about the aesthetics of shapes). 
But for now, as you said, this PR should sit there till a final decision on how normals are handled or maybe when and how to smooth shade actors.

m-agour:
> Hello @m-agour , Thank You for the PR, LGTM. I will play around with it to test more.

Hello, @ganimtron-10 Thank you for reviewing the PR and testing it.

Issue #472, opened 2021-07-26
Allow specifying the resolution of screenshots


jhlegarreta:
It would be advantageous to be able to specify the resolution of images when saving a screenshot of a scene.

**Is your feature request related to a problem? Please describe.**
The feature request is not related to a problem.

**Describe the solution you'd like**
It would be nice if `FURY` allowed to specify the resolution (i.e. in dpi) when saving a screenshot of a scene to disk (i.e. when calling `window.snapshot`).

**Describe alternatives you've considered**
Adding an additional argument to the `window.snapshot` method, and passing such an argument to `io.save_image` would be necessary. The specifics of the way this would be passed to the engines used to save images would dwell in this last method.

**Additional context**
Specifying the resolution of a scene is useful; at times we want to generate very high resolution images. Right now we are forced to do this manually with a third party tool, which is not very convenient.

Garyfallidis:
Thank you @jhlegarreta.  Which third party tool are you using for now?

jhlegarreta:
Unfortunately I was doing it manually, i.e. using GIMP, with all that implies.

Nibba2018:
Hello, may I work on this Issue?

jhlegarreta:
> Hello, may I work on this Issue?

@Nibba2018 Please go ahead. Thank you.

PR #610, opened 2022-06-24
Add DPI support for window snapshots

Nibba2018:
Users can now specify `dpi` for `window.snapshot`.
Related to #472 

To-Do:
* [X] Implement dpi with pillow
* [x] Tests

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from tempfile import TemporaryDirectory as InTemporaryDirectory
 from urllib.request import urlretrieve
 
@@ -181,7 +182,7 @@ def load_text(file):
 
 
 def save_image(arr, filename, compression_quality=75,
-               compression_type='deflation', use_pillow=True):
+               compression_type='deflation', use_pillow=True, dpi=(72, 72)):
     """Save a 2d or 3d image.
 
     Expect an image with the following shape: (H, W) or (H, W, 1) or
@@ -201,10 +202,16 @@ def save_image(arr, filename, compression_quality=75,
         select between: None, lzw, deflation (default)
     use_pillow : bool, optional
         Use imageio python library to save the files.
+    dpi : float or (float, float)
+        Dots per inch (dpi) for saved image.
+        Single values are applied as dpi for both dimensions.
 
     """
     if arr.ndim > 3:
         raise IOError("Image Dimensions should be <=3")
+
+    if isinstance(dpi, (float, int)):
+        dpi = (dpi, dpi)
 
     d_writer = {".png": PNGWriter,
                 ".bmp": BMPWriter,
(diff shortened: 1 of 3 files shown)


skoudoro:
Thank you for this @Nibba2018!  I will look into it before the end of this week

skoudoro:
Also, can you test this PR @jhlegarreta? thank you

fury/window.py
```diff
@@ -931,6 +931,9 @@ def snapshot(scene, fname=None, size=(300, 300), offscreen=True,
         Maximum number of peels for depth peeling (Default 4).
     occlusion_ratio : float
         Occlusion ration for depth peeling (Default 0 - exact image).
+    dpi : float or (float, float)
```
jhlegarreta, line 934: Can also be an `int` or a tuple of   `int` values.

fury/io.py
```diff
@@ -223,9 +230,12 @@ def save_image(arr, filename, compression_quality=75,
     if use_pillow:
         arr = np.flipud(arr)
         im = Image.fromarray(arr)
-        im.save(filename, quality=compression_quality)
+        im.save(filename, quality=compression_quality, dpi=dpi)
         return
 
+    if dpi != (0, 0):
+        warnings.warn(UserWarning('DPI value is ignored while saving images via vtk.'))
```
jhlegarreta, line 237: I saw the original PR comment, and @Nibba2018's question related to specifying the DPIs to the `VTK` writer. I know this was not going to be easy, but I think it should be eventually implemented.

As the current PR is already an improvement, I would merge it upon testing success but would keep the issue open; once the PR merge, I can add a word saying that using the DPI value for the `VTK` writer is still to be done.

jhlegarreta:
> Also, can you test this PR @jhlegarreta? thank you

Will do and will update here.

jhlegarreta:
BTW, reported failures are unrelated to this patch set:
https://github.com/fury-gl/fury/runs/7095506441?check_suite_focus=true#step:10:2654

Nibba2018:
Hello @jhlegarreta 

> I have checked that the previous implementation writes images at 72 ppp, and the default value above does also result in 72 ppp images. Modifying the file's DPI seems to happen in pillow e.g. in [here](https://github.com/python-pillow/Pillow/blob/2c58c3ee11dfc617480cdbf459bd02d989e48583/src/PIL/PngImagePlugin.py#L1353) for a PNG image. But I cannot see where pillow writes a default 72 if no value is specified.

Yes indeed. Apparently it set the dpi to `None` if you don't explicitly specify the dpi.

> Now, in either case, the default value of 0 might imply that we are relying on whatever pillow thinks is appropriate/deferring to their default value, unless I am missing something. If this changes in their future releases, then we would silently be changing that as well, which might not be desirable IMHO.

I agree with your point. Explicitly setting 72 as the default argument for `dpi` sounds appropriate as it will also have the same behaviour as before.

fury/io.py
jhlegarreta: @Nibba2018 It is some time since I have not worked with `VTK` so I probably do not have all the details with me. I did have a quick look before submitting the issue, and when reviewing your PR again, and did not find any immediate documentation. If you feel that you are willing to implement specifying DPIs for a `VTK` image, you may want to consider:
- Investigating how they accomplish this in other `VTK`-based projects (e.g. `PyVista`. `Slicer 3D` ?)
- Asking in the `VTK` discourse forum

Hope this helps.

fury/io.py
```diff
@@ -223,9 +230,12 @@ def save_image(arr, filename, compression_quality=75,
     if use_pillow:
         arr = np.flipud(arr)
         im = Image.fromarray(arr)
-        im.save(filename, quality=compression_quality)
+        im.save(filename, quality=compression_quality, dpi=dpi)
         return
 
+    if dpi != (72, 72):
```
jhlegarreta, line 236: This will trigger the warning for any value different from `(72, 72)`, e.g. `(150, 150)`, which is not the intent of the warning. It should be `if not use_pillow:` (regardless of the DPI value), right?

jhlegarreta (changes requested):

fury/io.py
Nibba2018: I saved images using vtk and it seems like it always saves them at 72 dpi. So I thought maybe I should put 72 as default for both the method but I do understand your point of view. I have updated the if else block.

fury/tests/test_io.py
```diff
@@ -153,11 +154,35 @@ def test_pillow():
...
+                    save_image(data, fname_path, use_pillow=opt1)
+            else:
+                save_image(data, fname_path, use_pillow=opt1)
             data2 = load_image(fname_path, use_pillow=opt2)
             npt.assert_array_almost_equal(data, data2)
             npt.assert_equal(data.dtype, data2.dtype)
 
+        save_image(data, fname_path, use_pillow=True)
+        img_dpi = Image.open(fname_path).info.get('dpi')
+        assert abs(72 - img_dpi[0]) < 0.01
```
jhlegarreta, line 168: Can this tolerance be defined as a constant in the test method?

jhlegarreta (commented):
Thanks for the effort @Nibba2018.

I had added an inline comment, and while addressing it I'd squash all commits into a single commit to make the commit history cleaner/easier to navigate.

fury/tests/test_io.py
jhlegarreta: > Also there's a weird issue where the code coverage drops to 50% for no evident reason.

I have observed `codecov` reporting drops in the past that did not match the changes, so I'd not be worried in this case.

Please, squash all commits into a single one so that the commit history is kept cleaner/easier to navigate. I'd say that the PR would be ready to be merged then. Thanks @Nibba2018.

Nibba2018: Hello @jhlegarreta , I have squashed the commits.

jhlegarreta: Done in d369613.

jhlegarreta (approved):
Test failures are unrelated. The PR is ready to be merged.

skoudoro (approved):
in the case of VTK, you might need to use  scene `magnification` to increase the resolution. This is what we do in `save_screenshot` and `record` function as you can see in  the docstring [here](https://github.com/fury-gl/fury/blob/33dd6f269a5b23641d306f1579466397fba63d28/fury/window.py#L751)  or [here](https://github.com/fury-gl/fury/blob/33dd6f269a5b23641d306f1579466397fba63d28/fury/window.py#L812).

But for now, this PR looks good to me. I will go ahead and merge it

jhlegarreta:
> in the case of VTK, you might need to use scene magnification to increase the resolution. This is what we do in save_screenshot and record function as you can see in the docstring [here](https://github.com/fury-gl/fury/blob/33dd6f269a5b23641d306f1579466397fba63d28/fury/window.py#L751) or [here](https://github.com/fury-gl/fury/blob/33dd6f269a5b23641d306f1579466397fba63d28/fury/window.py#L812).

I do not think the magnification will do the job; I'd also say that, contrary to the docs, increasing the magnification factor does not on improve the quality of the image: the image size will be larger, and smaller details will be larger, yes, but the quality will not increase.

skoudoro:
> I do not think the magnification will do the job; I'd also say that, contrary to the docs, increasing the magnification factor does not on improve the quality of the image: the image size will be larger, and smaller details will be larger, yes, but the quality will not increase.

ok, I see your point, so we need to ask VTK to provide a solution for that. No other alternatives.

PR #612, opened 2022-06-29
[WIP] Implemented a functional prototype of the keyframes animation API

m-agour:
This PR is only for testing and not to be merged since I'm reconstructing the API of this PR this week.

- This PR has an additional dependency only for this week and will not be needed in the future, which is `skimage`


m-agour:
No need for this PR anymore

PR #616, opened 2022-06-30
Fixing API limits reached issue in gltf fetcher

xtanion:
This PR adds a JSON file containing downloadable links of all models from [glTF samples](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0). This eliminates the need for API calls.

The json is generated using [this function](https://github.com/xtanion/fury/blob/gltf-json-gen/fury/data/fetcher.py#L330).

### DIFF
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -705,9 +705,18 @@ def list_gltf_sample_models():
         Lists the name of glTF sample from
         https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0
     """
-    models = urlopen(GLTF_DATA_URL).read()
-    models = json.loads(models)
-    model_names = [model['name'] for model in models if model['size'] == 0]
+    DATA_DIR = pjoin(dirname(__file__), 'files')
+    with open(pjoin(DATA_DIR, 'KhronosGltfSamples.json'), 'r') as f:
+        models = json.loads(f.read())
+    models = models.keys()
+    model_modes = [model.split('/')[0] for model in models]
+
+    model_names = []
+    for name in model_modes:
+        if name not in model_names:
+            model_names.append(name)
+    model_names = model_names[1:]  # removing __comments__
+
     default_models = ['BoxTextured', 'Duck', 'CesiumMilkTruck', 'CesiumMan']
 
     if not model_names:
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -338,6 +338,13 @@ async def _fetch_gltf(name, mode):
     mode: str
         Type of the glTF format.
         (e.g. glTF, glTF-Embedded, glTF-Binary, glTF-Draco)
+
+    Returns
+    -------
+    f_names : list
+        list of fetched all file names.
+    folder : str
+        Path to the fetched files.
     """
 
     if name is None:
@@ -395,6 +402,11 @@ def fetch_gltf(name=None, mode='glTF'):
         You can choose from different options
         (e.g. glTF, glTF-Embedded, glTF-Binary, glTF-Draco)
         Default: glTF, `.bin` and texture files are stored separately.
+
+    Returns
+    -------
+    filenames : tuple
+        tuple of feteched filenames (list) and folder (str) path.
     """
     filenames = asyncio.run(_fetch_gltf(name, mode))
     return filenames
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -316,8 +316,8 @@ async def _download(session, url, filename, size=None):
     if not os.path.exists(filename):
         print(f'Downloading: {filename}')
         async with session.get(url) as response:
-            block = 100
             size = response.content_length if not size else size
+            block = size
             copied = 0
             with open(filename, mode='wb') as f:
                 async for chunk in response.content.iter_chunked(block):
@@ -360,6 +360,7 @@ async def _fetch_gltf(name, mode):
             raise ValueError(
                 "Model name and mode combination doesn't exist")
 
+        path = pjoin(name, mode)
         path = pjoin('glTF', path)
         folder = pjoin(fury_home, path)
         if not os.path.exists(folder):
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -6,7 +6,7 @@ import contextlib
 import warnings
 import json
 
-from os.path import join as pjoin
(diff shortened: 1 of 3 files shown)


ganimtron-10 (commented):
Hi @xtanion ,
The fetcher works fine for the simple models like `Box`, `BoxTextured`, `CesiumTruck` but when I tried to download `Damaged Helmet` it took nearly 5-7 minutes to download some 1.56 MB file and at last gave this below error
```python
  File "Python\Python39\lib\asyncio\base_events.py", line 515, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

Similarly, I also tried to download the `BrainStem` model there too faced the same downloading issue along with the `Event loop is closed` error.

Also,
Please update the `_fetch_gltf` and `fetch_gltf` docs to also contain information regarding the return value.

Thanks!

fury/data/tests/test_fetcher.py
```diff
@@ -51,20 +53,29 @@ def tests_fetch_gltf():
...
     default_list = ['BoxTextured', 'Duck', 'CesiumMilkTruck', 'CesiumMan']
     result = [model in fetch_names for model in default_list]
     npt.assert_equal(result, [True, True, True, True])
+
+
+def test_read_viz_gltf():
+    gltf_dir = pjoin(fury_home, 'glTF')
+    filenames, path = fetch_gltf('Box', 'glTF-Binary')
+    filename = read_viz_gltf('Box', 'glTF-Binary')
+    npt.assert_equal(filename, pjoin(path, filenames[0]))
```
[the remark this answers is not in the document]
xtanion: Thanks @ganimtron-10 for the review, I get the issue, It didn't give me the error in the ubuntu system but as windows use `\` in their file system, It did. I'll try to fix it asap.
Thanks

ganimtron-10:
Hey @xtanion ,
Kudos! The downloading issue is resolved. Also, the path test is passing now but the last check regarding mode isn't passing. As I can see the tests are passing in GitHub Actions, I will check those again on my local machine.

Thanks for quick update!

xtanion:
Thanks, @ganimtron-10, Seems like `Runtime Error: Event loop is closed` is a windows only issue with events loop( as per this [StackOverflow](https://stackoverflow.com/questions/45600579/asyncio-event-loop-is-closed-when-getting-loop)). Please let me know if it works well on your machine.

m-agour:
Hello @xtanion ,
I confirm what @ganimtron-10 experienced. Got this runtime error `RuntimeError: Event loop is closed`

xtanion:
Thanks for confirming @ganimtron-10

m-agour:
Hello @xtanion ,
Now that you fixed the `Event loop is closed` error, it works fine on windows for all models, even this one `Unicode❤♻Test`. and all modes as well. Good job.

fury/data/fetcher.py
```diff
@@ -349,25 +357,30 @@ async def _fetch_gltf(name, mode):
...
+            os.makedirs(folder)
 
-            d_urls = [file['download_url'] for file in request]
-            sizes = [file['size'] for file in request]
-            f_names = [url.split('/')[-1] for url in d_urls]
-            f_paths = [pjoin(folder, name) for name in f_names]
-            zip_url = zip(d_urls, f_paths, sizes)
+        d_urls = [file['download_url'] for file in urls]
+        sizes = [file['size'] for file in urls]
+        f_names = [url.split('/')[-1] for url in d_urls]
```
skoudoro, line 379: I am surprised that this line works on `Windows`.

@ganimtron-10 and @m-agour, can you show me the result (return value) of `fetch_gltf(...)`?

fury/data/fetcher.py
```diff
@@ -701,9 +721,18 @@ def list_gltf_sample_models():
...
         https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0
     """
-    models = urlopen(GLTF_DATA_URL).read()
-    models = json.loads(models)
-    model_names = [model['name'] for model in models if model['size'] == 0]
+    DATA_DIR = pjoin(dirname(__file__), 'files')
+    with open(pjoin(DATA_DIR, 'KhronosGltfSamples.json'), 'r') as f:
+        models = json.loads(f.read())
+    models = models.keys()
+    model_modes = [model.split('/')[0] for model in models]
```
skoudoro, line 728: same comment as above:

I am surprised that this line works on `Windows`.

@ganimtron-10 and @m-agour, can you show me the result (return value) of `fetch_gltf(...)`?

fury/data/tests/test_fetcher.py
```diff
@@ -51,20 +53,29 @@ def tests_fetch_gltf():
...
+    box_gltf = pjoin(gltf_dir, 'Box')
+    for path in os.listdir(box_gltf):
+        mode = pjoin(box_gltf, path)
+        for file in os.listdir(mode):
+            os.remove(pjoin(mode, file))
+        os.rmdir(mode)
+    npt.assert_raises(ValueError, read_viz_gltf, 'Box')
+
+    filenames, path = fetch_gltf('Box')
+    out_path = read_viz_gltf('Box').split('/')
```
skoudoro, line 79: same here. you can use something like `os.sep` instead of `/`

skoudoro (commented):
Hi @xtanion,

Thank you for this update. I still need to check it in details but see below for some questions/suggestions

fury/data/fetcher.py
m-agour: @skoudoro, it returns `(['Fox.bin', 'Fox.gltf', 'Texture.png'], 'C:\\Users\\msi\\.fury\\glTF\\Fox\\glTF')`
I checked the folder. It's correct and includes the three files.

xtanion: Thanks for the review @skoudoro. `d_urls` is a list of all download URLs, as parts of the URL are joined by `/`. I'm splitting the URL and getting the last element which is the name of the file.

fury/data/tests/test_fetcher.py
xtanion: Yea, I think this needs to be changed.

fury/data/fetcher.py
ganimtron-10: Getting similar path value as given by @m-agour !!

skoudoro (approved):
Thanks all for the reviews and thanks @xtanion for this update.

Merging

PR #617, opened 2022-07-01
Added primitives count to the the Actor's polydata

m-agour:
As @filipinascimento, @guaje, and I was trying to find an approach to only apply the keyframe animation on a partial subset of the actor's primitives not all of them, we found that there is no currently implemented way to get the primitives count of a single actor.
In this PR, the primitive's count is saved inside the polydata of the actor.
This can be used to distinguish the vertices of different primitives

### DIFF
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -12,7 +12,7 @@ from fury import shaders
 from fury import actor, window, primitive as fp
 from fury.actor import grid
 from fury.decorators import skip_osx, skip_win
-from fury.utils import shallow_copy, rotate
+from fury.utils import shallow_copy, rotate, get_primitives_count_from_actor
 from fury.testing import assert_greater, assert_greater_equal
 from fury.primitive import prim_sphere
 
@@ -1492,3 +1492,12 @@ def test_marker_actor(interactive=False):
     report = window.analyze_snapshot(arr, colors=colors)
     npt.assert_equal(report.objects, 12)
 
+
+def test_actors_primitives_count():
+    centers = np.array([[1, 1, 1], [2, 2, 2]])
+    box_actor = actor.box(centers)
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
+    box_actor = actor.sphere(centers, (1, 0, 0))
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
+    box_actor = actor.sphere(centers, (1, 0, 0), use_primitive=True)
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -1560,10 +1560,10 @@ def test_actors_primitives_count():
 
     lines = np.array([[[0, 0, 0], [1, 1, 1]], [[1, 1, 1], [2, 2, 2]]])
     line_actor = actor.line(np.array(lines))
-    npt.assert_equal(get_actor_primitives_count(line_actor), len(centers))
+    npt.assert_equal(get_actor_primitives_count(line_actor), len(lines))
 
     streamtube_actor = actor.streamtube(np.array(lines))
-    npt.assert_equal(get_actor_primitives_count(streamtube_actor), len(centers))
+    npt.assert_equal(get_actor_primitives_count(streamtube_actor), len(lines))
 
     dots_actor = actor.dots(centers)
     npt.assert_equal(get_actor_primitives_count(dots_actor), len(centers))
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -1497,6 +1497,7 @@ def test_actors_primitives_count():
     centers = np.array([[1, 1, 1], [2, 2, 2]])
     directions = np.array([[1, 0, 0], [1, 0, 0]])
     colors = np.array([[1, 0, 0], [1, 0, 0]])
+    lines = np.array([[[0, 0, 0], [1, 1, 1]], [[1, 1, 1], [2, 2, 2]]])
 
     box_actor = actor.box(centers)
     npt.assert_equal(primitives_count_from_actor(box_actor), len(centers))
@@ -1571,7 +1572,6 @@ def test_actors_primitives_count():
     cylinder_actor = actor.cylinder(centers, directions, colors)
     npt.assert_equal(primitives_count_from_actor(cylinder_actor), len(centers))
 
-    lines = np.array([[[0, 0, 0], [1, 1, 1]], [[1, 1, 1], [2, 2, 2]]])
     line_actor = actor.line(np.array(lines))
     npt.assert_equal(primitives_count_from_actor(line_actor), len(lines))
 
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -1528,7 +1528,7 @@ def test_actors_primitives_count():
         [actor.cone, {**args_3, 'use_primitive': True}, cen_c],
         [actor.arrow, {**args_3, 'repeat_primitive': False}, cen_c],
         [actor.arrow, {**args_3, 'repeat_primitive': True}, cen_c],
-        [actor.dots, {'points': centers}, cen_c],
+        [actor.dot, {'points': centers}, cen_c],
         [actor.point, {'points': centers, 'colors': colors}, cen_c],
         [actor.line, {'lines': lines}, lin_c],
         [actor.streamtube, {'lines': lines}, lin_c],
--- a/fury/tests/test_actors.py
+++ b/fury/tests/test_actors.py
@@ -1499,85 +1499,43 @@ def test_actors_primitives_count():
     colors = np.array([[1, 0, 0], [1, 0, 0]])
     lines = np.array([[[0, 0, 0], [1, 1, 1]], [[1, 1, 1], [2, 2, 2]]])
 
-    box_actor = actor.box(centers)
-    npt.assert_equal(primitives_count_from_actor(box_actor), len(centers))
+    args_1 = {'centers': centers}
+    args_2 = {**args_1, 'colors': colors}
+    args_3 = {**args_2, 'directions': directions}
 
-    square_actor = actor.square(centers)
-    npt.assert_equal(primitives_count_from_actor(square_actor), len(centers))
+    cen_c = len(centers)
+    lin_c = len(lines)
 
-    sphere_actor = actor.sphere(centers, colors)
-    npt.assert_equal(primitives_count_from_actor(sphere_actor), len(centers))
-
-    prim_sphere_actor = actor.sphere(centers, colors, use_primitive=False)
-    npt.assert_equal(primitives_count_from_actor(prim_sphere_actor),
-                     len(centers))
-
-    prim_sphere_actor = actor.sphere(centers, colors, use_primitive=True)
-    npt.assert_equal(primitives_count_from_actor(prim_sphere_actor),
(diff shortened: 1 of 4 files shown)


filipinascimento (approved):
This PR is simple, but really important to anything that requires changing or interacting with individual elements of a multi entity actor.

It looks good to me. The tests are failing for other reasons.

Maybe for another PR, it would be interesting to have some auxiliary functions to directly access the memory view of positions, colors and other attributes of the independent entities.

fury/utils.py
```diff
@@ -690,7 +719,8 @@ def get_actor_from_primitive(vertices, triangles, colors=None,
         culling of polygons based on orientation of normal with respect to
         camera. If backface culling is True, polygons facing away from camera
         are not drawn. Default: True
-
+    prim_count: bool
+        primitives count inside the actor
```
filipinascimento, line 723: maybe something like: `primitives count to be associated with the actor`

filipinascimento (changes requested):

m-agour:
> This PR is simple, but really important to anything that requires changing or interacting with individual elements of a multi entity actor.
> 
> It looks good to me. The tests are failing for other reasons.
> 
> Maybe for another PR, it would be interesting to have some auxiliary functions to directly access the memory view of positions, colors and other attributes of the independent entities.

Thanks, @filipinascimento for your review. After today's meeting, I will check if the error is caused by this PR or something else.
Yes, the plan is to use what's done in this PR to do some operations on a single actor's primitives on both the GPU side (vertex shader maybe with the help of vertex-index attribute (not sure if this is a good solution)) and the CPU side (memory views as you mentioned).

ganimtron-10 (commented):
Hi @m-agour ,
Good Work. All the methods were well documented.
I have mentioned few comments. PTAL.
Thanks

fury/utils.py
```diff
@@ -690,7 +719,8 @@ def get_actor_from_primitive(vertices, triangles, colors=None,
         culling of polygons based on orientation of normal with respect to
         camera. If backface culling is True, polygons facing away from camera
         are not drawn. Default: True
-
+    prim_count: int
```
ganimtron-10, line 722: `,optional`

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,12 @@ def test_marker_actor(interactive=False):
     report = window.analyze_snapshot(arr, colors=colors)
     npt.assert_equal(report.objects, 12)
 
+
+def test_actors_primitives_count():
+    centers = np.array([[1, 1, 1], [2, 2, 2]])
+    box_actor = actor.box(centers)
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
+    box_actor = actor.sphere(centers, (1, 0, 0))
```
ganimtron-10, line 1500: Maybe `sphere_actor` or `sphere`

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,12 @@ def test_marker_actor(interactive=False):
...
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
+    box_actor = actor.sphere(centers, (1, 0, 0), use_primitive=True)
```
ganimtron-10, line 1502: Same as above
or above name with `prim_` prefix to differentiate the actors

m-agour: Thanks, @ganimtron-10 for your review! Helpful as always.

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,12 @@ def test_marker_actor(interactive=False):
     report = window.analyze_snapshot(arr, colors=colors)
     npt.assert_equal(report.objects, 12)
 
+
+def test_actors_primitives_count():
```
guaje, line 1496: First of all, thanks for taking the time to go through all those actors and adding support to the new functions. However, you are not testing if the proposed function works for all of them. Could you please improve this test by adding the rest of the actors you have modified?

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,12 @@ def test_marker_actor(interactive=False):
...
+    centers = np.array([[1, 1, 1], [2, 2, 2]])
+    box_actor = actor.box(centers)
+    npt.assert_equal(get_primitives_count_from_actor(box_actor), len(centers))
+    sphere_actor = actor.sphere(centers, (1, 0, 0))
+    npt.assert_equal(get_primitives_count_from_actor(sphere_actor), len(centers))
+    prim_sphere_actor = actor.sphere(centers, (1, 0, 0), use_primitive=True)
```
guaje, line 1502: Thanks for checking these cases (different ways to create the same actor).

fury/utils.py
```diff
@@ -715,6 +745,7 @@ def get_actor_from_primitive(vertices, triangles, colors=None,
 
     current_actor = get_actor_from_polydata(pd)
     current_actor.GetProperty().SetBackfaceCulling(backface_culling)
+    add_primitives_count_to_actor(current_actor, prim_count)
```
guaje, line 748: You might see in the `utils.py` file that we have both the `get/set_polydata_property` (e.g., [get_polydata_normals](https://github.com/fury-gl/fury/blob/1d8abba390301ad3ffdd538af244afe05889b499/fury/utils.py#L406) and [set_polydata_normals](https://github.com/fury-gl/fury/blob/1d8abba390301ad3ffdd538af244afe05889b499/fury/utils.py#L543)) and the `property_from/to_actor` (e.g., [normals_from_actor](https://github.com/fury-gl/fury/blob/1d8abba390301ad3ffdd538af244afe05889b499/fury/utils.py#L1237) and [normals_to_actor](https://github.com/fury-gl/fury/blob/1d8abba390301ad3ffdd538af244afe05889b499/fury/utils.py#L1296)). The reason for this is that the *polydata* functions are meant to be used for tasks like actor creations since at some point you'll have access to the polydata before it goes to the mapper. While the *actor* functions are for specific use cases, they generally require getting the mapper to get the polydata to get the desired attribute finally. This process is inefficient. Therefore, I would recommend adding the respective `get/set_polydata_primitives_count` and using them here since you can access the polydata [here](https://github.com/fury-gl/fury/blob/1d8abba390301ad3ffdd538af244afe05889b499/fury/utils.py#L701).

fury/utils.py
```diff
@@ -509,6 +509,35 @@ def add_polydata_numeric_field(polydata, field_name, field_data,
     return polydata
 
 
+def add_primitives_count_to_actor(actor, primitives_count):
```
guaje, line 512: Can you please add a test in `test_utils.py`?

fury/utils.py
```diff
@@ -509,6 +509,35 @@ def add_polydata_numeric_field(polydata, field_name, field_data,
...
+    actor: :class: `UI` or `vtkProp3D` actor
+    primitives_count : int
+
+    """
+    polydata = actor.GetMapper().GetInput()
+    add_polydata_numeric_field(polydata, "prim_count", primitives_count,
+                               array_type=VTK_INT)
+
+
+def get_primitives_count_from_actor(actor):
```
guaje, line 526: Can you please add a test in `test_utils.py`?

guaje (changes requested):
@m-agour This is a much-needed feature. Thanks for this PR.

I have added some comments, see below.

fury/utils.py
m-agour: Hello @guaje, thanks for your review. I made all the changes you requested I believe. 
I also noticed this PR was not applicable on some actors while doing the tests. so I had them added.
I hope I didn't miss anything.

guaje: Yeah, that has happened on several occasions. That was the main reason I asked you to test all the supported actors. In the past, we have assumed that some functions are supposed to work equally on all the actors, yet unfortunately, that's not the case since we have different methods to create them.

skoudoro:
Hi @m-agour,

What is the status of this PR?

m-agour:
> Member

I think it needs a final review after my last commits.

m-agour:
> Hi @m-agour,
> 
> What is the status of this PR?

I think it needs a final review after my last commits.

m-agour:
> Hi @m-agour,
> 
> What is the status of this PR?

Hi @skoudoro ,
I think it needs a final review after my last commits.

fury/tests/test_utils.py
```diff
@@ -814,12 +819,60 @@ def test_empty_list_to_polydata():
...
+
+def test_get_polydata_primitives_count():
+    polydata = PolyData()
+    add_polydata_numeric_field(polydata, "prim_count", 1, array_type=VTK_INT)
+
+    prim_count = get_polydata_primitives_count(polydata)
+    npt.assert_equal(prim_count, 1)
+
+
+def test_set_actor_primitives_count():
```
guaje, line 851: `test_primitives_count_to_actor`

fury/tests/test_utils.py
```diff
@@ -814,12 +819,60 @@ def test_empty_list_to_polydata():
...
+    act = actor.axes()
+    primitives_count_to_actor(act, 1)
+    polydata = act.GetMapper().GetInput()
+    prim_count = get_polydata_field(polydata, 'prim_count')[0]
+    npt.assert_equal(prim_count, 1)
+
+
+def test_get_actor_primitives_count():
```
guaje, line 859: `test_primitives_count_from_actor`

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,50 @@ def test_marker_actor(interactive=False):
...
+        [actor.point, {'points': centers, 'colors': colors}, cen_c],
+        [actor.line, {'lines': lines}, lin_c],
+        [actor.streamtube, {'lines': lines}, lin_c],
+    ]
+    for test_case in actors_test_cases:
+        act_func = test_case[0]
+        args = test_case[1]
+        primitives_count = test_case[2]
+        act = act_func(**args)
+        npt.assert_equal(primitives_count_from_actor(act), primitives_count)
```
guaje, line 1541: Awesome! Thanks for taking the effort of testing it in so many actors.

guaje (commented):
LGTM. Just some renaming for 2 functions left.

guaje:
LGTM. If there are no more comments, I'll merge this PR.

Anything else to add @filipinascimento, @skoudoro, @xtanion, @ganimtron-10?

fury/tests/test_actors.py
```diff
@@ -1492,3 +1492,50 @@ def test_marker_actor(interactive=False):
...
+        [actor.pentagonalprism, args_1, cen_c],
+        [actor.triangularprism, args_1, cen_c],
+        [actor.rhombicuboctahedron, args_1, cen_c],
+        [actor.cylinder, args_3, cen_c],
+        [actor.disk, args_3, cen_c],
+        [actor.cone, {**args_3, 'use_primitive': False}, cen_c],
+        [actor.cone, {**args_3, 'use_primitive': True}, cen_c],
+        [actor.arrow, {**args_3, 'repeat_primitive': False}, cen_c],
+        [actor.arrow, {**args_3, 'repeat_primitive': True}, cen_c],
+        [actor.dots, {'points': centers}, cen_c],
```
skoudoro, line 1531: `actor.dot` instead

skoudoro (approved):
Hi @m-agour,

it looks great!  Apart from the small comments below, +1 for merging. 

Thank you for this great review also @guaje. After tomorrow, Feel free to merge it.

fury/tests/test_actors.py
skoudoro: the plural is deprecated.

guaje:
@m-agour Good job! Merging.

PR #623, opened 2022-07-08
DrawPanel Feature: Adding Rotation of shape from Center

ganimtron-10:
Adding a `RingSlider2D` to interactively rotate any shape from the center.

Demo:


m-agour:
Also the rotation degree should be visible inside the slider not a percentage.

ganimtron-10:
Hello @m-agour ,
Thank you for such a wonderful review.

> * If the line was created with a slope, the rotation slider does not alight with the degree. Not sure if this should be left like this (maybe for compound shapes such as polyline I think).
> * When rotating a shape it might get out of the `DrawPanel` but when I move it it automatically gets bounded. maybe a solution for this is to update the position with the same position when rotation applied and see if it works.

Will address these changes in next commits.

> * When I try to delete a small line created by mistake I have a really hard time. maybe consider to make the eraser erases what it touches if you keep pressing the button

Thanks for the idea. I was thinking to find an approach to solve this but wasn't able to get anything useful.

> * I really missed "bring to top and back to bottom" feature.

Yeah! This is currently the main issue with the `DrawPanel`. I am working on it, will be fixed soon.

> * sometimes user might want to do nothing, so I tried to hide the rotation slider in order to take a screenshot but I can't get rid of it unless I go to drawing mode and click and have a shape drawn.

This is now fixed with the new commit. Now if you press anywhere else on the screen it will disappear.

ganimtron-10:
Hello @skoudoro @nasimanousheh @Garyfallidis ,
This PR is ready for review.
Thanks!

ganimtron-10:
Hello @skoudoro @nasimanousheh @Garyfallidis ,
I have updated this PR and made sure all things work as intended. PTAL.

python_r1bFQu08Aw


Thanks!

nasimanousheh:
Hi @ganimtron-10,
Thank you for updating the PR. The "rotation_slider_ actor" should always be front of all other shapes when you call it. Here the slider is hidden behind the square.
rotate-slider

ganimtron-10:
Hello @nasimanousheh ,
I have updated this PR as per the changes discussed in the meeting.
PTAL. Thanks!

fury/ui/tests/test_elements.py
```diff
@@ -792,9 +792,9 @@ def test_ui_combobox_2d(interactive=False):
 
 
 def test_ui_draw_shape():
```
skoudoro, line 794: this test is a bit too light for me.  You should use `snapshot` to save the scene and check if we have the expected color and the expected number of objects. 

You can look at this example to see how to do it: https://github.com/fury-gl/fury/blob/master/fury/tests/test_window.py#L26-L30

skoudoro (commented):
Hi @ganimtron-10, this PR is almost ready to go. Please, look at my comment below and update the test. Thanks !

fury/ui/elements.py
```diff
@@ -3147,16 +3213,28 @@ def rotate(self, angle):
...
+    def cal_bounding_box(self, position=None):
+        """Calculate the min, max position and the size of the bounding box.
 
         Parameters
         ----------
         position : (float, float)
             (x, y) in pixels.
         """
+        if position is None:
+            position = self.position
```
skoudoro, line 3237: `position = position or self.position` instead

fury/ui/elements.py
```diff
@@ -3172,30 +3250,30 @@ def cal_bounding_box(self, position):
...
+        new_center: ndarray(int)
+            New center for the shape.
         """
-        self.cal_bounding_box(position)
-        new_position = np.clip(self._bounding_box_min, [0, 0],
-                               self.drawpanel.size - self._bounding_box_size)
-        new_position = new_position + self._bounding_box_offset
-        return new_position.astype(int)
+        if center is None:
+            center = self.center
```
skoudoro, line 3273: same as above: `center = center or self.center` instead of the `if`

ganimtron-10: If I use this, it throws the below error
`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`.

Should I use something like `center = self.center if center is None else center` ??

skoudoro: yes

skoudoro:
Hi @ganimtron-10,

Do you have an update on this?

ganimtron-10:
Hello @skoudoro ,

This PR just lacks the improved test for `DrawShape`.
Other all changes are done.

Garyfallidis:
@ganimtron-10 some tests are failing due to ``test_ui_draw_shape``.

skoudoro:
Hi @ganimtron-10,

Tests failing on windows are expected and not due to your PR.

Tests failing on macOS are linked to your PR. Can you increase the snapshot window size? it might be a resolution problem between the different OS.

ganimtron-10:
Hello @skoudoro ,
I think there are some issues with the colors, maybe my calculated color values don't match the current color due to which the tests are failing.

skoudoro:
Thank you for the update @ganimtron-10. We still have the issue with macOS. I really do not know why, I can not generate the issue.

For now, let's just skip this OS, we will come back on it later. You can look here to see an example: https://github.com/fury-gl/fury/blob/master/fury/tests/test_actors.py#L230

ganimtron-10:
Hello @skoudoro ,
I skipped the test for macOS. PTAL.
Thanks!

skoudoro:
Thanks @ganimtron-10, merging! let's focus on the next step!

PR #626, opened 2022-07-13
Keyframe animation with camera support

m-agour:
This is the no-shader version of the animation system that consists of:

1. The `Timeline`
2. `PlaybackPanel`
3. Interpolators

I also included some tutorials. They are located at `tutorials/01_introductory`  and their file name starts with `viz_keyframe_animation`.

Docs are not complete. Will write them and the tests along this week.

Below is a record of tutorial `viz_keyframe_animation_camera.py`

- Rotation is not supported yet. Will be implemented this weak along with `slerp`.

https://user-images.githubusercontent.com/63170874/178742636-badcf8aa-bb9d-49af-8992-ac5640bd2894.mp4


fury/ui/elements.py
```diff
@@ -3461,3 +3462,139 @@ def left_button_dragged(self,  i_ren, _obj, element):
...
+        """Set max progress slider time value.
+
+        Parameters
+        ----------
+        value: float
+            Max time for the progress slider.
+        """
+        self._progress_bar.max_value = value
+
+    def get_value(self):
```
skoudoro, line 3558: too generic. also use `property`.    e.g `get_current_time` ?  `current_time`? `get_progress_value`?

fury/ui/elements.py
```diff
@@ -3461,3 +3462,139 @@ def left_button_dragged(self,  i_ren, _obj, element):
...
+        """Get current time of the progress slider.
+
+        Returns
+        -------
+        float
+            Progress slider current value.
+        """
+        return self._progress_bar.value
+
+    def get_max_value(self):
```
skoudoro, line 3568: same as above, too generic, be more specific

fury/animation.py
```diff
@@ -0,0 +1,744 @@
+import math
+import time
+import numpy as np
+from scipy import interpolate
+from fury import utils
+from fury.actor import Container
+from fury.colormap import rgb2hsv, hsv2rgb, rgb2lab, lab2rgb, xyz2rgb, rgb2xyz
+from fury.ui.elements import PlaybackPanel
+from vtkmodules.vtkRenderingCore import vtkActor
```
skoudoro, line 9: `from fury.lib import Actor` instead

fury/animation.py
```diff
@@ -0,0 +1,744 @@
...
+    def _lerp(v1, v2, t1, t2, t):
+        if t1 == t2:
+            return v1
+        v = v2 - v1
+        dt = (t - t1) / (t2 - t1)
+        return dt * v + v1
+
+    @staticmethod
+    def _get_time_delta(t, t1, t2):
+        return 0 if t <= t1 else 1 if t >= t2 else (t - t1) / (t2 - t1)
```
skoudoro, line 68: `staticmethod` and `protected`? strange. can you explain?

fury/animation.py
```diff
@@ -0,0 +1,744 @@
...
+import numpy as np
+from scipy import interpolate
+from fury import utils
+from fury.actor import Container
+from fury.colormap import rgb2hsv, hsv2rgb, rgb2lab, lab2rgb, xyz2rgb, rgb2xyz
+from fury.ui.elements import PlaybackPanel
+from vtkmodules.vtkRenderingCore import vtkActor
+
+
+class Interpolator(object):
```
skoudoro, line 12: no need of `object` since python 3.

fury/animation.py
```diff
@@ -0,0 +1,744 @@
...
+
+        self.tck = interpolate.splprep(points.T, k=self.degree, full_output=1,
+                                       s=self.smoothness)[0][0]
+        self.linear_lengths = []
+        for x, y in zip(points, points[1:]):
+            self.linear_lengths.append(
+                math.sqrt((x[1] - y[1]) * (x[1] - y[1]) +
+                          (x[0] - y[0]) * (x[0] - y[0])))
+
+    def get_points(self):
```
skoudoro, line 146: Why this function is only for this interpolator?

fury/animation.py
```diff
@@ -0,0 +1,744 @@
...
+
+class LABInterpolator(ColorInterpolator):
+    """LAB interpolator for color keyframes """
+
+    def __init__(self, keyframes):
+        super().__init__(keyframes, rgb2lab, lab2rgb)
+        self.id = 5
+
+
+class Timeline(Container):
```
skoudoro, line 253: To be more pythonic, you need to use more the `property` decorator.

skoudoro (commented):
Hi @m-agour,

Thank you for this PR.

See below for some basic comments. 

Overall, this is missing unit tests.  Also, I wonder if the interpolator should be function-based instead of class-based. All the functions can be independent and I do not see the value of all this inheritance apart from complexifying the codebase.  I will be happy to discuss it.

Concerning your use of a dictionary, I recommend using more `dict.get` instead of `dict[]` to avoid strange errors when a key is missing.

The last point is to increase your use of `property`decorator to be more pythonic. 

I would recommend focusing on this PR in the next few days. It will be great if it can follow the standard and get merged. After that, we could go to more advanced features. 

Thank you.

m-agour:
> Hi @m-agour,
> 
> Thank you for this PR.
> 
> See below for some basic comments.
> 
> Overall, this is missing unit tests (as you mentioned). Also, I wonder if the interpolator should be function-based instead of class-based. All the functions can be independent and I do not see the value of all this inheritance apart from complexifying the codebase. I will be happy to discuss it.
> 
> Concerning your use of a dictionary, I recommend using more `dict.get` instead of `dict[]` to avoid strange errors when a key is missing.
> 
> The last point is to increase your use of `property`decorator to be more pythonic.
> 
> I would recommend focusing on this PR in the next few days. It will be great if it can follow the standard and get merged. After that, we could go to more advanced features.
> 
> Thank you.

Hi @skoudoro, thank you so much for your review.
Making the interpolators function-based is doable and the data (color keyframes in targeted space and spline interpolator) can be stored in the `Timeline` itself, methods also. I will make an example of this for the spline interpolator then we may discuss which way to proceed.

fury/animation.py
m-agour: I needed the points to calculate the linear lengths only for spline interpolator.
Yeah, it was being used once, and it was simple, so I just added the implementation of it to the setup method directly.

skoudoro:
Hi @m-agour,

On your tutorial, Can you follow the syntax of all the other tutorials? [here](https://github.com/fury-gl/fury/blob/master/docs/examples/viz_bundles.py) is an example. 

Thanks

skoudoro:
Also, Do not forget to add the associated test for the new functions in `colormap.py`.
it would be great if could reshape the module.

animation
  | - __init__.py
  | - interpolator.py
  | - timeline.py

m-agour:
> Hi @m-agour,
> 
> On your tutorial, Can you follow the syntax of all the other tutorials? [here](https://github.com/fury-gl/fury/blob/master/docs/examples/viz_bundles.py) is an example.
> 
> Thanks



> Also, Do not forget to add the associated test for the new functions in `colormap.py`. it would be great if could reshape the module.
> 
> animation | - **init**.py | - interpolator.py | - timeline.py

Hi @skoudoro ,
Will do that in a few moments.

ganimtron-10:
Hi @m-agour ,
I tried playing around with the animations and it looks cool.
I don't know whether this is intended or not but when we complete our animation in the `play_once` mode and again click the `play` button it does nothing. I think it should restart the animation again.

I also felt it would be easier to directly enter the speed (for bigger values) instead of just pressing the plus icon many times. Maybe PR #499 would help in this. I will try to complete it ASAP.

Thanks!

m-agour:
Hello @ganimtron-10,
Thank you so much for reviewing this PR
> Hi @m-agour , I tried playing around with the animations and it looks cool. I don't know whether this is intended or not but when we complete our animation in the `play_once` mode and again click the `play` button it does nothing. I think it should restart the animation again.

I was just thinking about this few hours ago. I think the same as you it should restart.
> 
> I also felt it would be easier to directly enter the `speed` (for bigger values) instead of just pressing the plus icon many times. Maybe PR #499 would help in this. I will try to complete it ASAP.
> 
Yeah it would help a lot. Thanks. But don't let it distract you from your important work! Maybe I will make it change in a logarithmic way for now.

fury/animation/interpolator.py
```diff
@@ -0,0 +1,348 @@
+import numpy as np
+from scipy import interpolate
+from scipy.spatial import transform
+from fury.colormap import rgb2hsv, hsv2rgb, rgb2lab, lab2rgb, xyz2rgb, rgb2xyz
+
+
+class Interpolator:
```
filipinascimento, line 7: I've talked to @skoudoro about switching from Interpolator classes to using closures, but now looking at the code there are extra helper methods that need access to the cached data. So let's discuss this and make a decision in the next meting.

fury/animation/interpolator.py
```diff
@@ -0,0 +1,348 @@
...
+    def __init__(self, keyframes):
+        super(Interpolator, self).__init__()
+        self.keyframes = keyframes
+        self.timestamps = []
+        self.min_timestamp = 0
+        self.final_timestamp = 0
+        self._unity_kf = True
+        self.setup()
+        self.id = -1
```
filipinascimento, line 16: What the IDs are used for?

docs/tutorials/01_introductory/viz_keyframe_animation_colors.py
```diff
@@ -0,0 +1,99 @@
+"""
+==================
+Keyframe animation
+==================
+
+Color animation explained
```
filipinascimento, line 6: This may need an non interactive mode to be properly used in docs.

docs/tutorials/01_introductory/viz_keyframe_animation_minimal.py
```diff
@@ -0,0 +1,63 @@
+"""
+=====================
+Keyframe animation
+=====================
+
+Minimal tutorial of making keyframe-based animation in FURY.
```
filipinascimento, line 6: In this example, the arrow seems to disappear in my tests. This may be due to clipping boundaries problem.

docs/tutorials/01_introductory/viz_keyframe_animation_spline.py
```diff
@@ -0,0 +1,74 @@
+"""
+=====================
+Keyframe animation
+=====================
+
+Tutorial on making keyframe-based animation in FURY using Spline interpolators.
```
filipinascimento, line 6: See previous comment, may need non-interactive mode for docs building.

filipinascimento (changes requested):

fury/animation/interpolator.py
m-agour: It was for shaders to send it as a uniform. should I remove it from the current version?

docs/tutorials/01_introductory/viz_keyframe_animation_minimal.py
m-agour: Yes, sometimes I run into this issue mostly when I interact with the show manager. and it's because of the clipping boundaries. Resetting the clipping plans with each update slows down the animation. I will try to solve it before our meeting.

skoudoro:
superseded by #647. Closing

PR #600, opened 2022-06-11
Adding support for importing simple glTF files

xtanion:
This PR will allow us to load gltf files and get actors or `polydata` from it.

It currently supports:
- [x] multiple Nodes
- [x] multiple Meshes
- [x] Normals 
- [x] Cameras
- [x] Textures (baseColor)

Demo: 
image

 

WIP: glTF exporter, tests

PR #630, opened 2022-07-17
Adding function to export scenes as glTF

xtanion:
This PR adds the ability to convert a scene as gltf file and can be loaded back using glTF loader (added in #600 ).


m-agour (commented):
Hello @xtanion ,
Great job. It's working well. 
I had some issues trying to export actors with no indices such as dot and line actors. It's causing the following error: `ValueError: zero-size array to reduction operation maximum which has no identity` which is triggered by this section of code:
```
  if indices is not None:
      indices = indices.reshape((-1, ))
      amax = [np.max(indices)]
      amin = [np.min(indices)]
```
Also, there are some typos in the `gltf.py` file. It's not related to this PR:
-  extra `the`  in https://github.com/fury-gl/fury/blob/5f43d41daaf159ebbc5fad97bb3b3ed10f702312/fury/gltf.py#L91
- Typo `using accessors`  https://github.com/fury-gl/fury/blob/5f43d41daaf159ebbc5fad97bb3b3ed10f702312/fury/gltf.py#L195

fury/gltf.py
```diff
@@ -398,3 +399,404 @@ def load_camera(self, camera_id, transform_mat):
...
+def write_node(gltf, mesh=None, camera=None):
+    """Create node
+
+    Parameters
+    ----------
+    gltf : GLTF2
+        Pygltflib GLTF2 object
+    mesh : int, optional
+        Mesh index
+    camera : int, optional
```
ganimtron-10, line 608: I think `camera_index` would be a more explanatory name

fury/gltf.py
```diff
@@ -398,3 +399,404 @@ def load_camera(self, camera_id, transform_mat):
...
+        pers = gltflib.Perspective()
+        pers.aspectRatio = aspect_ratio
+        pers.znear, pers.zfar = clip_range
+        pers.yfov = angle * np.pi/180
+        cam.type = "perspective"
+        cam.perspective = pers
+    gltf.cameras.append(cam)
+
+
+def get_prim(verts, indices, cols, tcoords, normals, mat):
```
ganimtron-10, line 664: Maybe you should try using full terms for eg. `vertex`, `color`, `material`, or something similar depending on the value they have. I am also a bit confused because you have written `verts` (plural) but as per the docs they store `int` (single value).

fury/gltf.py
```diff
@@ -398,3 +399,404 @@ def load_camera(self, camera_id, transform_mat):
...
+    attr.TEXCOORD_0 = tcoords
+    attr.COLOR_0 = cols
+    prim.attributes = attr
+    prim.indices = indices
+    if mat is not None:
+        prim.material = mat
+    return prim
+
+
+def write_material(gltf, bct: int, uri: str):
```
ganimtron-10, line 700: Same as above

fury/gltf.py
```diff
@@ -398,3 +399,404 @@ def load_camera(self, camera_id, transform_mat):
...
+    pbr.metallicFactor = 0.0
+    material.pbrMetallicRoughness = pbr
+    texture.source = bct
+    image.uri = uri
+    gltf.materials.append(material)
+    gltf.textures.append(texture)
+    gltf.images.append(image)
+
+
+def write_accessor(gltf, bv, bo, ct, cnt, atype, max=None, min=None):
```
ganimtron-10, line 728: Same as above

fury/gltf.py
```diff
@@ -398,3 +399,404 @@ def load_camera(self, camera_id, transform_mat):
...
+    accessor.componentType = ct
+    accessor.count = cnt
+    accessor.type = atype
+    if (max is not None) and (min is not None):
+        accessor.max = max
+        accessor.min = min
+    gltf.accessors.append(accessor)
+
+
+def write_bufferview(gltf, buffer, bo, bl, bs=None):
```
ganimtron-10, line 762: Same as above

ganimtron-10 (commented):
Hi @xtanion ,
I had a few suggestions regarding the docs which I have mentioned below.
Also, we should discuss the name of the parameters as some of them don't clearly explain what they contain.

I haven't reviewed and tested the code yet, I will do it as soon as possible and let you know if I found anything.
Thanks!

fury/gltf.py
xtanion: writing the entire parameter name for these functions can go very long, also these parameters are explained in the docstring too.

xtanion:
Thanks for the review @ganimtron-10 and @m-agour. 

Also, thanks @m-agour for finding the method to differentiate between lines and points, using 
```py
Polydata.GetNumberOfPolys()
Polydata.GetNumberOfLines()
Polydata.GetNumberOfPolys()
``` 
is easier than comparing the ratio of `Number Of Points` vs. `Number Of Cells` of the polydata.

I figured out the [`mode`](https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#_mesh_primitive_mode) attribute in primitives decides if the rendered mesh should be a line or point or a triangle, there aren't any sample model that uses POINTS or LINES mode tho, Also the current glTF importer doesn't support rendering lines or points. I'll create a new PR with lines and points support.

skoudoro (approved):
Thank you for this work @xtanion, merging

PR #631, opened 2022-07-18
billboards using geometry shader

m-agour:
This is an example on how a geometry shader can be used along with some minimal calculations to generate billboards using the `dot` actor.

https://user-images.githubusercontent.com/63170874/179493243-55b28d24-2c94-485d-af7e-ccb296733f34.mp4


guaje (changes requested):
@m-agour, thanks for exploring this.

This feature was proposed as part of an entire [GSoC project](https://github.com/fury-gl/fury/wiki/Google-Summer-of-Code-2022-(GSOC2022)#project-3-tessellated-like-actors). Since this feature requires many things taken into account, we'll use this PR to ensure the bare minimum compared to the current state of the billboard actor, so we don't affect the tasks of your main GSoC project.

I'll add @tvcastillod to this conversation since this PR is related to her project.

filipinascimento (changes requested):

fury/actor.py
```diff
@@ -2383,6 +2383,17 @@ def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
     return bb_actor
 
 
+def billboard_gs(centers, colors, scales=(1, 1, 1)):
```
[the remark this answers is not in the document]
guaje: @filipinascimento I think it would be better to have both functionalities. We have seen how some things break when there's a major release of VTK, especially things related to shaders. If that happens in the future, we'll just change the default parameter to the original functionality while we fix the shaders.

Regarding the properties, the purpose of changing the dot actor in #584 was precisely this, using a new dot actor as the basis for a GS version of the billboard. In that case, the colors are handled by the polydata in the dot actor.

m-agour:
Hello @guaje @filipinascimento ,
I updated this PR and changed the tutorial to display impostor sphere actors on billboards as shown below

https://user-images.githubusercontent.com/63170874/182064895-27fdd00a-6372-4caa-aff6-3a4bad64e407.mp4

I will make an example of animating (scaling and changing the color of spheres) before today's meeting. Also will put it under the pressure of thousands of spheres at once.

fury/actor.py
```diff
@@ -2350,6 +2352,23 @@ def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
...
+        bb_actor = dot(centers, colors)
+        replace_shader_in_actor(bb_actor, 'geometry',
+                                import_fury_shader('gs_billboard.geom'))
+        shader_to_actor(bb_actor, 'vertex',
+                        impl_code=import_fury_shader('gs_billboard_impl.vert'),
+                        decl_code=import_fury_shader('gs_billboard_dec.vert'),
+                        block='prim_id')
+
+        shader_to_actor(bb_actor, 'fragment',
+                        decl_code=import_fury_shader('gs_billboard_dec.frag'),
```
guaje, line 2365: Add support to the custom shaders in the params `fs_dec` and `fs_impl`. As implemented [here](https://github.com/fury-gl/fury/blob/2b2caf849efd2fa894d33be661a8dd3d6679c6b2/fury/actor.py#L2371) and [here](https://github.com/fury-gl/fury/blob/2b2caf849efd2fa894d33be661a8dd3d6679c6b2/fury/actor.py#L2373).

fury/actor.py
```diff
@@ -2350,6 +2352,23 @@ def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
...
+                        block='prim_id')
+
+        attribute_to_actor(bb_actor, scales.flatten(), 'scale')
+        bb_actor.GetProperty().BackfaceCullingOff()
+        bb_actor.GetMapper().SetVBOShiftScaleMethod(False)
```
guaje, line 2370: These functions are used in both billboard creation approaches.

fury/actor.py
```diff
@@ -2317,8 +2317,8 @@ def have_2_dimensions(arr):
     return spq_actor
 
 
-def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
-              gs_prog=None, fs_dec=None, fs_impl=None):
+def billboard(centers, colors=(0, 1, 0), scales=1, using_gs=False, vs_dec=None,
```
guaje, line 2320: Maybe `use_gs` instead of `using_gs`.

fury/actor.py
```diff
@@ -2317,8 +2317,8 @@ def have_2_dimensions(arr):
...
+              vs_impl=None, gs_prog=None, fs_dec=None, fs_impl=None):
```
guaje, line 2321: Given the nature of the new functionality, we should find a way to reuse the `gs_prog` param. Something like:

```python
def billboard(..., gs_prog='gs_generated', ...):
    if gs_prog.lower() == 'gs_generated':
        # Your super awesome GS Billboard implementation
    elif gs_prog is None:
        # Use the current billboard functionality
    else:
        # Use the user's custom GS program
```

This way, we can dispose of the `use_gs/using_gs` parameter.

guaje (changes requested):
@m-agour these changes look much better. See my suggestions on better API integration. We still need to add a proper test and make sure we're scaling the fragments to keep the compatibility with the current implementation.

skoudoro:
Can you rebase and fix the conflicts of this PR? 

Thank you

PR #632, opened 2022-07-18
Bugfix: Visibility issues with ListBox2D

### LINKED ISSUE #418: ListBox2D has resizing issues when added into TabUI
## Description
When we add `ListBox2D` in the `TabUI` its creates a unnecessary scrollbar bigger than itself and `ListItems` too have big backgrounds, which on clicking resizes properly.

### My Code
```python
from fury import ui,window

sm = window.ShowManager(title='Resize Tuts',size=(1366,738))
# Individual ListBox
l = ui.ListBox2D(values=['Red1','Blue1','Green1'])
sm.scene.add(l)

# ListBox added to TabUI
l1 = ui.ListBox2D(values=['Red1','Blue1','Green1'])
l2 = ui.ListBox2D(values=['Violet','Indigo','Blue','Yellow','Green','Orange','Red'],position=(10,10))

t = ui.TabUI(size=(500,500),nb_tabs=3,draggable=True,position=(100,100))
t.add_element(0,l1,(0.1,0.1))
t.add_element(1,l2,(10,10))

sm.scene.add(t)

sm.start()
```

### ScreenShots
Issue-Img1
Issue-Img2

## System Info
`
{'fury_version': '0.7.0.post45+g31efc5a', 'pkg_path': '/home/ganimtron10/Desktop/VSCode/Fury-Workspace/fury/fury', 'commit_hash': '31efc5a8a4b18dab43093723d3259b30881ab6aa', 'sys_version': '3.8.5 (default, Jan 27 2021, 15:41:15) \n[GCC 9.3.0]', 'sys_executable': '/home/ganimtron10/Desktop/VSCode/Fury-Workspace/fury-env/bin/python', 'sys_platform': 'linux', 'numpy_version': '1.20.2', 'scipy_version': '1.6.2', 'vtk_version': '9.0.1'}
`

Nibba2018:
Setting the visibility of a ListBox2D from `False` to `True` after creation causes the scrollbar and unused ListBox2D items to render incorrectly.

This also fixes #418

### DIFF
--- a/fury/ui/elements.py
+++ b/fury/ui/elements.py
@@ -2405,6 +2405,7 @@ class ListBox2D(UI):
                                             scroll_bar_height))
         if len(self.values) <= self.nb_slots:
             self.scroll_bar.set_visibility(False)
+            self.scroll_bar.height = 0
         self.panel.add_element(
             self.scroll_bar, size - self.scroll_bar.size - self.margin)
 
--- a/fury/ui/elements.py
+++ b/fury/ui/elements.py
@@ -2622,6 +2622,8 @@ class ListBox2D(UI):
             if slot.textblock.scene is not None:
                 clip_overflow(slot.textblock, self.slot_width)
             slot.set_visibility(True)
+            if slot.size[1] != self.slot_height:
+                slot.resize((self.slot_width, self.slot_height))
             if slot.element in self.selected:
                 slot.select()
             else:
@@ -2653,6 +2655,7 @@ class ListBox2D(UI):
 
         if len(self.values) <= self.nb_slots:
             self.scroll_bar.set_visibility(False)
+            self.scroll_bar.height = 0
 
     def clear_selection(self):
         del self.selected[:]
--- a/fury/ui/elements.py
+++ b/fury/ui/elements.py
@@ -2631,6 +2631,7 @@ class ListBox2D(UI):
         for slot in self.slots[len(values_to_show):]:
             slot.element = None
             slot.set_visibility(False)
+            slot.resize((self.slot_width, 0))
             slot.deselect()
 
     def update_scrollbar(self):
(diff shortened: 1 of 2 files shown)


fury/ui/elements.py
```diff
@@ -2630,6 +2631,7 @@ def update(self):
         for slot in self.slots[len(values_to_show):]:
             slot.element = None
             slot.set_visibility(False)
+            slot.resize((self.slot_width, 0))
```
ganimtron-10, line 2634: Why do we need to `resize` this with `(self.slot_width, 0)`??
As there is no `height` it wouldn't be visible anyways??

ganimtron-10 (commented):
Hello @Nibba2018 ,
Thank you for the PR.
I tested the program again and now the issue is resolved.
I have mentioned a comment below. PTAL.
Thanks!

fury/ui/elements.py
Nibba2018: Hello @ganimtron-10 , Yes, this is the part where the slot height is set to zero and the width is kept unchanged.

skoudoro (commented):
LGTM, thanks @Nibba2018. 

Can you also update the test to handle this case?

Nibba2018:
Hello @skoudoro Thank you for the review.
Regarding the test, do I add a new assert statement within `test_ui_listbox_2d` or should I create a separate test? The said test is meant to be skipped it seems.

Also I have added two additional lines of code to make ensure expected behaviour when additional values added to the listbox.

skoudoro:
> Regarding the test, do I add a new assert statement within test_ui_listbox_2d or should I create a separate test? The said test is meant to be skipped it seems.

a new assert should be enough but if you want to do more, feel free to create a new test.

Thanks !

skoudoro:
Thank you @Nibba2018. merging

PR #634, opened 2022-07-19
Prevented calling `on_change` when slider value is set without user intervention

m-agour:
Usually, `on_change` hook inside the `LineSlider2D` is used to execute something when the user changes the slider.
If this something involves setting the slider's value, an infinite recursion situation occurs.
@ganimtron-10 mentioned some other sliders behave the same, but the tests of PR #626 are failing because of this issue so I had to make this PR to rebase my PR on it until @skoudoro decides whether we should keep the functionality of the `on_change` hook as it is and make another hook that gets triggered on user interaction with the slider only keep this PR as it is.


skoudoro:
Your last commit looks good, do not forget to add the tests

skoudoro (commented):
`test_ui_line_slider_2d_horizontal_bottom` is failing. Can you give a look? thank you

m-agour:
Hi @skoudoro,
I tried the tests on ubuntu using Python 3.7 and installed default deps and the tests succeeded.
Maybe @xtanion can confirm.

m-agour:
> Hi @skoudoro, I tried the tests on ubuntu using Python 3.7 and installed default deps and the tests succeeded. Maybe @xtanion can confirm.


> Hi @skoudoro, I tried the tests on ubuntu using Python 3.7 and installed default deps and the tests succeeded. Maybe @xtanion can confirm.
So I didn't let it finish but testing all the ui tests gives a segmentation fault on test `test_frame_rate_and_anti_aliasing`

skoudoro:
Thank you for checking @m-agour. I do not know why we have so many issues in Ubuntu. I need to find a machine and spend some time on it. 
In any case, Thanks, merging

m-agour:
You are welcome @skoudoro.

PR #647, opened 2022-07-30
Keyframe animations and interpolators

m-agour:
Hello @filipinascimento, @skoudoro,
As you suggested, in this PR, I restructured the interpolator classes to functions using closures.
Could you review it and let me know if I should document it and merge it into the main PR of animation?


skoudoro (commented):
That's Great! thank you for proposing another version. great job!

Before reviewing, I would like to know your point of view on these 2 versions. 
Do you think there is more limitation?
Drawbacks, Advantages between both versions? 

I would also like to hear the opinion of @xtanion and @ganimtron-10.

we can talk about it on Monday and then push forward with whatever we choose

m-agour:
> That's Great! thank you for proposing another version. great job!
> 
> Before reviewing, I would like to know your point of view on these 2 versions. Do you think there is more limitation? Drawbacks, Advantages between both versions?

I think it's better this way (only if it's acceptable to have `fury.animation.helper` which has the functions I'm using frequently) the  limitations I faced are:
1- Now I can't differentiate between normal time-based functions or equations and interpolators since now they are both functions, but this is solved using an additional argument. 
2- I can't know which interpolator I am using, since I can't get info from the function, this can be solved by returning an `id` alongside the interpolator closure as a tuple or dict. but for now, It's fine like this.

That being said, I think It's clearer this way and we may discuss it in more detail next meeting.

fury/animation/tests/test_helpers.py
```diff
@@ -0,0 +1,55 @@
+import numpy.testing as npt
+import fury.animation.helpers as helpers
+from fury.testing import *
+
+
+def test_helper_functions():
```
skoudoro, line 6: split the tests into smaller tests as we discuss recently

fury/animation/tests/test_interpolators.py
```diff
@@ -0,0 +1,210 @@
+from itertools import combinations
+import numpy.testing as npt
+from fury.animation.interpolator import linear_interpolator, \
+    step_interpolator, cubic_spline_interpolator, cubic_bezier_interpolator, \
+    spline_interpolator, hsv_color_interpolator, lab_color_interpolator, \
+    xyz_color_interpolator, slerp
+from fury.testing import *
```
skoudoro, line 7: same comment as above, never

fury/animation/tests/test_interpolators.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def assert_not_equal(x, y):
+    npt.assert_equal(np.any(np.not_equal(x, y)), True)
+
+
+def test_step_interpolator():
```
skoudoro, line 14: great ! you should do the same for test_helpers.py

also, check errors!  
- empty keyframes,  
- wrong key (not value)
- list instead of np.array, 
- etc...

skoudoro (commented):
Hi @m-agour,

I will let you update what we discuss and then continue to review. Let me know when the update is ready

m-agour:
Hi @skoudoro,
Thanks for your review. I made the changes and added the tests you mentioned above. It's ready now.

fury/ui/elements.py
```diff
@@ -3484,3 +3485,281 @@ def left_button_dragged(self,  i_ren, _obj, element):
...
+        Parameters
+        ----------
+        _scene : scene
+
+        """
+        self.panel.add_to_scene(_scene)
+        self._progress_bar.add_to_scene(_scene)
+        self.time_text.add_to_scene(_scene)
+
+    def _set_position(self, _coords):
```
filipinascimento, line 3759: Maybe raise a NotImplementedError for now, while this is not implemented.

fury/animation/timeline.py
```diff
@@ -0,0 +1,1499 @@
...
+        Other Parameters
+        ----------------
+        spline_degree: int, optional
+            The degree of the spline in case of setting a spline interpolator.
+
+        Notes
+        -----
+        If this evaluator is used to evaluate the value of actor's
+        properties such as position, scale, color, rotation, or opacity, it has
+        to return a value thathas the same shape as the evaluated property,
```
filipinascimento, line 383: If an evaluator is used to set the values of actor's properties...
thathas -> with

filipinascimento (changes requested):
The PR looks solid. It is working fine and I think it can be merged with just a few changes.

fury/ui/elements.py
m-agour: since these methods override methods on the UI class, and the UI requires these methods, I just want them to do nothing for now.

docs/tutorials/05_animation/viz_animation_using_time_equations.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+###############################################################################
+# Setting evaluator functions is the same as setting interpolators, but with
+# one extra argument: `is_evaluator=True` since these functions does not need
+# keyframes as input.
+timeline.set_position_interpolator(pos_eval, is_evaluator=True)
+timeline.set_rotation_interpolator(rotation_eval, is_evaluator=True)
+timeline.set_color_interpolator(color_eval, is_evaluator=True)
+timeline.set_interpolator('scale', scale_eval, is_evaluator=True)
```
skoudoro, line 61: this block needs more explanation. It seems obvious to you but it is clearly hard to understand for a researcher who starts with FURY.

Also, if `is_evaluator=False`, it does not work but the window appears. You should raise an error to explain. It is not the expected input set is_evalutor to True or something like that.

Currently, it is a bit too much magical without explanation.

docs/tutorials/05_animation/viz_keyframe_bezier_interpolator.py
```diff
@@ -0,0 +1,192 @@
...
+
+interactive = False
+
+if interactive:
+    showm.start()
+
+window.record(scene, out_path='viz_keyframe_animation_bezier_1.png',
+              size=(900, 768))
+
+###############################################################################
```
skoudoro, line 108: this line is not needed

docs/tutorials/05_animation/viz_keyframe_bezier_interpolator.py
```diff
@@ -0,0 +1,192 @@
...
+# Adding the callback function that updates the animation
+showm.add_timer_callback(True, 10, timer_callback)
+
+if interactive:
+    showm.start()
+
+window.record(scene, out_path='viz_keyframe_animation_bezier_2.png',
+              size=(900, 768))
```
skoudoro, line 192: The first animation is working.

The second animation generate a `Segmentation fault: 11`

docs/tutorials/05_animation/viz_keyframe_camera_animation.py
```diff
@@ -0,0 +1,190 @@
...
+###############################################################################
+# Adding the callback function that updates the animation
+showm.add_timer_callback(True, 10, timer_callback)
+
+interactive = False
+
+if interactive:
+    showm.start()
+
+window.record(scene, out_path='viz_keyframe_animation_camera.png',
```
skoudoro, line 189: This animation is really cool!

docs/tutorials/05_animation/viz_keyframe_color_interpolators.py
```diff
@@ -0,0 +1,104 @@
+"""
+==================
+Keyframe animation
+==================
+
+Color animation explained
```
skoudoro, line 6: need more explanation. too brief

skoudoro (approved):
This PR looks solid, great job @m-agour.

I appreciate all the tutorials. Some of them need much more explanation for a beginner

Despite my review, I will go ahead and merge this useful and valuable PR.  Please, try to address my concern on a new PR. 

Thanks!

m-agour:
@skoudoro Will fix these issues ASAP. Thank you!

PR #653, opened 2022-08-06
DrawPanel Feature: Grouping Shapes

ganimtron-10:
This feature adds grouping functionality to the shapes present in the `DrawPanelUI`.
PR #623 needs to be merged before this.

To group shapes:
- Go to selection mode
- Hold the `Ctrl` Key
- Select the shape you want to group together using `left mouse click`
- If you want to ungroup the shape, just select the shape again and it will be automatically deselected.

Features:
- Translation of all the shapes
- Rotation of all the shapes

Demo:
python_UiElpeke69


m-agour:
Hello @ganimtron-10,
Not sure if you noticed or not but all tests are failing for some reason.

nasimanousheh:
@ganimtron-10 Thank you for updating the PR with grouping the shapes functionality.
When another mode (for example delete) is selected, the white borders around the shapes and the rotation slide should get disappeared.

fury/ui/core.py
```diff
@@ -325,9 +327,13 @@ def mouse_move_callback(i_ren, obj, self):
             self.on_middle_mouse_button_dragged(i_ren, obj, self)
 
     @staticmethod
-    def key_press_callback(i_ren, obj, self):
+    def on_key_press_callback(i_ren, obj, self):
```
skoudoro, line 330: if you change that, please, check the whole codebase. I suppose, your PR is failing because of that. all `UI` depends on this

skoudoro (commented):
This PR needs to be rebased, also some conflict needs to be fixed.

fury/ui/elements.py
```diff
@@ -3084,6 +3084,113 @@ def directory_click_callback(self, i_ren, _obj, listboxitem):
...
+        """Update the position of all the shapes in the group.
+
+        Parameters
+        ----------
+        offset : (float, float)
+            Distance by which each shape is to be translated.
+
+        """
+        for shape in self.grouped_shapes:
+            shape.update_shape_position(shape.center + offset)
```
nasimanousheh, line 3181: In update_position function, keep the distance of the shapes (with each other) fixed. Otherwise, the shapes collapse in one position as you move them near the borders. To do this, the "shape.center" should be replace by "group.center" which represents the center of grouped shapes.

ganimtron-10:
I have updated the code and now we are able to translate the shapes at borders without overlapping on each.

python_br29AV7Bcx

ganimtron-10:
Currently, I am working on the tests for this PR.
The tests are behaving weirdly, sometimes it works fine some times it doesn't.
Till then you can play around with this feature and provide a review.

skoudoro:
There is some conflict in this PR. Also, all the tests are failing. 

I will wait for the  update before reviewing

ganimtron-10:
I tried various ways to find the issue and found out Disk2D's rotation was causing this failure so I removed that part and rerecorded the test but still as we can see the tests are still failing. Mohamed also tried running this test on his Windows, there the tests are passing sucessfully.

ganimtron-10:
Hello @skoudoro @Garyfallidis ,
According to the suggestions, I tried using the `Button2D` instead of the `RingSlider` to rotate the shapes, and maybe my guess was correct about the slider (that it was causing the tests to fail, but I can't validate it) because after these changes all the test is passing.
This is the link to see the changes and test results: https://github.com/ganimtron-10/fury/pull/8

Garyfallidis:
@ganimtron-10 I want to merge this but I saw that sometimes the "test_ui_draw_panel_grouping" is failing. Can you suggest how to move forward? Should we ignore the failing because this will come together with another PR? I am not sure how to proceed. Please advise.

ganimtron-10:
Hello @Garyfallidis ,
Currently, I am digging deep into the event propagation and finding out why the rotation function is called multiple times when clicked once. I think I would need more time to look into it.

As you suggest there are mainly two ways to merge this PR:
1. Skip these tests and create an issue regarding the same and fix it as soon as possible (but then these features remain untested and may cause some issues in future).
2. Second would be to replace the current rotation slider with two button to manually rotate the shape by some fixed step value on both clockwise as well as counter clockwise direction, but in this case the flexibility of rotation reduces a lot. (I tried integrating spinbox UI with this, but still theres some kind of error in event propogation which again hamper the working.)

Let me know, how should I go ahead with it!!

fury/ui/tests/test_elements.py
```diff
@@ -1124,6 +1124,37 @@ def test_ui_draw_panel_rotation(interactive=False):
...
+
+    drawpanel = ui.DrawPanel(size=(600, 600), position=(30, 10))
+
+    # Assign the counter callback to every possible event.
+    event_counter = EventCounter()
+    event_counter.monitor(drawpanel)
+
+    current_size = (680, 680)
+    show_manager = window.ShowManager(
+        size=current_size, title="DrawPanel Grouping UI Example")
```
Garyfallidis, line 1140: You do not need initialize here?

Garyfallidis:
Let's investigate deeper.

fury/ui/tests/test_elements.py
ganimtron-10: Nope, it is internally called in the `show_manager.start()`

ganimtron-10:
I tried investigating the issues but was unable to narrow it down.

I can show that the code repeats itself from the `propagate_event` but am unable to find what cause it to start and how it automatically stops after an even number of iterations.

skoudoro:
Thank you for your feedback @ganimtron-10.

We are also still investigating, and I wonder if it is not the `add_callback` function is the `customInteractor` that is failing sometimes. 

Investigations still in process. Thank you for still looking into it

Garyfallidis:
@ganimtron-10 please update these PRs.

PR #655, opened 2022-08-08
Fix test skybox

guaje:
This PR fixes the `test_skybox` issues and enables `test_scene` since the problem was related to the skybox functionality.

The problem lies in testing with `window.snapshot(scene)`, which could be the cause of potential issues with the tests of the material module. Now that we are working on the shader API we should consider ways to test functions running at different levels of the OpenGL rendering pipeline.


skoudoro (approved):
Hi @guaje,

> The problem lies in testing with window.snapshot(scene), which could be the cause of potential issues with the tests of the material module. Now that we are working on the shader API we should consider ways to test functions running at different levels of the OpenGL rendering pipeline.

Can you be more explicit? Not sure I understand what you mean here.

Otherwise, it looks good to me. I will go ahead and merge this PR as soon as you answer the question above

guaje:
> Hi @guaje,
> 
> > The problem lies in testing with window.snapshot(scene), which could be the cause of potential issues with the tests of the material module. Now that we are working on the shader API we should consider ways to test functions running at different levels of the OpenGL rendering pipeline.
> 
> Can you be more explicit? Not sure I understand what you mean here.
> 
Sure, to test many shader-related functionalities (Scene Skybox/IBL, Blinn-Phong or PBR materials), we take screenshots of the scene (`window.snapshot`) as seen [here](https://github.com/fury-gl/fury/blob/3830c313a24d8636271952611b86179eeebeba06/fury/tests/test_material.py#L40). Then we treat the screenshot as an MxNx3 matrix, and we assert for expected colors on selected pixels. Although this could make sense to check for expected diffuse/specular areas of the actor according to the applied material, I've seen this causing some inconsistencies when it comes to testing.

For these particular tests (`test_scene` and `test_skybox`), I removed all the screenshot-based tests and replaced them by asserting that the IBL/Skybox conditions are fulfilled. However, for some other cases: materials module, @tvcastillod's upcoming VS/GS billboard PRs or @m-agour's GS billboard PR, this strategy might not be available.

skoudoro:
ok, got it! thank you for this clear explanation. merging!

PR #661, opened 2022-08-15
Fixed `vector_text` not following camera

m-agour:
There is an issue with `vector_text` which was not following the camera because no camera was assigned for it to follow.
Not sure if this is intended for the vector text to follow the camera or should it also have a direction as well if user wants.
Below is a comparison
image
image

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -2384,8 +2384,9 @@ def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
     return bb_actor
 
 
-def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.0),
-                color=(1, 1, 1), direction=(0, 0, 1), align_center=False):
+def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.2),
+                color=(1, 1, 1), direction=(0, 0, 1), extrusion=0.0,
+                align_center=False):
     """Create a label actor.
 
     This actor will always face the camera
@@ -2402,6 +2403,8 @@ def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.0),
         Label color as ``(r,g,b)`` tuple.
     direction : (3,) array_like, optional, default: (0, 0, 1)
         The direction of the label. If None, label will follow the camera.
+    extrusion : float, optional
+        The extrusion amount of the text in Z axis.
     align_center : bool, optional, default: True
         If `True`, the anchor of the actor will be the center of the text.
         If `False`, the anchor will be at the left bottom of the text.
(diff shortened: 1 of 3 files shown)


fury/actor.py
```diff
@@ -2383,8 +2383,8 @@ def billboard(centers, colors=(0, 1, 0), scales=1, vs_dec=None, vs_impl=None,
     return bb_actor
 
 
-def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.2),
-                color=(1, 1, 1)):
+def vector_text(text='Origin', pos=(0, 0, 0), direction=None,
```
skoudoro, line 2386: Please, do not change the order of the argument. Otherwise, we need to start a deprecation cycle.

some people are using this function for a long time and the next release will mess up their code if we keep it like that.

Also, we should keep the default behavior for the same reason.

fury/actor.py
```diff
@@ -2416,16 +2421,45 @@ def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.2),
     """
     atext = VectorText()
     atext.SetText(text)
-
     textm = PolyDataMapper()
+
+    t = Transform()
```
skoudoro, line 2426: avoid 1 letter variable. 3 letters minimum.

skoudoro (commented):
Hi @m-agour,

Thank you for doing this!

It was intended for the vector text to not follow the camera. 

So, you should keep the current behavior as it was and add this new functionality as optional.

Also, do not forget to add some tests. 

Thanks!

m-agour:
Hi @skoudoro ,
Thank you for reviewing this PR. 
It makes sense to keep the defaults as you said. I thought since there is no way to change direction and the base actor is `vtkFollower`, it was meant to follow the camera so I set it to do that by default.  
Sure I will fix this and add tests and let you know when it's ready.

skoudoro:
is it ready to be reviewed @m-agour ?

m-agour:
Yes I just pushed some changes and it's ready now

skoudoro:
Hi @m-agour,

Can you address the requested changes above? thanks

m-agour:
> Hi @m-agour,
> 
> Can you address the requested changes above? thanks

I addressed all of them and updated the tests. It just needs your final review.

fury/actor.py
```diff
@@ -2395,10 +2397,17 @@ def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.2),
         Text for the label.
     pos : (3,) array_like, optional
         Left down position of the label.
+    direction : (3,) array_like, optional, default: (0, 0, 1)
+        The direction of the label. If None, label will follow the camera.
```
skoudoro, line 2401: should be below `color` in order to respect the argument order in the function

fury/actor.py
```diff
@@ -2395,10 +2397,17 @@ def vector_text(text='Origin', pos=(0, 0, 0), scale=(0.2, 0.2, 0.2),
...
     scale : (3,) array_like
         Changes the size of the label.
     color : (3,) array_like
         Label color as ``(r,g,b)`` tuple.
+    align_center : bool, default: True
+        If `True`, the anchor of the actor will be the center of the text.
+        If `False`, the anchor will be at the left bottom of the text.
+    extrusion: float, default: 0
+        Extrusion length of the text gives it a 3D look.
```
skoudoro, line 2410: `extrusion`?  it does not appear on the function arguments above

also, it is not used

skoudoro (commented):
oh ok, I see, I never submitted this review below.

fury/actor.py
m-agour: I was using it to specify extrusion but I thought making scale in Z axis do the job is a better choice

m-agour: @skoudoro  It's done

skoudoro (commented):
Hi @m-agour,

I have a small comment on this PR about the extrusion. See below. I really would like to keep the same behavior as before and some text on tutorials is unreadable due to the scale. 

I think we should split the `scale` and `extrusion` variables into 2 different arguments.

Apart from that, it is good to go

skoudoro (approved):
All good, Thank @m-agour, merging

PR #673, opened 2022-08-30
DrawPanel Update: Removing `in_progress` paramter while drawing shapes

ganimtron-10:
Currently `draw_shape` method that creates shapes on `DrawPanel` was having a parameter called `in_progress` which was keeping a check on whether the shape are already created or not.
This parameter does give a clear idea as why and for what it is used.
So removing this and separating the functions to make things more clear.


skoudoro (approved):
LGTM, thanks @ganimtron-10 merging

PR #674, opened 2022-08-30
DrawPanel Update: Seperating tests to test individual features

ganimtron-10:
As we are now adding new features to the `DrawPanel` the current test is becoming bigger and bigger.
Due to this creating, debugging, and managing tests is becoming harder.
So to keep things simple, separating tests to validate individual parts or features of the `DrawPanel`.


skoudoro:
I do not see any issue. all the tests are running ok in my laptop so I am going ahead and merge this PR.

Thanks @ganimtron-10

PR #679, opened 2022-09-03
DrawPanel Update: Moving repetitive functions to helpers

ganimtron-10:
Some code from the `DrawShape` UI is going to be reused in the Grouping Shape Feature #653 .
So to avoid these repetitive codes, moving them into `helper.py` file.

- [x] cal_bounding_box
- [ ] rotate


skoudoro (approved):
All good, Thank you @ganimtron-10, merging

PR #689, opened 2022-09-18
GLTF actor colors from material

xtanion:
This small PR adds the ability to add colors to the actor from `materials` in the glTF file.

Example: Run the `BrainStem` model in `viz_gltf.py`
viz_gltf

Other models with material colors:
- BoxAnimated
- RiggedSimple
- [2 Cylinder Engine](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/2CylinderEngine)
- [Water Bottle](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/WaterBottle)
- [Buggy](https://github.com/KhronosGroup/glTF-Sample-Models/blob/master/2.0/Buggy) and many more.


fury/gltf.py
```diff
@@ -169,6 +171,7 @@ def load_mesh(self, mesh_id, transform_mat):
 
             if attributes.NORMAL is not None and self.apply_normals:
                 normals = self.get_acc_data(attributes.NORMAL)
+                normals = transform.apply_transfomation(normals, transform_mat)
```
skoudoro, line 174: typo to change everywhere: `apply_transformation`

fury/gltf.py
```diff
@@ -77,6 +77,8 @@ def actors(self):
             if self.materials[i] is not None:
                 base_col_tex = self.materials[i]['baseColorTexture']
                 actor.SetTexture(base_col_tex)
+                base_color = self.materials[i]['baseColor']
+                actor.GetProperty().SetColor(tuple(base_color[:3]))
```
skoudoro, line 81: do we have a function in `fury.utils` for doing that?

skoudoro (commented):
Hi @xtanion,

See below for some comments

fury/gltf.py
xtanion: No, we have a similar function `set_polydata_colors`

skoudoro:
merging, Thank @xtanion

PR #692, opened 2022-09-21
Set position and width of the `PlaybackPanel`

m-agour:
In this PR, I introduce some improvements to the `PlaybackPanel` by allowing setting its position and width as well as making it responsive to window resizing by default.

https://user-images.githubusercontent.com/63170874/191489960-0ccb0b26-983c-4708-bf27-905afb55307c.mp4

### DIFF
--- a/fury/ui/elements.py
+++ b/fury/ui/elements.py
@@ -3704,25 +3704,30 @@ class PlaybackPanel(UI):
         self.current_time = 0
 
     def play(self):
+        """Play the playback"""
         self._playing = True
         self._play_pause_btn.set_icon_by_name('pause')
         self.on_play()
 
     def stop(self):
+        """Stop the playback"""
         self._playing = False
         self._play_pause_btn.set_icon_by_name('play')
         self.on_stop()
 
     def pause(self):
+        """Pause the playback"""
         self._playing = False
         self._play_pause_btn.set_icon_by_name('play')
         self.on_pause()
 
     def loop(self):
+        """Set repeating mode to loop."""
         self._loop = True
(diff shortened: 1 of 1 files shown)


fury/ui/elements.py
```diff
@@ -14,7 +14,7 @@
 import numpy as np
 
 from fury.data import read_viz_icons
-from fury.lib import PolyDataMapper2D
+from fury.lib import PolyDataMapper2D, Command
```
skoudoro, line 17: pep8: Can you remove `PolyDataMapper2D` it seems to not be used

fury/ui/elements.py
```diff
@@ -3578,14 +3578,15 @@ class PlaybackPanel(UI):
        such as play, pause, stop, and seek.
     """
 
-    def __init__(self, loop=False, position=(0, 0)):
-        super(PlaybackPanel, self).__init__()
-        self.position = position
+    def __init__(self, loop=False, position=(0, 0), width=None):
```
skoudoro, line 3581: I wonder if make sense to have position at `(0,0)` instead of `None`

skoudoro (commented):
Hi @m-agour,

Thank you for this feature, it works well. See below for few comment but it should be ready to merge soon

fury/ui/elements.py
m-agour: I think it should be `(0, 0)` to keep consistency with other UI elements that has `position=(0, 0)`. That is true if the position is measured from the bottom left of the panel

skoudoro: ok, fair enough. LGTM, merging

PR #701, opened 2022-10-08
Updating `fetch_viz_new_icons` to fetch new icons

ganimtron-10:
As we are adding new features to `DrawPanel`, new icons are used to select and play around with it. Adding those icons to `fury-data` [here](https://github.com/fury-gl/fury-data/pull/11). Updating the `fetch_viz_new_icons` to fetch these icons.
This PR needs to be merged before #695 and #696 .


ganimtron-10:
These tests might fail as the PR https://github.com/fury-gl/fury-data/pull/11 isn't merged yet, so it should be merged before this.

skoudoro:
Thank you for adding thoses icons @ganimtron-10. Merging

PR #708, opened 2022-11-15
Strips to triangles 

nasimanousheh:
TriangleFilter() is added to vtkFiltersCore Modules. 
TriangleFilter() converts input polygons and strips to triangles. This allows Fury to do fast selection.


Garyfallidis:
This PR needs to be updated to master.

PR #709, opened 2022-11-17
Commented the self.initialize

nasimanousheh:
When I upgraded FURY in Furious Atoms [(https://furiousatoms.com),](https://furiousatoms.com) I noticed a second window appears and the loaded file is shown in the second window as you see here:

demo

The issue is in `ShowManager`. I commented the lines 395 and 396 in `window.py` and now it works perfectly:

demo2

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -392,8 +392,8 @@ class ShowManager(object):
         self.iren.SetInteractorStyle(self.style)
         self.iren.SetRenderWindow(self.window)
 
-        if is_win:
-            self.initialize()
+        # if is_win:
+        #     self.initialize()
 
     def initialize(self):
         """Initialize interaction."""


Garyfallidis:
Okay, I think people were not careful with where they added self.initialize in the ShowManager. @nasimanousheh can you also show the gif of the changes. Initialize was always on the ShowManager methods but not in the constructor. Otherwise every other GUI system will be incompatible with FURY.

Garyfallidis:
When I see gif of changes. I actually mean the git diff between current version and a previous version before that initialization was added in the constructor.

nasimanousheh:
Here is the difference:

demo5

Garyfallidis:
Addressed in #744

PR #710, opened 2022-11-17
Principled update

guaje:
This PR includes the changes to the Principled BRDF used for the Horizon manuscript.


skoudoro:
This PR start to be big and hard to review. is it possible to split it in smaller PR?

Also, when do you think we could have a first initial version ? Thank you for the feedback

guaje:
> This PR start to be big and hard to review. is it possible to split it in smaller PR?

Definitely! I'll remove many of the python files, so in the end the PR will consist of a few modifications of the `material.py` and `test_material.py` files and the addition of all the PBR/principled functions.

> Also, when do you think we could have a first initial version ? Thank you for the feedback

I'm trying to fix something with the geometry shadowing functions and then it'll be ready.

Garyfallidis:
@guaje we need to rebase this.

Garyfallidis:
Great work @guaje !

Issue #578, opened 2022-04-20
Pytest failing on mac when calling showm.initialize() but no start in any test


filipinascimento:
## Description
We are adding new multithreading functionality to the show manager. However, by adding a new test to the `test_window.py`, we start getting segmentation fault on macs. By further isolating the issue, we found out that this seems to be related to calling the `initialize()` method for a show manager without starting it (which is already present in the `test_save_screenshot` test.

## Way to reproduce
By using the following code as a pytest

```python
from fury import window

def test1():
    scene = window.Scene()
    show_m = window.ShowManager(scene)
    show_m.initialize()

def test2():
    scene = window.Scene()
    show_m = window.ShowManager(scene)
    show_m.initialize()
    show_m.start()
```

it fails with a segmentation fault on macs (not sure about other OSes).

This is probably a vtk bug, so I will further explore it using only vtk functions.

skoudoro:
Hi @filipinascimento,

I can not reproduce the error on my old mac (macbook pro mid-2015). I will try with another OS.

![image](https://user-images.githubusercontent.com/23106443/180563305-362b275a-4a90-44a0-b9b3-185ec6c24252.png)

PR #712, opened 2022-11-26
Fix: segfault created by record method

### LINKED ISSUE #603: UI tests are failing in Ubuntu OS due to a "segmentation error"
## Description
UI tests are failing because of a "segmentation error" in Ubuntu only.
@xtanion confirmed it on his machine.
The error message is:
`tests/test_elements_callback.py Fatal Python error: Segmentation fault`
I tried to trace back the cause of this error and it's not caused by the test in which this error occurred.
Running only `test_elements_callback.py` gives a success.

## Way to reproduce
Running the UI tests in Ubuntu OS.

devmessias:
This closes #603  and maybe  #578

### DIFF
--- a/fury/tests/test_window.py
+++ b/fury/tests/test_window.py
@@ -382,9 +382,52 @@ def test_stereo():
     npt.assert_array_equal(stereo[150, 150], [0, 0, 0])
 
 
-@pytest.mark.skipif(skip_linux or skip_win,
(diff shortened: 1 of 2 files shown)


skoudoro (approved):
LGTM. Thank @devmessias, I will go ahead and merge this PR. 

I think there are other place where we need to removerenderer and finalize

PR #720, opened 2023-01-03
fix image load flip issue

Mk996:
This PR aims to fix the image loading (Flip) issue.

The possibility is that the underlying pillow must have changed something in the newer versions. That will now not require flipping the NumPy array.


Garyfallidis:
It will be great to track at which version of pillow the behavior changed so we can update our dependency requirements.

ganimtron-10:
Hey @Mk996 ,
Thanks for doing this! I haven't tested this yet but make sure to pass all the tests.
You can ignore the test fails that aren't due to your code, but I can see some of the tests related to ` interpolators` is failing due to some changes you did. PTAL at them.

Thanks!

skoudoro:
superseded and fixed by #728. closing

PR #721, opened 2023-01-09
Fix: Color changes on docs pages fixed

Mk996:
## Fix
- This PR fixes the font and footer color change in the documentation pages.

## Bug Details
- The bug got introduced due to the change of the version of the PyData theme over time. The code to configure the theme's colors was written in the style of `v0.8.1` but the current version getting used is, `v0.10.1`.

## Changes
- I have made changes in the `custom.css` file for defining the color scheme.
- I have made changes in the `sphinx-footer.css` file, originally the class applied to the footer from the theme was `.footer` but after the newer version, it started using the bootstrap, and the footer is getting the `.bd-footer` class applied from the theme.
- I have also removed all the error-creating imports, which were non-existent.
- I have fixed the hover issue on the fury logo text.
- I have introduced a common `navbar.css` file for a single source of truth of the navbar CSS and this file will grow as we will refractor all the hodgepodge code of the different CSS files.

## Backward Compatibility
- This PR will support the older version of the pyData theme.


skoudoro (changes requested):
Hi @Mk996,

Thank you for this update.

- It is strange, but all the menu links from Homepage stopped to work. We cannot go to the installation page for example. I suppose, there is a conflict somewhere.
-  Also, see below: (The top right menu should be red and not black)
image

Apart from that, it looks good. 

Thanks

docs/source/_static/css/navbar.css
```diff
@@ -0,0 +1,31 @@
+/* Introducing a new css file for handling navbar css */
```
Mk996, line 1: Introduced a new CSS file which will be the single source of truth for both `home.html`, `custom-title.html`, and `custom-navbar.html`.

This file will be expanded in the upcoming PR, I will shift the navbar CSS to this file

docs/source/_static/vendor/new-promotions.css
```diff
@@ -12816,134 +12816,6 @@ only screen and (min-device-pixel-ratio:2) {
     background-color: #fff
 }
 
-@font-face {
```
Mk996, line 12819: Not used font faces and the files were also not present, causing errors in the console.

docs/source/_static/vendor/styles.css
```diff
@@ -102,12 +102,12 @@ em {
     cursor: pointer;
 }
 
-#navbar-item {
```
Mk996, line 105: removed the duplicate id from navbar item and converted the CSS to follow the class

docs/source/_templates/custom-footer.html
```diff
@@ -8,10 +8,14 @@
                 <h5 class="footer-h5">About</h5>
```
Mk996, line 8: formatted file

docs/source/_templates/custom-navbar.html
```diff
@@ -1,6 +1,6 @@
 <ul id="navbar-main-elements" class="navbar-nav">
-    <li class="nav-item ">
-        <a class="nav-link" href="installation.html">Installation</a>
```
Mk996, line 3: Updated the installation link to be dynamic as it was failing in docs page if you got API reference page

docs/source/_templates/home.html
```diff
@@ -82,19 +83,19 @@
             <div id="navbar-menu" class="navbar-menu is-static">
 
                 <div class="navbar-end">
-                    <a href="installation.html" id="navbar-item" class="navbar-item is-secondary">
```
Mk996, line 85: Good coding practice,
removed duplicate id attribute as that should be unique across elements

docs/source/_templates/home.html
```diff
@@ -774,11 +775,9 @@ <h5>Support</h5>
     <script src="https://unpkg.com/feather-icons"></script>
     <script src="_static/js/fresh.js"></script>
     <script src="_static/js/popup.js"></script>
-    <script src="_static/js/jquery.panelslider.min.js"></script>
```
Mk996, line 777: removed few non-existing import, removing errors from the browser console

skoudoro (approved):
All good, thanks for this work @Mk996 , merging!

PR #729, opened 2023-01-26
Fix links in CONTRIBUTING.rst

RelativisticMechanic:
The original file seems to have used markdown links instead of reStructuredText links. It must have been a mistake. I have fixed the three links.


skoudoro (approved):
Good catch! Thank you @RelativisticMechanic.

Looks good, merging and thank for looking at the documentation!

PR #744, opened 2023-02-22
Update initialize management

skoudoro:
this PR:
- remove initialize from `ShowManager` init.  : it creates a unwanted window (render is call inside initialize)
- add initialize in `add_timer_callback` .  iren need to be initialized before adding an observer

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -1,10 +1,8 @@
 # -*- coding: utf-8 -*-
-import time
-from threading import Lock
-
 import gzip
 import time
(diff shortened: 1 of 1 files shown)


Garyfallidis:
A clarification here. The initializer is needed only with ``timer_callbacks`` but not in multithreaded cases.

skoudoro:
> A clarification here. The initializer is needed only with timer_callbacks but not in multithreaded cases.

I confirm

Garyfallidis:
Merging. This closes  https://github.com/fury-gl/fury/pull/709 too.

PR #747, opened 2023-03-02
Connected the sliders to the right directions

nasimanousheh:
In viz_ui_slider.py, both horizontal and vertical sliders were connected to horizontal movement "cube.SetPosition(value, 0, 0)".
I created two functions for horizontal and vertical movements and connected them to the right sliders.


skoudoro (commented):
All good, thank you for this fix, merging

PR #748, opened 2023-03-03
some fixs and ex addition in docstrings in actor.py

SaraSaadoun:
in this PR:
<ul>
  <li>fixed things in docstrings such as underline length, some spaces, and other minor fixs</li>
  <li>added a missing parameter in the docstring of marker function "edge_opacity"</li>
  <li>added an example of marker function in its docstring</li>
</ul>

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -123,8 +123,8 @@ def slicer(
         as slices. The key method of this object is ``display_extent`` where
         one can input grid coordinates and display the slice in space (or grid)
         coordinates as calculated by the affine parameter.
-
     """
+
     if value_range is None:
         value_range = (data.min(), data.max())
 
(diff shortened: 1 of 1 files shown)


fury/actor.py
```diff
@@ -123,8 +123,8 @@ def slicer(
         as slices. The key method of this object is ``display_extent`` where
         one can input grid coordinates and display the slice in space (or grid)
         coordinates as calculated by the affine parameter.
-
     """
+
```
skoudoro, line 127: Why this change everywhere?  we prefer the opposite. (recommended by pycodestyle).  We are open to change that but you need to explain why. 

Thanks!

skoudoro (commented):
Hi @SaraSaadoun,

Thank you for doing this. 

Which code formatter are using?  I know that some rules differs. For example, I see that it removes the empty in the docstring and add one at the beginning of the code. We prefer the opposite. 
Please, explain this change and we can move forward

skoudoro:
You can also look numpy doc style.  This is what we follow

SaraSaadoun:
I've fixed them now, actually, the missing parameter for example is caught by Pycharm IDE, not the new line one. 
I think I was wrong about this one.

skoudoro (approved):
Looks good to me. 

Thank you for doing this @SaraSaadoun. Merging

PR #749, opened 2023-03-03
Remove initialize

skoudoro:
- Small PR to remove initialize everywhere. 
- in `UI` module, `iren.initialize()` needs to be called since we add an Observer sometimes


skoudoro:
We stopped to use vtk. Closing

PR #756, opened 2023-03-08
Triangle strips 2 Triangles

Garyfallidis:
Adds an option to change surface representation in streamtube from triangle strips to triangles. This makes life easy with the SelectionManager. This closes PR https://github.com/fury-gl/fury/pull/708 but I am also adding an issue to improve direct picking of triangle strip representations. See here https://github.com/fury-gl/fury/issues/757

In order to know if an actor is using triangle strips or not, you need to use something like
``actor.GetMapper().GetInput().GetStrips()``

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -54,6 +54,7 @@ from fury.lib import (
     TransformPolyDataFilter,
     TubeFilter,
     VectorText,
+    TriangleFilter,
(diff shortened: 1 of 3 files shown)


skoudoro:
Thank you @Garyfallidis for adding the tests. Merging

PR #764, opened 2023-03-17
ComboBox2D drop_down_button mouse callback was inside for loop

dwijrajhari:
The left_mouse_button_clicked callback for the drop_down_button was inside a for loop. Fixed indentation.


skoudoro (commented):
Good catch, Thanks @dwijrajhari.

Can you also update the tests? Have you tried to run the tests? they should failed after this change

dwijrajhari:
Thanks @skoudoro
I looked into `fury/ui/tests` but couldn't find anything to change there. The tests that are failing were also failing before (on my machine) and I think are not related to this change.

skoudoro:
thanks @dwijrajhari

Issue #770, opened 2023-03-22
 Directions of arrow actor do not change in `repeat_primitive = False` method (VTK)


Clarkszw:
## Description
There are two issues present in the picture below:
Note: red arrow is `repeat_primitive = True`, green arrow is `repeat_primitive = False`
1. Directions of arrow actor do not change when applying no_repeat_primitive (`repeat_primitive = False`) method (VTK)
2. When direction is [-1, 0, 0], the repeat_primitive (`repeat_primitive = True`) does not return the arrow actor.

<img width="660" alt="image" src="https://user-images.githubusercontent.com/110199757/226866982-db2915a6-f828-4f97-9f03-28419a9c679e.png">

below is the code:
```python
from fury import window, actor
import numpy as np

scene = window.Scene()
showm = window.ShowManager(scene,size=(1024, 720), reset_camera=False)
showm.initialize()

centers = np.array([[0, 0, 0]])
directions = np.array([1, 0 ,0 ])
heights = 1
colors_primitive = np.array([1,0,0])
colors_no_primitive = np.array([0,1,0])

for i in range(35):
    
    centers = centers + np.array([[1.2, 0, 0]])

    theta_rad = np.deg2rad(10)
    rotation_z = np.array([[np.cos(theta_rad), -np.sin(theta_rad), 0],
                           [np.sin(theta_rad),  np.cos(theta_rad), 0],
                           [0,  0, 1]])

    
    directions = np.dot(rotation_z, directions.T).T
    
    arrow_primitive = actor.arrow(centers, directions, colors_primitive, 
                           heights=heights, repeat_primitive=True)

    arrow_no_primitive = actor.arrow(centers, directions, colors_no_primitive, 
                           heights=heights, repeat_primitive=False)

    scene.add(arrow_primitive,arrow_no_primitive)

showm.start()
```

skoudoro:
Thank you for this issue and analyse.

Clearly, we have to spend time to understand what is the rotation difference between primitive and vtk source. This will fix the issue of many actors

Clarkszw:
I am investigating it now, interesting to refresh some mathematic concept in rotation ;)

Clarkszw:
The problem is `directions = np.array([1, 0 ,0 ])`
It is ok for `repeat_primitive` to use 1d array, it will be processed to `np.ndarray` in the process.
But for VTK method, `directions` parameter has to be `np.ndarry(N,3)`

[...]

PR #771, opened 2023-03-23
add one condition in `repeat_primitive` to handle direction [-1, 0, 0], issue #770

Clarkszw:
Fix the issue #770.

The cause is `directions = np.array([1, 0 ,0 ])`, which is a (3, ) 1 dimension array.
It is ok for `repeat_primitive()` to use 1d array, it will be processed to `np.ndarray` in the function.
But for VTK method, directions parameter has to be `np.ndarry(N,3)`

Issue 1 can be resolved by modify the directions to `directions = np.array([[1, 0, 0]])` as a (1,3) ndarray.

For Issue 2, I have add a condition to cover the case directions = [-1, 0, 0]

image

Code to reproduce:
```python
import numpy as np

scene = window.Scene()
showm = window.ShowManager(scene,size=(1024, 720), reset_camera=False)
showm.initialize()

centers = np.array([[0, 0, 0]])
directions = np.array([[1, 0 ,0 ]])
colors_primitive = np.array([1,0,0])
colors_no_primitive = np.array([0,1,0])

for i in range(35):
    
    centers = centers + np.array([[1.2, 0, 0]])

    theta_rad = np.deg2rad(10)
    rotation_z = np.array([[np.cos(theta_rad), -np.sin(theta_rad), 0],
                           [np.sin(theta_rad),  np.cos(theta_rad), 0],
                           [0,  0, 1]])

    
    directions = np.dot(rotation_z, directions.T).T
    
    arrow_primitive = actor.arrow(centers, directions, colors_primitive, repeat_primitive=True)
    
    centers_2 = centers + np.array([[0, 1.2, 0]])

    arrow_no_primitive = actor.arrow(centers_2, directions, colors_no_primitive, repeat_primitive=False)

    scene.add(arrow_primitive,arrow_no_primitive)

showm.start()
```

### DIFF
--- a/fury/primitive.py
+++ b/fury/primitive.py
@@ -203,11 +203,18 @@ def repeat_primitive(
             dirs = dirs / dir_abs
             v = np.cross(normal, dirs)
             c = np.dot(normal, dirs)
+
             v1, v2, v3 = v
-            h = 1 / (1 + c)
+
             Vmat = np.array([[0, -v3, v2], [v3, 0, -v1], [-v2, v1, 0]])
 
-            rotation_matrix = np.eye(3, dtype=np.float64) + Vmat + (Vmat.dot(Vmat) * h)
+            if c == -1.0:
+                rotation_matrix = -np.eye(3, dtype=np.float64)
+            else:
+                h = 1 / (1 + c)
+                rotation_matrix = np.eye(3, dtype=np.float64) + \
+                    Vmat + (Vmat.dot(Vmat) * h)
+
         else:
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Thank you for looking into that @Clarkszw?

Can you also update the unit tests?

Thank you!

Clarkszw:
unit test for directions [-1, 0, 0], [0, -1, 0] and [0, 0, -1] added

skoudoro (approved):
Thank you @Clarkszw, LGTM! merging

Issue #458, opened 2021-07-09
Standardize actors colors (follow-up of #430)


skoudoro:
Hi All, 

This PR can be merged so I will go ahead. However, we need to clarify the rules concerning the color ([0-1], vs [0-255]).

Let's create an issue to remember that we need to handle the color in all actors. I suppose, there is a function for that in utils.py or we need to create it.

We also need to manage the out of bounds

_Originally posted by @skoudoro in https://github.com/fury-gl/fury/issues/430#issuecomment-877037326_

Aditi840:
I saw this issue open, can I work on it.

Clarkszw:
How is going with this? @Aditi840

skoudoro:
I think you can go ahead @Clarkszw 👍🏾

KeaganKozlowski:
Is this issue still open to work on?

PR #773, opened 2023-03-26
Standardize actors colors

Clarkszw:
I have added a new function ~~`check_color_range()`~~ `normalize_color()' in utils.py for #458 
It does two things:
1. check if the color array is in range [0, 1]
2. if not, it will ~~change it to red color [1, 0, 0] and print the out of bounds information~~ normalize the color array.

I have written an unit test for this function and added in test_utils.py

All tests in test_actors.py and test_utils.py are passed :)


fury/utils.py
```diff
@@ -1547,6 +1548,55 @@ def color_check(pts_len, colors=None):
...
+def check_color_range(color_array):
+    """
+    Check if all values in a color array are within the range [0, 1].
+    If there is out of bounds values, the function sends a message, 
+    replaces the out-of-bounds colors with the red color [1, 0, 0].
+
+    Args:
+        color_array (numpy.ndarray): An array of shape (N, 3) or (N, 4).
+
+    Returns:
```
skoudoro, line 1560: no colon

fury/utils.py
```diff
@@ -1547,6 +1548,55 @@ def color_check(pts_len, colors=None):
...
+        return color_array
+
+    # Convert tuple or list to ndarray (N, 3)
+    if isinstance(color_array, list):
+        color_array = np.asarray(color_array)
+
+    is_tuple = False
+    if isinstance(color_array, tuple):
+        is_tuple = True
+        color_array = np.asarray(color_array)
```
skoudoro, line 1575: `isinstance(color_aray, (list, tuple))` instead

skoudoro (commented):
Hi @Clarkszw,

Thank you for starting this.

See below for some comments:

1. Concerning docstring, I recommend you looking at the numpy style guide. Many python projects follow those rule which you are not following right now on this PR. here the link:
https://numpydoc.readthedocs.io/en/latest/format.html

2. Why red?  I recommend you to normalize your color values to go from [0-255] to [0-1]

fury/tests/test_utils.py
```diff
@@ -865,6 +865,43 @@ def test_color_check():
...
+    npt.assert_array_equal(utils.check_color_range(
+        invalid_color_array), invalid_color_array_expected)
+
+    # Test for input of type tuple
+    color_tuple = (0.1, 0.2, 0.3)
+    color_tuple_expected = (0.1, 0.2, 0.3)
+    assert utils.check_color_range(color_tuple) == color_tuple_expected
+
+    # Test for input of type list (invalid)
+    color_list = [100, 150, 200, 0.4]
```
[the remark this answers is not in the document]
Clarkszw: That's a good idea to normalize [0, 255] to [0, 1], I will do it.

The initiation is to take care of all out of bounds even more than 255 and less than 0, then give a red color to let the process finish and show a hint and warning to the user. This will also skip the modification of all actor docstring, saying color parameter accept both [0,1] and [0,255].

But I agree normalization is more user friendly and the value more than 255 can be taken care by other function that process rgb values.

fury/utils.py
Clarkszw: It was `isinstance(color_aray, (list, tuple))`, but when I went through all the test, I found it is a requirement for input 1d list keeps as a 1d list and input tuple keeps a tuple. Therefore you can find in the end of the function, there is one line to convert np.ndarray back to tuple if it is the case.

fury/utils.py
```diff
@@ -1547,6 +1548,55 @@ def color_check(pts_len, colors=None):
...
+
+    elif color_array.ndim == 2:
+        for i, row in enumerate(color_array):
+            if np.any(row > 1) or np.any(row < 0):
+                print(f"{row} in the color array are outside the valid range [0, 1]")
+                color_array[i] = [1, 0, 0, 1][:len(row)]
+                print("It has been modified to red color")
+
+    if is_tuple:
+        color_array = tuple(color_array)
```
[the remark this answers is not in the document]
Clarkszw: Some actors function accept tuple as input to apply the same color in all N actors. Therefore I add this line to convert the (1,3) ndarray back to tuple.

fury/utils.py
```diff
@@ -1547,6 +1548,55 @@ def color_check(pts_len, colors=None):
     return color_array, global_opacity
 
 
+def check_color_range(color_array):
+    """
+    Check if all values in a color array are within the range [0, 1].
+    If there is out of bounds values, the function sends a message, 
+    replaces the out-of-bounds colors with the red color [1, 0, 0].
+
+    Args:
```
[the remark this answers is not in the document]
Clarkszw: modified

fury/utils.py
Clarkszw: Done 😃

fury/tests/test_utils.py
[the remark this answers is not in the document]
Clarkszw: `normalize_color()` is done.

fury/utils.py
Clarkszw: It is required for in the color_check() function in actor.py
```
def color_check(pts_len, colors=None):
        """
        Returns a VTK scalar array containing colors information for each one of
        the points according to the policy defined by the parameter colors.

        Parameters
        ----------
        pts_len : int
            length of points ndarray
        colors : None or tuple (3D or 4D) or array/ndarray (N, 3 or 4)
            If None a predefined color is used for each point.
            **If a tuple of color is used. Then all points will have the same color.
            If an array (N, 3 or 4) is given, where N is equal to the number of
            points. Then every point is colored with a different RGB(A) color.**

        Returns
        -------
        color_array : vtkDataArray
            vtk scalar array with name 'colors'.
        global_opacity : float
            returns 1 if the colors array doesn't contain opacity otherwise -1.
            If colors array has 4 dimensions, it checks values of the fourth
            dimension. If the value is the same, then assign it to global_opacity.

        """
```

Clarkszw: Have applied `isinstance(color_aray, (list, tuple)`

fury/utils.py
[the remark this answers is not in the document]
Clarkszw: Removed.
The function that requires tuple as input, `coloc_check()`, is updated to work with `normalize_color()`

skoudoro (commented):
Hi @Clarkszw,

it looks much better. I need to play a bit more with this PR to make sure we do not fall in weird cases. 

Thank you for this and I will give you an update next week.

skoudoro:
also, can you solve the conflict in this PR ? thank you

Clarkszw:
I have no idea why it will happen..But the conflict has been fixed now:)

PR #774, opened 2023-04-03
WIP: Double arrow actor and a few utility functions

Sparsh-N:


skoudoro (changes requested):
Hi @Sparsh-N,

Can you explain what you are trying to do here?
Can you clean your tutorial and give more explanation? Looks at our tutorials.
Why are you commenting the material ?
Can you add some test also ?

Thank you in advance for your feedback

Sparsh-N:
Greetings @skoudoro , I'm trying to create a double arrow actor which can create an arrow within a given length parameter (building upon the original arrow actor). I'll create a tutorial, along with accompanying tests for the upcoming pull request. Some of the changes to material.py were commented out as those were separate changes for different functions (to set culling/opacity, but those would be shared in a separate PR). Thank you for informing and updating me with the necessary changes for a PR.

Sparsh-N:
i'll place a new PR for the dual arrow actor and utility functions separately.

PR #790, opened 2023-06-04
Fixing `TextBlock2D` justification issue

ganimtron-10:
Fixing the issue of alignment in `TextBlock2D` when using the `justification`  and `vertical_justification` property as we can see below.
image
image

This was happening because when we were updating the text justification it was with the reference of the current position of the UI element which is the bottom left corner of the element.

The fix proposed solve this issue as shown below.

image
image

### DIFF
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -785,6 +785,9 @@ class TextBlock2D(UI):
             self.background.resize(size)
         scene.add(self.background, self.actor)
 
+        self.justification = self.justification
+        self.vertical_justification = self.vertical_justification
+
     @property
     def message(self):
         """Get message from the text.
@@ -904,8 +907,10 @@ class TextBlock2D(UI):
             text_property.SetJustificationToLeft()
         elif justification == 'center':
             text_property.SetJustificationToCentered()
+            self.actor.SetPosition(self.position + [self.background.size[0]//2, 0])
         elif justification == 'right':
             text_property.SetJustificationToRight()
+            self.actor.SetPosition(self.position + [self.background.size[0], 0])
         else:
             msg = 'Text can only be justified left, right and center.'
             raise ValueError(msg)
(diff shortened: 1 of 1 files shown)


fury/ui/core.py
```diff
@@ -785,6 +785,9 @@ def _add_to_scene(self, scene):
             self.background.resize(size)
         scene.add(self.background, self.actor)
 
+        self.justification = self.justification
+        self.vertical_justification = self.vertical_justification
```
ganimtron-10, line 789: I am calculating the relative position of the text using the background `Rectangle2D` which is set once the UI is added to the scene. That is why I need to perform this justification after the element is added to the scene.

ganimtron-10:
Before creating the tests, I would like to know whether this approach of setting the justification in the `add_to_scene` method would be valid? or do I have to plan out some other approach to obtain this?

@skoudoro @guaje

fury/ui/core.py
skoudoro: it is strange to me. I can test it only on friday. What happens when you move the RectangleUI or when you resize it?

Does the justification stay in place ?

skoudoro (commented):
Thank you for this @ganimtron-10!

See below for my comment

skoudoro (commented):
Thank you for this @ganimtron-10!

See below for my comment

skoudoro:
> I would like to know whether this approach of setting the justification in the add_to_scene method would be valid?
 
It is not valid.  `add_to_scene` should have the minimum code possible. So, you should avoid to put code there. 
Also, as I said on my comment above, justification should stay in place if we move /resize an UI.

it would be great if you could fix this before friday. Thanks @ganimtron-10 !

skoudoro:
small note, please, take the habit to add labels and reviewers in your PRs

fury/ui/core.py
ganimtron-10: Yes the justification stay in place when we resize it!

ganimtron-10:
> It is not valid. `add_to_scene` should have the minimum code possible. So, you should avoid to put code there.
Also, as I said on my comment above, justification should stay in place if we move /resize an UI.

But as I mentioned I get the size of the background only when the UI element is added to the scene so I need to do the required changes after adding to the scene. You get a good idea on Friday after testing.

> small note, please, take the habit to add labels and reviewers in your PRs

I was unable to add @JoaoDell and @tvcastillod as reviewers and was waiting for their review so didn't added you to the same.
Will do this from next PR onwards.

skoudoro:
From the previous meeting, you can either update this PR or close it and create a new PR.

boundingbox need to be implemented. see our taskboard.

Thank you @ganimtron-10

ganimtron-10:
Thanks for the update @skoudoro. Will create a new PR.

PR #791, opened 2023-06-05
Ellipsoid actor implemented with SDF

tvcastillod:
Hi, this PR aims to create a new ellipsoid actor defined with SDF and raymarching. The goal is to make an implementation that allows displaying a large amount of data, with good visual quality, and without compromising performance. I'm still working on improvements and ways to test the implementation.


skoudoro:
Hi @guaje,

I see that you requested a review from all but I see that this PR is a draft with WIP tag. 

Is this PR ready for review?  What is the status ? @tvcastillod, can you answer also ?

tvcastillod:
> Is this PR ready for review? What is the status ? @tvcastillod, can you answer also ?

@skoudoro Not yet. I was waiting for today's meeting to know how to proceed with this PR, I received some comments that I will address soon, and when ready I will change the status and request the review again.

fury/tests/test_actors.py
```diff
@@ -734,7 +735,7 @@ def test_peak():
 
 
 # @pytest.mark.skipif(not have_dipy, reason="Requires DIPY")
-def test_tensor_slicer(interactive=False):
+def test_tensor_slicer(interactive=True):
```
skoudoro, line 738: back to `False`and remove the skipif

skoudoro (commented):
Hi @tvcastillod,

Here is an initial review. I will try a bit later your PR

fury/actor.py
```diff
@@ -3795,3 +3796,71 @@ def callback(
     shader_to_actor(sq_actor, 'fragment', impl_code=fs_impl_code, block='light')
 
     return sq_actor
+
+
+def tensor(
```
skoudoro, line 3801: we will need to talk with @guaje and @Garyfallidis concerning the naming. 

or, we need to find a way to explain what is the difference between `tensor` and `tensor_slicer`

fury/actors/tensor.py
```diff
@@ -0,0 +1,213 @@
...
+
+def tensor_ellipsoid(centers, axes, lengths, colors, scales, opacity):
+    """
+    Visualize one or many Tensor Ellipsoids with different features.
+
+    Parameters
+    ----------
+    centers : ndarray(N, 3)
+        Tensor ellipsoid positions
+    axes : ndarray (3, 3) or (N, 3, 3)
```
skoudoro, line 18: Does it have to be between [0-1] ? if yes, it would be good to use `np.clip` in case there is crazy values

fury/actors/tensor.py
```diff
@@ -0,0 +1,213 @@
...
+        Axes of the tensor ellipsoid
+    lengths : ndarray (3, ) or (N, 3)
```
skoudoro, line 20: same as above: Does it have to be between [0-1] ? if yes, it would be good to use `np.clip` in case there is crazy values.

fury/actors/tensor.py
```diff
@@ -0,0 +1,213 @@
...
+        Axes lengths
+    colors : ndarray (N,3) or (N, 4) or tuple (3,) or tuple (4,), optional
+        RGB or RGBA (for opacity) R, G, B and A should be at the range [0, 1]
+    scales : float or ndarray (N, ), optional
+        Tensor ellipsoid size, default(1)
+    opacity : float, optional
```
skoudoro, line 26: We might  need to create an issue for that. `Opacity` and `RGBA` are confusing. who has the priority? it seems that opacity have the priority.  We might need to remove `opacity` parameter.

skoudoro (commented):
Great job @tvcastillod, Looks good to me. 

I have few questions below to clarify or correct and then I think this PR will be ready to be merged.

fury/actors/tensor.py
tvcastillod: I do a normalization inside the shader so it is not necessary to put values within that range.

fury/actors/tensor.py
tvcastillod: I agree. For now, I will avoid the alpha channel, leave the color as ndarray (N,3) or tuple (3,) and keep the opacity.

fury/actor.py
tvcastillod: In the end, we decided to go with `ellipsoid` to keep the actor more generic. Additionally, I will work on a tutorial to show the use that can be made of this new actor in the visualization of diffusion tensor ellipsoids, in comparison with the `tensor_slicer` actor.

tvcastillod:
@skoudoro @guaje would you mind taking a last look to see if there is anything else I need to adjust?

skoudoro:
@JoaoDell, Can you look and review the shader part. 

@ganimtron-10, can you look and review the python part. 

Both, please, let @tvcastillod knows if you have any questions or if you see something.

After that, we can go ahead and merge this PR.

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
+import os
+
+import numpy as np
+
+from fury import actor
+from fury.shaders import (attribute_to_actor, import_fury_shader,
+                          shader_to_actor, compose_shader)
+
+
+def tensor_ellipsoid(centers, axes, lengths, colors, scales, opacity):
```
JoaoDell, line 10: Will it still be named `tensor_ellipsoid` after the renaming decision?

JoaoDell (commented):
Just finished reviewing mainly the shader part. The tests worked and sucessfully rendered the ellipsoids, so it is working. I just got some questions to better understand some choices, but everything seems fine 👍

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
+import os
+
+import numpy as np
+
+from fury import actor
+from fury.shaders import (attribute_to_actor, import_fury_shader,
+                          shader_to_actor, compose_shader)
```
ganimtron-10, line 7: `from fury.shaders import (attribute_to_actor, compose_shader, import_fury_shader,
                          shader_to_actor)` maybe

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
...
+    Parameters
+    ----------
+    centers : ndarray(N, 3)
+        Tensor ellipsoid positions
+    axes : ndarray (3, 3) or (N, 3, 3)
+        Axes of the tensor ellipsoid
+    lengths : ndarray (3, ) or (N, 3)
+        Axes lengths
+    colors : ndarray (N,3) or tuple (3,), optional
+        Default red color. R, G and B should be at the range [0, 1]
```
ganimtron-10, line 23: `at` -> `in`

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
...
+    scales : float or ndarray (N, ), optional
+        Tensor ellipsoid size, default(1)
```
ganimtron-10, line 25: `scales` is not set to default here!

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
...
+    opacity : float, optional
+        Takes values from 0 (fully transparent) to 1 (opaque).
```
ganimtron-10, line 27: `opacity` is not optional as default value is not provided.

ganimtron-10 (changes requested):
Hey @tvcastillod,
I had the few below comments. Please them out then it would be ready to go.

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
...
+        vertexMCVSOutput = vertexMC;
+        centerMCVSOutput = center;
+        scaleVSOutput = scale;
+        """
+
+    # Normalization
+    n_evals = "evalsVSOutput = evals/(max(evals.x, max(evals.y, evals.z)));"
+        
+    # Values constraint to avoid incorrect visualizations
+    evals = "evalsVSOutput = clamp(evalsVSOutput,0.05,1);"
```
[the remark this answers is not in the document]
tvcastillod: I chose it by trial and error, seeing what was the smallest possible value, to get a good visualization. This has to be with the eigenvalues that sometimes take values that are too small, and as I have to take the inverse of those values to do the scaling, the contribution to the result is too large, generating wrong visualizations. There should be ways to estimate the lowest reliable eigenvalue more precisely but it needs deeper research, so for now I leave it with that value. You can check more about this issue [here](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix#Practical_implications).

fury/actors/tensor.py
```diff
@@ -0,0 +1,214 @@
...
+                                     evalsVSOutput.z));
+                                     
+            /*
+            The approximation of distance is calculated by stretching the
+            space such that the ellipsoid becomes a sphere (multiplying by
+            the transformation matrix) and then computing the distance to
+            a sphere in that space (using the sphere SDF).
+            */
+            return sdSphere(tensorMatrix * (position - centerMCVSOutput),
+                scaleVSOutput*0.48) * scFactor;
```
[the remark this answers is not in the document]
tvcastillod: I set the sphere (that is transformed into an ellipsoid) with a radius of 0.5 (half of the box) for the ellipsoid to occupy the maximum space in the box. There is a small issue when drawing ellipsoids with one or two axes where the values that are too small, for example (1, .3, .00003), for some reason the ellipsoid goes a bit out of the box, as you can see in the image, so that's why set a value close to 0.5, to avoid an erroneous visualization.
image

fury/actors/tensor.py
tvcastillod: As the actor is named `ellipsoid` and I have to call the function in `actor.py`, I cannot use the same name for the function so I used `tensor_ellipsoid` instead.

tvcastillod:
@skoudoro I did all the requested changes, I think now it is ready.

skoudoro (approved):
Thank you for this work @tvcastillod.

Thank you for the review @JoaoDell, @ganimtron-10  and @guaje. It is good to have other pair of eye

merging

PR #803, opened 2023-06-15
UI: Adding Bounding Box & Fixing Alignment issue in TextBlock2D

ganimtron-10:
The `TextBlock2D` component in the UI system was encountering alignment and scaling issues. When resizing the `TextBlock2D`, the text scaling mode was set to "SetTextScaleModeToProp," which resulted in inconsistent alignment of the text actor and improper justification of the text with respect to the background rectangle. Additionally, the background rectangle's size was calculated only after it was added to the scene, which is not an ideal practice.

To address these issues, this PR introduces the following enhancements:
1. BoundingBox Property: A new property called "boundingbox" is implemented in the `TextBlock2D`component. This property calculates the text bounding box based on the content of the `TextBlock2D`. It ensures accurate alignment and justification of the text within the component.

2. Auto Font Scale: The scaling mode is now separated from the resizing action. A new property named "auto_font_scale" is introduced, which, when set to True, enables automatic font scaling according to the bounding box. This allows for consistent and proportional text sizing without manual intervention.

By implementing these changes, the `TextBlock2D` component will provide improved text alignment, justified text with the background rectangle, and a more intuitive font scaling experience.

### DIFF
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -757,7 +757,8 @@ class TextBlock2D(UI):
         if size is not None:
             self.resize(size)
         elif not self.dynamic_bbox:
-            raise ValueError("TextBlock size is required as it is not dynamic.")
+            # raise ValueError("TextBlock size is required as it is not dynamic.")
+            self.resize((0, 0))
 
     def _setup(self):
         self.actor = TextActor()
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -1057,28 +1057,28 @@ class TextBlock2D(UI):
         text_property = self.actor.GetTextProperty()
         updated_text_position = [0, 0]
 
-        if self.justification == 'left':
+        if self.justification.lower() == 'left':
             text_property.SetJustificationToLeft()
             updated_text_position[0] = self.boundingbox[0]
-        elif self.justification == 'center':
+        elif self.justification.lower() == 'center':
             text_property.SetJustificationToCentered()
             updated_text_position[0] = self.boundingbox[0] + \
                 (self.boundingbox[2]-self.boundingbox[0])//2
-        elif self.justification == 'right':
+        elif self.justification.lower() == 'right':
             text_property.SetJustificationToRight()
             updated_text_position[0] = self.boundingbox[2]
         else:
             msg = 'Text can only be justified left, right and center.'
             raise ValueError(msg)
 
-        if self.vertical_justification == 'bottom':
+        if self.vertical_justification.lower() == 'bottom':
             text_property.SetVerticalJustificationToBottom()
             updated_text_position[1] = self.boundingbox[1]
-        elif self.vertical_justification == 'middle':
+        elif self.vertical_justification.lower() == 'middle':
             text_property.SetVerticalJustificationToCentered()
             updated_text_position[1] = self.boundingbox[1] + \
                 (self.boundingbox[3]-self.boundingbox[1])//2
-        elif self.vertical_justification == 'top':
+        elif self.vertical_justification.lower() == 'top':
             text_property.SetVerticalJustificationToTop()
             updated_text_position[1] = self.boundingbox[3]
         else:
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -786,10 +786,6 @@ class TextBlock2D(UI):
             self.background.resize(size)
         scene.add(self.background, self.actor)
 
-        # self.background.resize(
-        #     (self.boundingbox[2]-self.boundingbox[0], self.boundingbox[3]-self.boundingbox[1]))
-        # self.background.position = [self.boundingbox[0], self.boundingbox[1]]
-
     @property
     def message(self):
         """Get message from the text.
@@ -1083,7 +1079,6 @@ class TextBlock2D(UI):
             self.have_bg = True
             self.background.set_visibility(True)
             self.background.color = color
-
 
     def update_bounding_box(self):
         lines = self.message.split("\n")
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -768,6 +768,8 @@ class TextBlock2D(UI):
         self.actor.SetTextScaleModeToProp()
         self.actor.SetPosition2(*size)
 
+        self.update_bounding_box()
+
     def _get_actors(self):
         """Get the actors composing this UI component."""
         return [self.actor] + self.background.actors
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -747,6 +747,9 @@ class TextBlock2D(UI):
         self.shadow = shadow
         self.vertical_justification = vertical_justification
         self.message = text
+        lines = self.message.split("\n")
+        self.boundingbox = [self.position[0], self.position[1],
+                            len(lines[0])*font_size, len(lines)*font_size]
 
     def _setup(self):
         self.actor = TextActor()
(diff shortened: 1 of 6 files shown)


skoudoro:
Hi @ganimtron-10, What is the status of this PR? 

Many tests are failing! I suppose that you might need to record again many tests because of this change.  Before doing that, let me know when I can check and test this PR

ganimtron-10:
Sure @skoudoro will let you know, Currently working on the tests itself.

JoaoDell:
Hey @ganimtron-10 is this ready for review or should we wait for you to ping us again? I see that you made some commits yesterday, so I don't know whether I should start or wait.

ganimtron-10:
Hey @JoaoDell you can go ahead!
I was just waiting for the tests to get completed so that I could make sure the changes doesn't hamper anything!
As the tests are complete now you can go ahead!

fury/ui/core.py
```diff
@@ -1078,6 +1046,71 @@ def background_color(self, color):
             self.background.set_visibility(True)
             self.background.color = color
 
+    def update_alignment(self):
+        """Update Text Alignment.
+
```
JoaoDell, line 1051: Extra space between `"""Update Text Alignment.` and `"""`

JoaoDell (commented):
Hey @ganimtron-10, I have just finished my first review here and, overall, it seems fine and working. I have ran the ui test and everything passed, so it does not seem to have any problem. I would just like to know if you recommend any example to be ran so I can check the new features and changes visually. Nice work!

ganimtron-10:
Hey @JoaoDell , You can try out this below code to try out the feature visually.

```python
from fury import ui, window
from fury.data import fetch_viz_icons

fetch_viz_icons()

c = ui.TextBox2D(width=12, height=3, justification='center') #left, right, center


show_manager = window.ShowManager(size=(500, 500),
                                  title="FURY textbox Example")


show_manager.scene.add(c)

show_manager.start()
```

fury/ui/core.py
```diff
@@ -730,23 +732,33 @@ def __init__(
             Adds text shadow.
         size : (int, int)
             Size (width, height) in pixels of the text bounding box.
+        auto_font_scale : bool, optional
+            Automatically scale font according to the text bounding box.
+        dynamic_bbox : bool, optional
+            Automatically reisize the bounding box according to the content.
```
tvcastillod, line 738: Automatically **_resize_** the bounding box

fury/ui/core.py
```diff
@@ -730,23 +732,33 @@ def __init__(
...
+        self._justification = justification
         self.bold = bold
         self.italic = italic
         self.shadow = shadow
-        self.vertical_justification = vertical_justification
+        self._vertical_justification = vertical_justification
+        self.auto_font_scale = auto_font_scale
+        if self.auto_font_scale:
+            self.actor.SetTextScaleModeToProp()
+        self.dynamic_bbox = dynamic_bbox
```
tvcastillod, line 754: Is it not necessary to do a setter for these parameters, or are they supposed to be defined only when creating the instance of the object?

fury/ui/core.py
tvcastillod, line 738: Shouldn't these new parameters also be specified as attributes of `TextBlock2D`?

fury/ui/core.py
```diff
@@ -1078,6 +1046,71 @@ def background_color(self, color):
...
+        ----------
+        size : (int, int) or None
+            If None, calculates bounding box.
+            Otherwise, uses the given size.
+
+        """
+        if size is None:
+            size = self.cal_size_from_message()
+
+        self.boundingbox = [self.position[0], self.position[1],
```
tvcastillod, line 1104: I'm getting this comment in my editor, so I wonder if this is a good way to handle `boundingbox` variable, as it is not an attribute from the class itself but you are using _self._ to define it.

image

tvcastillod (commented):
Hi @ganimtron-10, I think it is looking good, you have made great progress. I left a few comments and questions below for you to check them. And I have another question related to the two enhancements mentioned.

Between `auto_font_scale` and `dynamic_bbox`, which one has the priority (what happens when I set both true)? because both seem to be opposite as with_auto_font_scale_, the bounding box is the one which set the size of the text, but with _dynamic_bbox_, it is the text size that sets the bounding box size. Did I understand correctly?

ganimtron-10:
> Between `auto_font_scale` and `dynamic_bbox`, which one has the priority (what happens when I set both true)? because both seem to be opposite as with_auto_font_scale_, the bounding box is the one which set the size of the text, but with _dynamic_bbox_, it is the text size that sets the bounding box size. Did I understand correctly?

Both of the parameter can co exist! 
See `auto_font_scale` parameter helps to scale the font according to the background boundingbox, whereas `dynamic_bbox` helps to dynamically modify the boundingbox according to the text not according to the text font. So when both is applied the background box is dynamic and the text font is scaled accordingly!

fury/ui/core.py
ganimtron-10: Maybe yes!
Will update it accordingly.

ganimtron-10:
Thanks @JoaoDell @tvcastillod for the review!

skoudoro:
Hi @ganimtron-10, 

Tests are still failing, do you plan to look at them ? please, look at tabui

fury/ui/core.py
```diff
@@ -685,6 +685,10 @@ class TextBlock2D(UI):
         Adds text shadow.
     size : (int, int)
         Size (width, height) in pixels of the text bounding box.
+    auto_font_scale : bool
+        Automatically scale font according to the text bounding box.
+    dynamic_bbox : bool
+        Automatically resize the bounding box according to the content.
```
skoudoro, line 691: if you add them here, I agree with @tvcastillod, we need setter/getter functions.

fury/ui/core.py
```diff
@@ -1078,6 +1051,71 @@ def background_color(self, color):
             self.background.set_visibility(True)
             self.background.color = color
 
+    def update_alignment(self):
+        """Update Text Alignment.
+        """
+        text_property = self.actor.GetTextProperty()
+        updated_text_position = [0, 0]
+
+        if self.justification == 'left':
```
skoudoro, line 1060: I will use  `self.justification.lower()` everywhere, just in case....

fury/ui/core.py
```diff
@@ -1078,6 +1051,71 @@ def background_color(self, color):
...
+            text_property.SetJustificationToLeft()
+            updated_text_position[0] = self.boundingbox[0]
+        elif self.justification == 'center':
```
skoudoro, line 1063: same as above

skoudoro (commented):
Overall, it looks good.

The code is better like that, it is much simpler. I still need to play a bit more. 
see below few comments that can be fixed quickly

Thanks @ganimtron-10

ganimtron-10:
Hey @skoudoro ,

> Tests are still failing, do you plan to look at them ? please, look at tabui

I checked those last time and it worked fine! Maybe I saw outdated version, My bad.
Also it is a bit weird to me that in the tabui test, we are equating `title_font_size` to 12 where as the default size is 18. Am I missing some thing? I checked out the code are we are not manually setting the font_size too.

ganimtron-10:
I was also wondering do we need `wrap_overflow` helper function, as the functionality is now directly available in the `TextBlock2D` as `dynamic_bbox`. I searched the whole codebase and we don't directly use it anywhere.

fury/ui/core.py
ganimtron-10: I will remove this for now and will create a subsequent PR for adding these attributes.

ganimtron-10:
Hey @skoudoro ,
I have made most of the necessary changes and the test are also working fine!

skoudoro:
Great!  let me play a bit with this and we can move forward

skoudoro (approved):
Nice work @ganimtron-10! Let's move forward. merging

PR #424, opened 2021-04-26
Allows fury to deal with Uniforms in shaders and change them dynamically

devmessias:
An uniform variable should be chosen when we have any kind
of information which act as constants, at least during the  draw call.

This commit implements the following behaviors:
1 - avoids creating an observable-vtk for each uniform variable
2 - allows the update of an uniform variable dynamically

Here it's a simple example

```python

markerUniforms = [
        Uniform(name='edgeColor', type='3f', value=edgeColor),
        Uniform(name='edgeWidth', type='f', value=edgeWidth),
        Uniform(name='edgeOpacity', type='f', value=edgeOpacity),
        Uniform(name='markerOpacity', type='f', value=markerOpacity),
    ]
CustomUniforms = Uniforms(markerUniforms)
observerId = add_shader_callback(
            sq_actor, CustomUniforms)
CustomUniforms.observerId = observerId
sq_actor.CustomUniforms = CustomUniforms

sq_actor.CustomUniforms.edgeWidth = 0.5

```

PR #804, opened 2023-06-15
KDE Rendering Experimental Program 

JoaoDell:
[UPDATED]
Hello everyone, this PR contains an experimental implementation of Kernel Density Estimation calculations for a set of points. This uses some of fury's existing APIs and some straight vtk ones. This program may be a good example of shader usage and how to apply post-processing effects inside FURY. The result from running that should be:

image

Below, what is the idea here:

1. Render billboards with KDE calculations of points to an offscreen window.
2. Capture the offscreen window with a texture.
3. Render a second billboard, this time to the screen, with the texture of the prior rendered window, having a post processing effect applied to it (color map application).

In more details, what is done:

1. The offscreen and normal screen managers are created. The offscreen window is set to be for offscreen rendering with `SetOffscreenRendering(True)`.
2. The parameters of the rendering are created: Random points are initialized, the sigma for the KDE is set, and the scale of the billboards is defined. The points are passed to the billboard actor as its centers.
3. `shaders.shader_apply_effects()` is used to tell OpenGL additive blending is desired for this code.
4. The billboards are rendered to the offscreen window.
5. After this render, the second billboard, the one that will receive the texture from the prior rendering, is created.
6. The colormap is chosen from `matplotlib.colormaps` and passed to the billboard as a texture named `"colormapTexture"` with the function `colormap_to_texture()`
7. The window is captured as a texture and passed to the second billboard with `window_to_texture()`
8. The billboard is then finally rendered to the screen.

Some minor details that may be useful for a better understanding of it:
* The billboard actor got a set o shaders for the [KDE calculation](https://en.wikipedia.org/wiki/Kernel_density_estimation). In its declaration, a function that calculates the non-normalized (yet, I intend to implement a normalized version) KDE as: $KDE(x, \sigma) = exp(-\frac{1}{2}(\frac{x}{\sigma})^2)$.
* For all of that to work, I have created some functions to ease the work, functions that may be even used later in other FURY applications, so I wonder it would be a good idea to make a separate small PR for them.


skoudoro:
thank you for your feedback. we will look at all this links.

JoaoDell:
After a pair with @devmessias, we made some progress on the FBO setup issue. Now, the FBO is sucessfully generated, and it seems ready to be drawn to. Apparently, the issue was that the context needed to have the window interactor initialized with `ShowManager.initialize()` before the it was set to the FBO. I will prepare a better explaination and investigate further on why is that soon.

docs/experimental/framebuffer_example.py
```diff
@@ -0,0 +1,246 @@
...
+                    
+billboard_frag_impl ="""                   
+                        vec3 texture = texture(screenTexture, out_tex);
+                        vec3 color = vec3(1.0, 0.0, 0.0);
+                        
+                        fragOutput0 = vec4(color, 1.0);"""
+
+
+frag_decl = """varying vec2 out_tex; 
+               uniform vec2 res0;
```
devmessias, line 100: @JoaoDell  you should declare `in vec3 normalizedVertexMCVSOutput;` here

```python
frag_decl = """varying vec2 out_tex;
               in vec3 normalizedVertexMCVSOutput;
               uniform vec2 res0;
               uniform vec3 point0;"""
```

docs/experimental/framebuffer_example.py
```diff
@@ -0,0 +1,246 @@
...
+
+FBO.SaveCurrentBindingsAndBuffers()
+FBO.Bind()
+FBO.ActivateBuffer(0)
+
+print("FBO Start:", FBO.Start(width, height))
+
+# Render every point existing inside loop
+for i in range(n_points):
+    shader_custom_uniforms(billboard, "fragment").SetUniform3f("point0", points[i, :])
```
devmessias, line 225: I don't believe this action is feasible as it would attempt to create new instances of 'point0' and 'res0' variables. This leads to an error. The error message I received :
```
2023-06-26 19:22:22.899 (   0.328s) [        B8DAB000]   vtkShaderProgram.cxx:453    ERR| vtkShaderProgram (0x558732a3cb20): 0:109(15): error: `res0' redeclared
0:110(15): error: `point0' redeclared
```

We have a old PR that deals with that 
https://github.com/fury-gl/fury/pull/424 maybe we need to look into this again @skoudoro  and @filipinascimento

JoaoDell: Interesting, I didn't know how VTK handled uniforms so I thought the `SetUniform` methods only passed uniform values to the shader. That seems like something worth checking out indeed 👍

docs/experimental/framebuffer_example.py
JoaoDell: It seems like the problem was somewhere else, but i have just updated the PR with some changes that don't conflict with the shaders

docs/experimental/framebuffer_example.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+frag_2 = """
+
+varying vec2 out_tex;
+                    
+uniform sampler2D screenTexture;
+
+void main(){ 
+    vec4 texture = texture(screenTexture, out_tex);
```
devmessias, line 81: texture should be 2D vec right? What the meaning of out_text here?
try this

```
void main(){
    //vec2 texture = texture(screenTexture, out_tex);
   // gl_FragColor = texture;
  vec2 coords = gl_TexCoord[0].xy;
  vec3 normalColor = texture2D(screenTexture, coords).rgb;
  gl_FragColor = vec4(normalColor.r, normalColor.g, normalColor.b, 1);
    //gl_FragColor = vec4(0.0, 0.0, 1.0, 1.0);
}
```

JoaoDell: I was passing out_tex as texture coordinates, but I will try what you recommended

JoaoDell:
Hey everyone, I just updated and completely refactored this PR as, as well as it is base program, as it went on a slight different direction in comparison to its original intetion. Everything seems to be working now, my only question is if it would be better if I closed this PR and opened a new one with a new branch containing only the recent commits, as the older ones have been completely replaced, or should I rebase it, I don't know. What should I do? 

Anyway, this version is ready for review 👍

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,356 @@
...
+    texture.SetBorderColor(r, g, b, a)
+    texture.SetWrap(wrap_mode_dic[wrap_mode])
+    texture.SetInterpolate(interpolate)
+    texture.MipmapOn()
+    texture.SetBlendingMode(blending_mode_dic[blending_mode])
+
+    target_actor.GetProperty().SetTexture(texture_name, texture)
+
+
+def texture_to_actor(
```
tvcastillod, line 82: You are not using this function

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,356 @@
...
+        return actor.GetShaderProperty().GetVertexCustomUniforms()
+    elif shader_type == "fragment":
+        return actor.GetShaderProperty().GetFragmentCustomUniforms()
+    elif shader_type == "geometry":
+        return actor.GetShaderProperty().GetGeometryCustomUniforms()
+    else:
+        raise ValueError("Shader type unknown.")
+
+
+def normalize(array : np.array, min : float = 0.0, max : float = 1.0, axis : int = 0):
```
tvcastillod, line 211: `axis` parameter is not used

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,356 @@
...
+"""
+
+kde_impl = """
+float current_kde = kde(normalizedVertexMCVSOutput, sigma);
+color = vec3(current_kde);
+fragOutput0 = vec4(color, 1.0);
+"""
+
+tex_dec = """
+const float normalizingFactor = 1.0; // Parameter to be better defined later
```
tvcastillod, line 249: Is it necessary to define it now? can you explain what is this parameter for?
I think you might delete `normalizingFactor` as is not doing relevant changes right now. Maybe leave a comment as a reference for future changes.

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,356 @@
...
+if(intensity<=0.0){
+    discard;
+}else{
+    color = color_mapping(intensity, normalizingFactor).rgb;
+    fragOutput0 = vec4(color, 1.0);
+}
+"""
+
+
+fs_dec = compose_shader([kde_dec])
```
tvcastillod, line 270: This is not necessary since it is just one string, you can pass to the billboard actor `kde_dec` and `kde_impl` directly.

tvcastillod (commented):
Hi @JoaoDell,
I need to take a deeper look at your work, but I've left some comments below. Overall is looking good.

docs/experimental/viz_points_kde.py
JoaoDell: Indeed, I forgot to take it out, thanks Tania

docs/experimental/viz_points_kde.py
JoaoDell: I don't think so as well, I left it as a reminder here but I guess this will be better resolved on the API, so I will take it out

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+window_to_texture(
+    off_manager.window,
+    "screenTexture",
+    textured_billboard,
+    blending_mode="Interpolate")
+
+manager.scene.add(textured_billboard)
+
```
ganimtron-10, line 276: If this is a demo or example we would need the usual

```python
if interactive:
   manager.start()

window.record(...)

```

ganimtron-10 (changes requested):
Hey @JoaoDell ,
Giving a first look I found out these below points. Some of them would require more discussion from mentors to clarify few things.
If you think any of these comments doesn't align according to your approach while implementing it feel free to ping me or mentors at the comment.

Thanks in advance!

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,277 @@
...
+    wrap_mode_dic = {"ClampToEdge" : Texture.ClampToEdge,
+                     "Repeat" : Texture.Repeat,
+                     "MirroredRepeat" : Texture.MirroredRepeat,
+                     "ClampToBorder" : Texture.ClampToBorder}
+
+    blending_mode_dic = {"None" : 0, "Replace" : 1,
+                         "Modulate" : 2, "Add" : 3,
+                         "AddSigned" : 4, "Interpolate" : 5,
+                         "Subtract" : 6}
```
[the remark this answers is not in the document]
JoaoDell: I set those like that because it is how their names are defined inside VTK, however this suggestion makes sense as a way to prevent this kind of problems a user may have. I think it is an easy and minor thing to be modified, so for me that seems alright to do 👍

docs/experimental/viz_points_kde.py
JoaoDell: This is indeed a demo of this experimental feature I am about to implement as an API, so that is indeed needed, thank you 👍

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,277 @@
...
+    r, g, b, a = border_color
+
+    windowToImageFilter = WindowToImageFilter()
+    windowToImageFilter.SetInput(window)
+
+    windowToImageFilter.Update()
+
+    texture = Texture()
+    texture.SetInputConnection(windowToImageFilter.GetOutputPort())
+    texture.SetBorderColor(r, g, b, a)
```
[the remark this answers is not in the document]
JoaoDell: Nice, I didn't know about that, thanks

JoaoDell:
@ganimtron-10 I have fixed and updated most of your recommendations. Indeed, `typedhint` isn't a FURY standard, but I will wait Filipi and Serge comments on that to change anything. Thanks for the comprehensive review!

JoaoDell:
Also, I just realised there is a native colormap generator in `fury.colormp` named `create_colormap()` so I just updated it with that.

skoudoro (commented):
Hi @JoaoDell,

Thank you for this. This experimental is much more advanced than the other experimental.

I would recommend to start adding some text and explanation and then, we can move it to the examples folder. Maybe we can create a new section named "advanced Topic". 

To discuss....

docs/experimental/viz_points_kde.py
```diff
@@ -0,0 +1,277 @@
...
+        texture_name : str,
+        target_actor : actor.Actor,
+        blending_mode : str = "None",
+        wrap_mode : str = "ClampToBorder",
+        border_color : tuple = (
+            0.0,
+            0.0,
+            0.0,
+            1.0),
+        interpolate : bool = True):
```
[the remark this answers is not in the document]
JoaoDell: Okay, so should I leave it like that and keep using `typedhint` for my future commits or should I stay on today's standards and not use it for now until the make it standard decision for FURY is made @skoudoro?

JoaoDell:
Okay @skoudoro, I will be looking into the existing examples to understand how I should explain and detail this one so we can try using it as an example then, thanks for the opinion!

PR #806, opened 2023-06-21
Week 3 Blogpost

JoaoDell:
Hello everyone, sorry for the delay on my blogpost, I got seriously ill and needed some rest, but now I am better and could finish it, so here it is!
image


tvcastillod (commented):
Hi @JoaoDell, nice and detailed blogpost. It's generating correctly so it looks good to me.

skoudoro (approved):
all good! merging!

PR #818, opened 2023-07-10
[WIP] Tutorial on using ellipsoid actor to visualize tensor ellipsoids for DTI

tvcastillod:
This PR consists of a tutorial that is intended to show two ways of displaying diffusion tensor ellipsoids for DTI visualization. The first is using the basic `tensor_slicer` actor which allows us to slice many tensors as ellipsoids, and the second with the generic `ellipsoid` actor I'm currently working on (fury-gl/fury#791), which can be used to display different amounts of ellipsoids. The idea is to show the use that can be made of the ellipsoid actor in the visualization of diffusion tensor ellipsoids, compared to the tensor_slicer actor, contrasting visual quality and the amount of data that can be rendered.

I still working on this, it is not ready for review yet.

### DIFF
--- a/docs/examples/viz_dt_ellipsoids.py
+++ b/docs/examples/viz_dt_ellipsoids.py
@@ -14,7 +14,172 @@ import numpy as np
 
 from dipy.io.image import load_nifti
 
-from fury.actor import _fa, _color_fa
+from fury import window, actor
+#from fury.actor import _fa, _color_fa
 from fury.data import fetch_viz_dmri, read_viz_dmri
+from fury.primitive import prim_sphere
 
+# Now, we fetch and load the data needed to display the Diffusion Tensor
+# Images.
 
+fetch_viz_dmri()
+
+# The tensor ellipsoids are expressed as eigen values and eigen vectors
+# which are the decomposition of the diffusion tensor that describes the water
+# diffusion within a voxel.
+
+evecs, _ = load_nifti(read_viz_dmri('slice_evecs.nii.gz'))
+evals, _ = load_nifti(read_viz_dmri('slice_evals.nii.gz'))
+roi_evecs, _ = load_nifti(read_viz_dmri('roi_evecs.nii.gz'))
+roi_evals, _ = load_nifti(read_viz_dmri('roi_evals.nii.gz'))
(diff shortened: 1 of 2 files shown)


ganimtron-10:
Hey @tvcastillod,
I tried out the tutorials and they are fabulous, the visualization looks beautiful.

docs/examples/viz_dt_ellipsoids.py
```diff
@@ -0,0 +1,227 @@
...
+
+###############################################################################
+# To render the same tensor slice using a different sphere we redefine the
+# vertices and faces of the sphere using prim_sphere with other sphere
+# specification, let's say 'repulsion200' and 'repulsion724'.
+#
+# If we zoom in at the scene to see with detail the tensor ellipsoids displayed
+# with the different spheres, we get the following results.
+#
+
```
tvcastillod, line 99: I would like to display an image like this

image

Is it possible just to upload this image or do I have to create the scene in that way? Also, how can I zoom in the scene to get an image like this?

image

I tried `scene.zoom(2)` but it is not working, what is the correct way to do it?

docs/examples/viz_dt_ellipsoids.py
```diff
@@ -0,0 +1,227 @@
...
+###############################################################################
+# Next, we set up a new scene to add and visualize the tensor ellipsoids
+# created.
+
+scene = window.Scene()
+scene.add(tensor_slice)
+
+# Enables/disables interactive visualization
+interactive = True
```
skoudoro, line 83: Set it to False

docs/examples/viz_dt_ellipsoids.py
skoudoro: I agree, please, use `ShowManager`

skoudoro: put the background white to increase contrast also

skoudoro (commented):
Nice Tutorial @tvcastillod.

- Please use `ShowManager` everywhere.
- do not forget to add your tutorial in https://github.com/fury-gl/fury/blob/master/docs/examples/_valid_examples.toml

skoudoro (approved):
Excellent Tutorial ! Thank you @tvcastillod, this is great and explain well the difference.

I am going ahead and merge this PR. Thank you

PR #819, opened 2023-07-10
GSoC: Week 6 Blogpost

JoaoDell:
Hey everyone, here is my week 6 blogpost. One important thing to point out is that I've included the local link for my week 5 blogpost, but it has not been merged yet, so the link is not currently working, but everything else seems to be.
image


ganimtron-10 (changes requested):
Hey @JoaoDell,
Nice Detailed Blogpost.
Please see my below comment.

skoudoro (approved):
LGTM, merging

PR #826, opened 2023-07-19
First Stage of the KDE Rendering API

JoaoDell:
Hello, this is a draft PR to start discussing the details of the KDE rendering API for FURY. Talking with @devmessias, some points arised regarding how this API should be done:

The way this kind of render work is different from other actors because it deals with **post-processing effects**. This means this render needs first to render the scene into a renderer, capture that rendered scene and pass it as a texture to another actor inside another scene assigned to another window, so the effect can be applied. This involves offscreen renderers and an implicit callback function declared when that actor is created to do that job of calling the offscreen rendering, so it is kinda more complex than a simple actor and a standard pipeline. Having that in mind, we thought in two options:

1. Create a simple actor that deals with one or more KDE renders.
2. Create a class to manage KDE actors, or maybe in a more general way, a class that manages post-processing effects. 

The first one may be a more simple and user-friendly approach at first, but may present some issues regarding optimization and needs to be carefully thought. For example, it needs to be decided whether the user will pass an offscreen manager together with the onscreen one, or if this will be created inside the function. Leaving that to the user may difficult the actor usage as several pieces would need to be put together by them to work, but creating the offscreen manager inside the function may slow the rendering if the user decides to create multiple KDE actors and not multiple KDE renders inside only one KDE actor. 

The second one considers the possibility of creation of another future feature in FURY, which is the post-processing feature, as the implementation of capturing the screen to a texture and pass it to be rendered in a billboard can have more general usages than only a KDE render. It also may deal with the problems pointed above internally, as a post-processing manager may reduce the probability of several managers being created, resulting in faster processing time. However, it needs to be carefully thought as well, as it needs to be simple and straightforward, existing only to avoid conflicts and diminish user misusage, even though the latter can never be extinguished, I think.

Having that in mind, we decided to take the second path, as it seems to solve issues that may arise. However, it would be important to have other opinions on that, so feel free to use this PR as a place to do that 👍 

[UPDATE]

The first draft of the API is done and ready for review. Below, a brief explanation of how this works:

The main programs here are:
* `effect_manager.py`
* `viz_kde_render.py`

Those two are where the implementation of the API is done and tested, so you should be looking into that when reviewing then. Also, there is the testing script for the API:
* `test_effect_manager.py`

And the test for the kde actor under:
* `test_actors.py` (I have put it down the end)

There are also new shaders to be used:
* `*_distribution.glsl`
* `color_mapping.glsl`

The first ones are different kernels for the kde calculations, and the last one is the function to map intensities to a color map passed by the user. I have also altered some files:
* `shaders/__init__.py`
* `shaders/base.py`
* `_valid_examples.toml`
* `window.py`

With some minor changes necessary for this API to work. In case anyone got any questions, feel free to tell me down below so we can discuss about all of this. I got some doubts regarding some implementation choices, like, for example, if the "effects" name won't cause any confusion as FURY already has `shader_apply_effects()` function, so I would like to hear from that as well.

exponential

An important thing to say is that there are some important details missing from this API:
* This is **not** a KDE rendering, exactly. It still lacks the division by the number of points rendered so the density can be calculated, however this is a more complicated detail as it steps on the problem of not having enough precision as it lacks a float framebuffer, being replaced by an unsigned integer one. To understand that, you can check my week 9 blogpost on that on #831. Bruno and Filipi oriented me to leave that for a later PR.
* Talking with Filipi, he hinted me it would be good to have a slider for the intensities. I tried that, however it seems to be crashing the whole program, but I didn't understand exactly why yet, so we decided to leave that for later as well.


fury/shaders/utils/cosine_distribution.glsl
```diff
@@ -0,0 +1,12 @@
...
+float kde(vec3 point, float sigma){
+    float norm = (PI/(4.0*sigma)); 
+    return norm*cos(PI*length(point)/(2*sigma))*int(length(point) < sigma);
+}
+
+
+// This requires a center to be passed
+// float kde(vec3 point, vec3 center, float sigma){
+//     return cos(PI*length(center - point)/(2*sigma))*int(length(center - point) < sigma);
+// }
```
devmessias, line 12: Remove this

fury/shaders/utils/epanechnikov_distribution.glsl
```diff
@@ -0,0 +1,11 @@
...
+float kde(vec3 point, float sigma){
+    float norm = (3.0/(4.0*sigma)); 
+    return norm*(1.0 - (length(point)*length(point))/(sigma*sigma));
+}
+
+
+// This requires a center to be passed
+// float kde(vec3 point, vec3 center, float sigma){
+//     return 1.0 - (length(center - point)*length(center - point))/(sigma*sigma);
+// }
```
devmessias, line 11: Remove this

fury/shaders/utils/exponential_distribution.glsl
```diff
@@ -0,0 +1,11 @@
...
+#define E 2.7182818
+float kde(vec3 point, float sigma){
+    return exp(-1.0*length(point)/sigma);
+}
+
+
+// This requires a center to be passed
+// float kde(vec3 point, vec3 center, float sigma){
+//     return exp(-1.0*length(center - point)/sigma);
+// }
```
devmessias, line 11: Remove this

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,431 @@
...
+            fragOutput0 = vec4(final_color.rgb, u_opacity*final_color.a);
+        }
+        """
+
+        fs_dec = compose_shader([varying_dec, kde_dec])
+
+        # Scales parameter will be defined by the empirical rule:
+        # 1*sima radius = 68.27% of data inside the curve
+        # 2*sigma radius = 95.45% of data inside the curve
+        # 3*sigma radius = 99.73% of data inside the curve
```
devmessias, line 298: """
Scales parameter will be defined by the empirical rule:
1*sima radius = 68.27% of data inside the curve
 2*sigma radius = 95.45% of data inside the curve
3*sigma radius = 99.73% of data inside the curve
"""

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,431 @@
...
+from fury.actor import Actor, billboard
+from fury.colormap import create_colormap
+from fury.io import load_image
+from fury.lib import Texture, WindowToImageFilter
+from fury.shaders import (attribute_to_actor,
+                          compose_shader,
+                          import_fury_shader,
+                          shader_apply_effects,
+                          shader_custom_uniforms)
+from fury.ui import LineSlider2D, TextBlock2D, Panel2D
```
devmessias, line 12: Because this is the first PR for the effect manager, avoid adding any additional features to it. If you implement everything now, it could make creating tests and debugging your code more difficult.

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,431 @@
...
+        wrap_mode : str = "ClampToBorder",
+        border_color : tuple = (
+            0.0,
+            0.0,
+            0.0,
+            1.0),
+        interpolate : bool = True,
+        d_type : str = "rgb"):
+    """Capture a rendered window and pass it as a texture to the given actor.
+    Parameters
```
devmessias, line 45: Change 

```
  """Capture a rendered window and pass it as a texture to the given actor.
    Parameters
```
to

```
  """Capture a rendered window and pass it as a texture to the given actor.
   
    Parameters
```
Docstrings follow a set of well-defined patterns, allowing for automatic documentation generation through tools like Sphinx.

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,431 @@
...
+        blending_mode : str = "None",
+        wrap_mode : str = "ClampToBorder",
+        border_color : tuple = (
+            0.0,
+            0.0,
+            0.0,
+            1.0),
+        interpolate : bool = True):
+    """Pass an imported texture to an actor.
+    Parameters
```
devmessias, line 109: same as before, break a line between the first docstring line and the Parameters

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,431 @@
...
+
+
+def colormap_to_texture(
+        colormap : np.array,
+        texture_name : str,
+        target_actor : Actor,
+        interpolate : bool = True):
+    """Convert a colormap to a texture and pass it to an actor.
+    Parameters
+    ----------
```
devmessias, line 159: same as before, break a line between the first docstring line and the Parameters

JoaoDell:
Hello everyone, after some weeks, I think this PR is ready for its first review. I have updated its description with some information that may help you with that.

JoaoDell:
Also, I just noticed I may need to replace `sigma` name by `bandwidth`, to be accurate with the KDE literature:[https://en.wikipedia.org/wiki/Kernel_density_estimation](https://en.wikipedia.org/wiki/Kernel_density_estimation)

fury/shaders/base.py
```diff
@@ -412,3 +412,22 @@ def attribute_to_actor(actor, arr, attr_name, deep=True):
...
+    ----------
+    actor : actor.Actor
+          Actor which the uniform values will be passed to.
+    shader_type : str
+          Shader type of the uniform values to be passed. It can be:
+          * "vertex"
+          * "fragment"
+          * "geometry"
+    """
+    SHADER_FUNCTIONS = {"vertex" : actor.GetShaderProperty().GetVertexCustomUniforms(),
```
devmessias, line 429:
```
    SHADER_FUNCTIONS = {"vertex" : actor.GetShaderProperty().GetVertexCustomUniforms(), 
                        "fragment" : actor.GetShaderProperty().GetFragmentCustomUniforms(),
                        "geometry" : actor.GetShaderProperty().GetGeometryCustomUniforms()}
    if shader_type not in SHADER_FUNCTIONS:
        raise ValueError('Shader type should be: ...')
    return SHADER_FUNCTIONS[shader_type]

File: fury/actors/effect_manager.py [hunk 10]
```diff
@@ -0,0 +1,401 @@
...
+        attribute_to_actor(bill, np.repeat(scales, 4), "in_scale")
+
+        if self._n_active_effects > 0:
+            self.off_manager.scene.GetActors().GetLastActor().SetVisibility(False)
+        self.off_manager.scene.add(bill)
+
+        bill_bounds = bill.GetBounds()
+        max_sigma = 2*4.0*np.max(sigmas)
+
+        actor_scales = np.array([[bill_bounds[1] - bill_bounds[0] +
```
tvcastillod, line 329: May I know how are you calculating the correct scaling?

fury/tests/test_actors.py
```diff
@@ -1819,3 +1820,84 @@ def test_actors_primitives_count():
         primitives_count = test_case[2]
         act = act_func(**args)
         npt.assert_equal(primitives_count_from_actor(act), primitives_count)
+
+def test_kde_actor(interactive = True):
```
tvcastillod, line 1824: set `interactive` to `False`

tvcastillod, line 1824: Not sure if this should be tested here, as you don't have a `kde` actor definition on `actor.py`, maybe take a look at `actors/peak.py` and `peak` in `actor.py`, to see how you could adjust that.

tvcastillod (commented):
Hi @JoaoDell,
I know you have to do some refactoring, but I left a few comments on the current implementation for you to take into account for the changes you are working on.

fury/window.py
```diff
@@ -740,8 +740,9 @@ def play_events_from_file(self, filename):
 
     def add_window_callback(self, win_callback, event=Command.ModifiedEvent):
         """Add window callbacks."""
-        self.window.AddObserver(event, win_callback)
+        window_id = self.window.AddObserver(event, win_callback)
         self.window.Render()
+        return window_id
```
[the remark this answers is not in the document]
JoaoDell: This change was made to assure anyone working with the window callback or the interactor callback could have access to the added callback id, in case they need to remove or change it later. I use it, for example, to remove the callbacks added by an effect inside the `remove_effect()` method

fury/actors/effect_manager.py
JoaoDell: Sure. The actor_scales is a dynamic variable that calculates the general scale of the `textured_billboard` depending on some parameters:
* `bill_bounds` : The bounds of the offscreen rendered billboards
* `center_of_mass` : The average of all of the points positions. This accounts for the possible displacement of the points
* `max_sigma` : The maximum value that includes all of the non-zero values of the KDE calculation. It relates to the empirical rule of the gaussian disstribution, that says that $3\sigma$ includes 99.73% of the values, so $4\sigma$ should include the remaining ones.

All of them are them summed together to calculate a possible size so the billboard would not cut the renderization. This is like this for now, but I have identified that I may need to change the scaling to a way that it occupies the whole screen, as it is affecting the real size of the points, so expect some change regarding that.

fury/tests/test_actors.py
JoaoDell: Ok, I will be looking into that :+1

skoudoro:
@ganimtron-10, can you look and review this PR?

docs/examples/viz_kde_render.py
```diff
@@ -0,0 +1,112 @@
...
+
+n_points = 1000
+points = np.random.rand(n_points, 3)
+points = normalize(points, -5, 5)
+offset = np.array([0.0, 0.0, 0.0])
+points = points + np.tile(offset, points.shape[0]).reshape(points.shape)
+
+###################################################################
+# For this KDE render, we will use a set of random sigmas as well, 
+# generated with ``numpy.random.rand`` as well, which are also
```
ganimtron-10, line 78: This sentence sounds a bit weird to me `as well` is repeated twice.

fury/actors/tests/test_effect_manager.py
```diff
@@ -0,0 +1,325 @@
...
+
+    npt.assert_equal(1, n_textures)
+    npt.assert_equal(WRAP_MODE_DIC[wrap_mode.lower()], texture.GetWrap())
+    npt.assert_equal(
+        BLENDING_MODE_DIC[blending_mode.lower()], texture.GetBlendingMode())
+    npt.assert_array_almost_equal(list(border_color), texture.GetBorderColor())
+    npt.assert_equal(interpolate, texture.GetInterpolate())
+    npt.assert_equal(True, texture.GetMipmap())
+
+    record(on_manager.scene, out_path="test_window.png", size=(width, height))
```
ganimtron-10, line 151: Do you need this screen shot? As I see you are not testing anything on this it is better to remove it!

fury/actors/tests/test_effect_manager.py
```diff
@@ -0,0 +1,325 @@
...
+    scene = Scene()
+    scene.set_camera(position=(-24, 20, -40),
+                     focal_point=(0.0,
+                                  0.0,
+                                  0.0),
+                     view_up=(0.0, 0.0, 0.0))
+
+    manager = ShowManager(
+        scene,
+        "demo",
```
ganimtron-10, line 275: some specific window name instead of `demo`

fury/actors/tests/test_effect_manager.py
```diff
@@ -0,0 +1,325 @@
...
+    scene = window.Scene()
+    scene.set_camera(position=(-24, 20, -40),
+                     focal_point=(0.0,
+                                  0.0,
+                                  0.0),
+                     view_up=(0.0, 0.0, 0.0))
+
+    manager = window.ShowManager(
+        scene,
+        "demo",
```
ganimtron-10, line 302: same as above

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,402 @@
...
+    blending_mode : str
+        Texture blending mode. The options are:
+        1. None
+        2. Replace
+        3. Modulate
+        4. Add
+        5. AddSigned
+        6. Interpolate
+        7. Subtract
+    wrap_mode : str
```
ganimtron-10, line 127: `, optional`

fury/actors/effect_manager.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Parameters
+    ----------
+    colormap : np.array (N, 4) or (1, N, 4)
+        RGBA color map array. The array can be two dimensional, although a three dimensional one is preferred.
+    texture_name : str
+        Name of the color map texture to be passed to the actor.
+    target_actor : Actor
+        Target actor to receive the color map texture.
+    interpolate : bool
```
ganimtron-10, line 167: `, optional`

fury/shaders/__init__.py
```diff
@@ -8,6 +8,7 @@
     replace_shader_in_actor,
     shader_apply_effects,
     shader_to_actor,
+    shader_custom_uniforms
 )
 
 __all__ = [
```
ganimtron-10, line 14: you would also need to add `shader_custom_uniforms` in the `__all__`

fury/shaders/base.py
```diff
@@ -412,3 +412,24 @@ def attribute_to_actor(actor, arr, attr_name, deep=True):
     mapper.MapDataArrayToVertexAttribute(
         attr_name, attr_name, DataObject.FIELD_ASSOCIATION_POINTS, -1
     )
+
+def shader_custom_uniforms(actor, shader_type):
+    """Eases the passing of uniform values to the shaders by returning ``actor.GetShaderProperty().GetVertexCustomUniforms()``,
```
ganimtron-10, line 417: `Ease` singular.

fury/shaders/base.py
```diff
@@ -412,3 +412,24 @@ def attribute_to_actor(actor, arr, attr_name, deep=True):
...
+    shader_type : str
+          Shader type of the uniform values to be passed. It can be:
+          * "vertex"
+          * "fragment"
+          * "geometry"
+    """
+    SHADER_FUNCTIONS = {"vertex" : actor.GetShaderProperty().GetVertexCustomUniforms(), 
+                        "fragment" : actor.GetShaderProperty().GetFragmentCustomUniforms(),
+                        "geometry" : actor.GetShaderProperty().GetGeometryCustomUniforms()}
+    if shader_type not in SHADER_FUNCTIONS:
```
ganimtron-10, line 432: `shader_type.lower()`

fury/tests/test_actors.py
```diff
@@ -13,6 +13,7 @@
 from fury.actor import grid
 from fury.decorators import skip_linux, skip_osx, skip_win
 from fury.deprecator import ExpiredDeprecationError
+from fury.actors.effect_manager import EffectManager
```
ganimtron-10, line 16: following the alphabetical order just shift this line at ln 14

fury/tests/test_actors.py
```diff
@@ -1819,3 +1820,84 @@ def test_actors_primitives_count():
...
+    scene = window.Scene()
+    scene.set_camera(position=(-24, 20, -40),
+                    focal_point=(0.0,
+                                 0.0,
+                                 0.0),
+                    view_up=(0.0, 0.0, 0.0))
+
+    manager = window.ShowManager(
+        scene,
+        "demo",
```
ganimtron-10, line 1835: some specific window name

fury/tests/test_actors.py
```diff
@@ -1819,3 +1820,84 @@ def test_actors_primitives_count():
         primitives_count = test_case[2]
         act = act_func(**args)
         npt.assert_equal(primitives_count_from_actor(act), primitives_count)
+
+def test_kde_actor(interactive = False):
+    width, height = (400, 400)
```
ganimtron-10, line 1825: maybe you could use `size = ...` as you are not using the width and height individually

ganimtron-10 (changes requested):
Hey @JoaoDell ,
Please take a look at my below comments.

docs/examples/viz_kde_render.py
```diff
@@ -55,15 +55,15 @@ def normalize(array : np.array, min : float = 0.0, max : float = 1.0, axis : int
 
 manager = ShowManager(
     scene,
-    "demo",
-    (width,
-     height))
+    "KDE Render",
+    (size[0],
```
ganimtron-10, line 59: you can directly use `size`

JoaoDell:
Hey @skoudoro, @filipinascimento and @devmessias, just finished rewriting the tests and cleaning some details on the refactor of the API, so you can review now 👍

docs/examples/viz_kde_render.py
```diff
@@ -0,0 +1,112 @@
...
+# if wished, like the kernel to be used or the colormap desired.
+# The colormaps are by default taken from *matplotlib*, but a
+# custom one can be passed. After calling it, just pass the actor
+# to the scene, and start it as usual.
+
+kde_actor = effects.kde(points, sigmas, kernel="gaussian", colormap="inferno")
+
+manager.scene.add(kde_actor)
+
+interactive = True
```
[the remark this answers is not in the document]
ganimtron-10: This hasn't been updated! While testing you would need it to be true but make sure to set it to false once you are pushing code!

fury/effects/effects.py
```diff
@@ -187,14 +188,50 @@ def colormap_to_texture(
...
     def __init__(self,
-            points : np.ndarray,
-            bandwidths,
-            kernel : str = "gaussian",
-            opacity : float = 1.0,
-            colormap : str = "viridis",
-            custom_colormap : np.array = None):
-        self.points = points
+                 points : np.ndarray,
+                 bandwidths,
```
ganimtron-10, line 220: Why there is no type specified for `bandwidth`?

ganimtron-10 (commented):
While giving a brief look I found these below comments! PTAL.

fury/effects/effects.py
JoaoDell: It is because also allow the passing of single values, not only `np.ndarrays`. I guess I just need to change the description of it to be more clear, what do you think?

filipinascimento (changes requested):
I think the PR is solid and it seems to work fine across all the platforms. I've just a couple comments listed in this review. Please make a new PR for the other changes and for the UI example. We also need to showcase these functionalities with a killer application. Lets try to converge in the next GSOC meeting.

fury/effects/effects.py
JoaoDell: @filipinascimento for some reason `Union` is outputting the following error:

  File "...", line 220, in KDE
    bandwidths : Union(float, np.ndarray),
  File "...", line 387, in __call__
    raise TypeError(f"Cannot instantiate {self!r}")
TypeError: Cannot instantiate typing.Union

Do you know how to fix it?

filipinascimento: Ops it is tUnion[float, np.ndarray]

PR #837, opened 2023-08-11
UI: Adding Text Offset to contain text into the Background

ganimtron-10:
We noticed an issue with the TextBox2D which was, the Text was sometimes touching or getting a bit outside of the background boundary. To fix this I propose to offset the text actor in the TextBlock2D with some offset amount which would contain the text inside the background bounding box.


ganimtron-10:
This PR needs more discussion. So closing this PR for now!

Issue #322, opened 2020-10-01
Render video frames on a cube


Garyfallidis:
Load a video and rendered it on a plane or cube.
The video should be playing as the cube or plane are rotating.

LoopThrough-i-j:
Can I work on the issue?

LoopThrough-i-j:
Hi, I am working on this issue and I was wondering that we should not have an Interactive mode for this as this might lead to crashes because of it being CPU intensive. Should we just have an output video file for this or support some kind of interaction @Garyfallidis what do you think?

skoudoro:
Hi @LoopThrough-i-j,

Do you have a branch to share?  Can you explain why is it so much CPU intensive? How do you do it? It will be good if you could explain more. Thanks

LoopThrough-i-j:
@skoudoro I have raised a PR for the issue.

skoudoro:
- plane is done with #356  
- cube needs to be done
- tutorial needs to be improve with more explanation

SaraSaadoun:
I've tried to work with this issue.
Here is what I've reached by using 4 textures, but there are 2 sides remaining and I don't know what the expected output for this is. 

https://user-images.githubusercontent.com/68401001/222796410-9788e98c-7fe5-4cf8-bd4a-44588b1c353a.mp4

SaraSaadoun:
I've tried to work with this issue.
Here is what I've reached by using 4 textures, but there are 2 sides remaining and I don't know what the expected output for this is, what should it be @skoudoro?



https://user-images.githubusercontent.com/68401001/222967066-16e6c4ca-d9af-4a29-b192-f990fe49d7c6.mp4

robinroy03:
can I work on this issue?

skoudoro:
you can go ahead for the cube @robinroy03

robinroy03:
@skoudoro I have sent an update in Discord (#shaders), could you please check and let me know?

skoudoro:
> @skoudoro I have sent an update in Discord (#shaders), could you please check and let me know?

Hi @robinroy03,

Thank you for the update. Please, if you have something, create a pull request.  Any code discussion are on issue or pull request. Discord are mainly for ideas, debate, communication, easy issue, etc...

robinroy03:
I managed to render frames on a cube using pure VTK, if this is ok I'll convert it to work inside Fury.

[...]

PR #862, opened 2024-02-20
Render a video on a Cube #322

robinroy03:
## Render a video on a cube (#322)

This PR adds a new actor - `texture_on_cube` to `fury/actor.py` which takes 6 different texture arguments for the different sides of a cube. It also adds the tutorial and tests for the same.

### How it is done
I took 6 planes and merged them to form a cube.

### How is video rendered
Video is rendered by changing the texture repeatedly.

## Simple static texture demo:
```py
import numpy as np

arr_0 = np.full((720, 1280), 0, dtype=np.uint8)
arr_255 = np.full((720, 1280), 255, dtype=np.uint8)

arr_red = np.dstack((arr_255, arr_0, arr_0))
arr_white = np.full((720, 1280, 3), 255, dtype=np.uint8)
arr_aqua = np.dstack((arr_0, arr_255, arr_255))
arr_green = np.dstack((arr_0, arr_255, arr_0))
arr_blue = np.dstack((arr_0, arr_0, arr_255))
arr_yellow = np.dstack((arr_255, arr_255, arr_0))

cube = actor.texture_on_cube(arr_white, arr_red, arr_green, arr_blue, arr_yellow, arr_aqua)
cube2 = actor.texture_on_cube(arr_white, arr_red, arr_green, arr_blue, arr_yellow, arr_aqua, (3, 3, 3))

scene = window.Scene()
scene.add(*cube)
scene.add(*cube2)

window.show(scene, size=(1280, 720))
```

https://github.com/fury-gl/fury/assets/115863770/2dd11e67-38f2-4fcd-84a9-a2af09aa1b81


skoudoro:
Hi @robinroy03,

Thank you for this, I will try tomorrow and give you an update

skoudoro:
during this time, can you fix all the pep8 issues?

robinroy03:
@skoudoro I've fixed all the pep8 issues and have placed some of my thoughts in the code as comments. (and also in the PR description)
I'll add tests and finish this once you're ok with them.

skoudoro (changes requested):
Hi @robinroy03,

Thank you for this. This is a good start.

The goal was to show, a different video at each face of the cube.  if you want to use the same video, they should start at different time point.

it is ok to use the texture, but you need to control the different part of your texture. 

After, I will be able to look deeper at the code.

Thank you in advance for your future update.

robinroy03:
Hi @skoudoro,

Are there any inbuilt VTK functions that do this? (mapping different textures to different sides of a cube)
Because when I searched I found this (https://discourse.vtk.org/t/how-to-apply-different-textures-to-each-face-of-a-tetrahedron/2333/2), and I'm not sure if I should go this path directly. 
This is doable as the discourse says, but are there any better alternatives you know of?

robinroy03:
I found this `fury/io/load_cubemap_texture` function and used it. But I am unable to map the cubemap to the cube. (the below code does not work).

```python
import vtk

from fury.io import load_cubemap_texture


tx = load_cubemap_texture(fnames = [
    "negx.jpg",
    "negy.jpg",
    "negz.jpg",
    "posx.jpg",
    "posy.jpg",
    "posz.jpg"
])

scene = window.Scene()

cubeSource = vtk.vtkCubeSource()

mapper = vtk.vtkPolyDataMapper()
mapper.SetInputConnection(cubeSource.GetOutputPort())

cube_actor = vtk.vtkActor()
cube_actor.SetTexture(tx)
cube_actor.SetMapper(mapper)

scene.add(cube_actor)
window.show(scene)
```

Can I get some help regarding the general direction?

robinroy03:
@skoudoro,

Please check the new commit, I've used 6 planes to make the cube. I can use `vtkAssembly()` to make them 1 single actor if needed.

Use a cubemap texture from [here](https://www.humus.name/index.php?page=Textures) to test.

If this is ok, I'll finish this code with some more methods such as `texture_update`, `get_actor` etc ... (or should I format the code differently? I believe this class should go inside actor.py)

Demo Video:

https://github.com/fury-gl/fury/assets/115863770/75f43bd2-c81b-4b60-8f53-c21ef9ae1b5b

robinroy03:
@skoudoro 
I've finish everything, including tests. Please take a look and let me know if I've to make some modifications.

docs/examples/viz_play_cube.py
```diff
@@ -0,0 +1,64 @@
...
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4'
+]
+
+video_captures = [cv2.VideoCapture(source) for source in sources]
+rgb_images = []
+for video_capture in video_captures:
+    _, bgr_image = video_capture.read()
+    rgb_image = cv2.cvtColor(bgr_image, cv2.COLOR_BGR2RGB)
+    rgb_images.append(rgb_image)
```
skoudoro, line 55: needs to be above the timer_callback

fury/actor.py
```diff
@@ -3954,3 +3956,152 @@ def uncertainty_cone(
     angles = main_dir_uncertainty(evals, evecs, signal, sigma, b_matrix)
 
     return double_cone(centers, evecs, angles, colors, scales, opacity)
+
+
+class TexturedCube:
```
skoudoro, line 3961: We prefer to avoid class and be more functional. What is the advantage of using a class here?

skoudoro (commented):
Hi @robinroy03,

Thank you for this. See my initial comment. My main concern is about the use of a class. I think everything can be simpler and easier to maintain with function

robinroy03:
I've made the changes. For `test_actors`, I only did a test for 1 side colour render. I avoided remaining because it all comes from the same pipeline, so must be similar.

skoudoro (commented):
Hi @robinroy03,

After skimming the code , it looks much better!  Please, check our other tutorials to see how it is formatted. Currently, your tutorial will not render well. You can try to render it locally. 

I will try to look deeper on your code later today. Thanks for this

robinroy03:
@skoudoro, I'm sorry but how do I render it locally? Every time I try I get a segf (it's an open issue).

skoudoro:
> What is the procedure to add my code to the docs?

you need to add it in `docs/examples/_valid_examples.toml`

> Every time I try I get a segf

run it multiple times consecutively

robinroy03:
All the said modifications are done. I also committed the new file `_valid_examples.toml`.

robinroy03:
Hi, I was thinking about this, this PR is about adding texture to a `cube`, but it is just as easy to make it a cuboid. Also, we have `box` and `cube` actors already, but they are not very customizable. 

The `texture` actor is effectively `rectangle` but it can take RGB image data directly. 

I feel like we can simplify and merge some actors, idk how but I think it makes some sense. (ex: `rectangle` can be wrapper of vtkPlaneSource with a vtkTransform for size, this will make it super easy to map RGB over it)

robinroy03:
Hi @skoudoro, PTAL.

Thank you :)

robinroy03:
Hi @skoudoro, PTAL.

Thank you.

docs/examples/viz_play_cube.py
```diff
@@ -0,0 +1,108 @@
...
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4',
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4',
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4',
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4',
+    'http://commondatastorage.googleapis.com/gtv-videos-bucket/'
+    + 'sample/BigBuckBunny.mp4'
```
skoudoro, line 30: do you have the https version of this ?

fury/actor.py
```diff
@@ -3954,3 +3956,101 @@ def uncertainty_cone(
...
+         |----|
+
+    Returns
+    -------
+    actors : list[Actor]
+        A list of Actor objects, one for each face of the cube, in order.
+
+    Implementation
+    --------------
+    Check docs/examples/viz_play_cube.py
```
skoudoro, line 3996: to remove

fury/actor.py
```diff
@@ -3954,3 +3956,101 @@ def uncertainty_cone(
...
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    negz : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posx : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posy : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posz : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    centers : tuple (3,)
```
skoudoro, line 3978: `tuple, optional` instead

fury/actor.py
```diff
@@ -3954,3 +3956,101 @@ def uncertainty_cone(
...
+    negy : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    negz : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posx : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posy : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
+    posz : ndarray
+        Input 2D RGB or RGBA array. Dtype should be uint8.
```
skoudoro, line 3977: why it should be uint8?

fury/actor.py
```diff
@@ -3954,3 +3956,101 @@ def uncertainty_cone(
...
+    -------
+    actors : list[Actor]
+        A list of Actor objects, one for each face of the cube, in order.
+
+    Implementation
+    --------------
+    Check docs/examples/viz_play_cube.py
+
+    """
+    plane_objects = [PlaneSource() for _ in range(6)]
```
skoudoro, line 3999: on the docstring, please, add a note that we create 6 actors.

fury/actor.py
```diff
@@ -3954,3 +3956,101 @@ def uncertainty_cone(
...
+        plane_centers,
+        plane_normals
+    ):
+        plane.SetCenter(*center)
+        plane.SetNormal(*normal)
+
+    image_grids = [negx, negy, negz, posx, posy, posz]
+    image_data_objects = [
+        numpy_to_vtk_image_data(grid) for grid in image_grids
+    ]
```
skoudoro, line 4031: is it crashing if not uint8 ?  if yes, you need to check and raise a `ValueError`if there is any issue

skoudoro (commented):
Hi @robinroy03, 

Thank you for this.

6 actors is a lot. I am not sure we would like to have this in `actor.py`. Maybe the function will be only in the tutorial. I will need the opinion of other members of the team. 
See above some comments

docs/examples/viz_play_cube.py
robinroy03: done

fury/actor.py
robinroy03: I added this because I saw `actor.py/texture_update` using the same `Implementation` tag. Removed.

fury/actor.py
robinroy03: resolved.

fury/actor.py
robinroy03: I tested other datatypes and the video is working ok. I added it because `actor.py/texture_update` had this constraint for `arr` parameter. Looks like it doesn't affect `viz_play_cube`. 

I tested different videos with uint16, uint8 and all of them worked without crashing. The only weird thing was uint16 causes a red screen for the first frame (and this is before `texture_update` gets even called inside `viz_play_cube` if I set dtype as uint16. But I do not think that should be taken care by FURY. `cv.read()` default return dtypes work well without errors)

Removed the constraint. Resolved.

fury/actor.py
robinroy03: `uint8` constraint removed. 

Resolved.

fury/actor.py
robinroy03: Done. The `returns` section of docstring now clearly mentions we return 6 Actor objects, one for each face of the cube.

robinroy03:
@skoudoro,

All the requested changes are made. Should I modify the function to be placed inside the tutorial? We'd discussed using 6 different actors to render the cube multiple times.

skoudoro:
> All the requested changes are made. 

Thanks! 

> Should I modify the function to be placed inside the tutorial? 

I do not know yet

> We'd discussed using 6 different actors to render the cube multiple times.

I know... the fact that we talk about it multiple time show that I am not completely satisfy with this solution because I can see long term issue with it. I have hard time to block time to dig in the texture actor to see if there are alternatives or something that we are missing.

robinroy03:
great, I'll also keep researching whether there are better ways to do this.

PR #879, opened 2024-04-20
STYLE: Transition to `ruff` to enforce import statement sorting

jhlegarreta:
Prefer using `ruff` instead of `isort` to enforce import statement sorting.

`ruff`'s import sorting rules are near-equivalent to isort's when using `isort`'s `profile="black"`:
https://docs.astral.sh/ruff/faq/#how-does-ruffs-import-sorting-compare-to-isort

Change the `tool.isort` section in `pyproject` so that the correspoding tasks fall under the responsibility of `flake8`.

Adapt accordingly the pre-commit hook config file to apply the import sorting to local files prior to commits.

Add a GHA workflow to apply the pre-commit hooks in order to ensure that files modified across commits/in PRs comply with the expected formatting.


jhlegarreta:
@skoudoro Slight differences between `isort` and `ruff`. I've checked that the changes are also consistent with what `ruff` did for DIPY, e.g.:
https://github.com/dipy/dipy/pull/3184/files#diff-0b50e0854fbf05412be4bc9cbd9852d4627500eb598d0fc05850998fe93e73e6R6

Would have expected all `from y import z` come after all `import x`, but looks like it is not the case always.

Anyways, green light to fix the issues and continue making steps towards adopting `ruff`?

skoudoro:
Hi @jhlegarreta,

All good.

However, the code format is failing. do you want to fix it here or in a new PR ?

jhlegarreta:
> However, the code format is failing. do you want to fix it here or in a new PR ?

Doing it here.

Issue #912, opened 2024-07-19
Deprecator bug with setuptools >=71.0.3


AlexVCaron:
## Description

Installing `fury 0.10.0` with the new version of `setuptools (71.0.3)` succeeds. However, using the library fails if there is a call to a deprecated component. It comes from the usage of `setuptools.extern` to try to access `packaging`, which is not standard and discouraged by setuptools, see this issue pypa/setuptools#4498 . The problem seems to stem deeper, but note that on our side, everything was working correctly before `setuptools` version `71.0.3`.

## Stacktrace

```
Traceback (most recent call last):
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/pytest_console_scripts.py", line 187, in console_script
    s = entry_points[0].load()
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2749, in load
    return self.resolve()
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/pkg_resources/__init__.py", line 2755, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/home/runner/_work/scilpy/scilpy/scripts/scil_bundle_diameter.py", line 35, in <module>
    from fury import actor
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/actor.py", line 12, in <module>
    from fury.actors.peak import PeakActor
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/actors/peak.py", line 16, in <module>
    from fury.shaders import (
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/shaders/__init__.py", line 1, in <module>
    from fury.shaders.base import (
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/shaders/base.py", line 140, in <module>
    @deprecate_with_version(
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/deprecator.py", line 189, in deprecate_with_version
    'Raises' if is_bad_version(until) else 'Will raise', error_class, until
  File "/home/runner/_work/_tool/Python/3.10.14/x64/lib/python3.10/site-packages/fury/deprecator.py", line 132, in is_bad_version
[...]

PR #913, opened 2024-07-23
Fix: Replace setuptools.extern.packaging with direct packaging import

WassCodeur:
### Description

This PR corrects a problem related to the use of `setuptools.extern.packaging`, which is not standard and not recommended by setuptools. The problem occurs with versions of setuptools >= 71.0.3, causing an ImportError when deprecated components are called.

### Changes

 - Replacement of `setuptools.extern.packaging` by direct import of `packaging` into `deprecator.py`.

### Reason

The `setuptools.extern` module was removed in `>=71.0.0` versions of setuptools, causing import errors. By updating the import declaration to use `packaging`, we ensure that FURY remains compatible.

### Tests

- Run all existing tests to ensure no regression.
- Checked depreciation warning and error handling logic.

Fix issue #912

### DIFF
--- a/fury/deprecator.py
+++ b/fury/deprecator.py
@@ -13,13 +13,9 @@ from inspect import signature
 import re
 import warnings
 
-from fury import __version__
-from fury.optpkg import optional_package
+from packaging import version
 
-# packaging.version.parse is a third-party utility but is used by setuptools
-# (so probably already installed) and is conformant to the current PEP 440.
-# But just if it is not the case, we use distutils
-packaging, have_pkg, _ = optional_package("setuptools.extern.packaging")
+from fury import __version__
 
 _LEADING_WHITE = re.compile(r"^(\s*)")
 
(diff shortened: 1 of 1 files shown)


fury/deprecator.py
```diff
@@ -19,7 +19,7 @@
 # packaging.version.parse is a third-party utility but is used by setuptools
 # (so probably already installed) and is conformant to the current PEP 440.
 # But just if it is not the case, we use distutils
-packaging, have_pkg, _ = optional_package("setuptools.extern.packaging")
+packaging, have_pkg, _ = optional_package("packaging")
```
skoudoro, line 22: Packaging is a mandatory package so you do not need optional_package function and you can import it directly. 

Thank you for the future update!

skoudoro (changes requested):
Hi @WassCodeur , thank you for this update. See below my comment

fury/deprecator.py
WassCodeur: Hi @skoudoro 

Thank you for your feedback. I'll take it into account in my next update.

skoudoro (approved):
LGTM, merging, Thanks @WassCodeur

PR #933, opened 2024-08-27
NF: Consider warnings as errors in documentation CI build

jhlegarreta:
Consider warnings as errors in documentation CI build.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -46,7 +46,7 @@ jobs:
     - name: Build docs
       run: |
         cd docs
-        make -C . html-no-examples
+        make -C . html-no-examples SPHINXOPTS="-W --keep-going"
     - name: Deploy Devs
       if: success() && github.ref == 'refs/heads/master'
       uses: JamesIves/github-pages-deploy-action@v4


.github/workflows/doc.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Build docs
       run: |
         cd docs
-        make -C . html-no-examples
+        make -C . html-no-examples SPHINXOPTS="-W --keep-going -n"
```
skoudoro, line 49: the options `-W --keep-going` are ready to go and can be merged.

the option `n` generates 100 warnings treated as error. Some of them needs to be fixed and some of them needs to ignores using `nitpick-ignore` option: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-nitpick_ignore

So I think, it should be split in 2 or more PRs

Also, we need

- a PR with `-b linkcheck` option and fix the broken link
- a PR with numpydoc check activated

skoudoro (commented):
Hi @jhlegarreta,

Thank you for this, see my comment below

.github/workflows/doc.yml
jhlegarreta: @skoudoro Removed `-n` in the last commit; force pushed.

Issue #924, opened 2024-08-15
Doc Generation Error - HTTP Error 429


robinroy03:
## Description
Doc Generation Error - HTTP error 

Instead of hitting the endpoint repeatedly, trying some exponential backoff might help.

![image](https://github.com/user-attachments/assets/69bad378-7add-4076-a6b4-2db626891b8a)

## Way to reproduce

Run `make -C . clean && make -C . html-no-examples` multiple times.

- [X] Operating system and versions (run `python -c "from fury import get_info; print(get_info())"`)
`{'fury_version': '0.1.0.dev5404+g14851c6', 'pkg_path': '/home/robin/Desktop/playground/fury/fury', 'commit_hash': '14851c6', 'sys_version': '3.11.9 (main, Apr  6 2024, 17:59:24) [GCC 11.4.0]', 'sys_executable': '/home/robin/Desktop/playground/venv/bin/python', 'sys_platform': 'linux', 'numpy_version': '2.0.1', 'scipy_version': '1.14.0', 'vtk_version': '9.3.1', 'matplotlib_version': '3.9.2', 'dipy_version': '1.9.0'}`

deka27:
Maybe API quota is over, i believe it tries to get the contributors from API call everytime you build.

skoudoro:
like @robinroy03 github app, we should start using graphql API

annapurna-gupta:
can i work on this issue?

robinroy03:
@annapurna-gupta sure!

annapurna-gupta:
ok.. i am working on it.

annapurna-gupta:
Hello @robinroy03, I’ve submitted a pull request and would appreciate it if you could take some time to review it.

MohamedSharfan:
Hi @skoudoro, I noticed this issue is still open and i would like to work on this #924

williamcancodee:
Hi team, I am preparing for GSoC 2026, specifically focusing on Project 1: Doc Generation. Currently, I am looking into the codebase in the v2 branch. I noticed that this 429 issue is a huge performance bottleneck for local doc testing. I am currently writing a fix to add the shift/backoff logic for the GraphQL API. I will be opening a PR very soon!

williamcancodee:
Hi, just opened PR #1181  for this!

[...]

PR #941, opened 2024-10-21
Fix: Resolve Documentation Generation HTTP Error

annapurna-gupta:
##Description:

This pull request fixes an issue that caused an HTTP error during the documentation generation process for the FURY project. The error was related to repeatedly hitting an endpoint, which was resolved by implementing an exponential backoff strategy.

##Fixes: issue #924

### DIFF
--- a/fury/tests/test_stream.py
+++ b/fury/tests/test_stream.py
@@ -26,7 +26,6 @@ from fury.stream.server.async_app import (
 from fury.stream.server.main import RTCServer, web_server, web_server_raw_array
 from fury.stream.widget import Widget, check_port_is_available
 
-
 @pytest.fixture
 def loop():
     """Use this fixture to get the event loop.
@@ -103,10 +102,13 @@ def test_pillow():
     centers = np.array([[0, 0, 0], [-1, 0, 0], [1, 0, 0]])
     colors = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
 
-    actors = actor.sphere(centers, colors=colors, radii=0.1)
+    actors = [actor.sphere(center, color, radii=0.1) for center, color in zip(centers, colors)]
+    print(f"Number of actors created: {len(actors)}")
+
 
     scene = window.Scene()
     scene.add(actors)
+
     showm = window.ShowManager(
         scene=scene,
         reset_camera=False,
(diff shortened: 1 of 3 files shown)


fury/tests/test_stream.py
```diff
@@ -103,10 +102,13 @@ def test_pillow():
     centers = np.array([[0, 0, 0], [-1, 0, 0], [1, 0, 0]])
     colors = np.array([[1, 0, 0], [0, 1, 0], [0, 0, 1]])
 
-    actors = actor.sphere(centers, colors=colors, radii=0.1)
+    actors = [actor.sphere(center, color, radii=0.1) for center, color in zip(centers, colors)]
+    print(f"Number of actors created: {len(actors)}")
```
skoudoro, line 106: Can you explain this?  Not sure why you replace 1 actor to multiple actors.

fury/tests/test_stream.py
```diff
@@ -128,21 +130,29 @@ def test_pillow():
...
+        npt.assert_equal(report.objects, 3)
+
+    except Exception as e:
+        print(f"An error occurred during testing: {e}")
+
+    finally:
+        img_buffer_manager.cleanup()
+        stream.stop()
+        stream.cleanup()
+        print("Cleanup completed.")
```
skoudoro, line 156: Why this change fix http error, Can you explain ?

skoudoro (changes requested):
Hi @annapurna-gupta,

Thank you for this. See below my initial review

fury/tests/test_stream.py
annapurna-gupta: The original code used a single actor to represent all the spheres. By creating multiple actors, one for each sphere, we can better control individual properties of each sphere, such as color and position. This approach allows for more flexibility, especially if we need to update or interact with each sphere independently. Additionally, creating multiple actors can improve the clarity of the code when working with individual objects in the scene. The print statement helps confirm the correct number of actors have been created, ensuring that all spheres are accounted for.

fury/tests/test_stream.py
annapurna-gupta: The changes aim to handle potential HTTP errors more gracefully by adding error handling and cleanup processes. The try-except-finally block ensures that, even if an HTTP error or any other exception occurs during streaming or rendering, the img_buffer_manager and stream objects will always be properly cleaned up. This prevents resource leaks and ensures that the stream is stopped and cleaned up correctly, which helps avoid repeated HTTP requests and mitigates potential server overload that could be causing the HTTP errors. Additionally, cleanup ensures that stale data from previous sessions is removed, reducing the likelihood of errors caused by residual states.

fury/tests/test_stream.py
skoudoro: > we can better control individual properties of each sphere, such as color and position? This approach allows for more flexibility, especially if we need to update or interact with each sphere independently.

Unfortunately, I do not agree with this.


Also, having multiple actors impact memories and performance which is why it has been done with one actor. I would recommend to put it back as it was.

fury/tests/test_stream.py
skoudoro: it looks like a chatgpt answer :-). Unfortunately, it does not make sense to me. 

where is the http call?  which http call this change try to fix?

annapurna-gupta: You're right, the response was from ChatGPT. I'm still a beginner, and I'm trying to understand the codebase, but it seems difficult. Because of that, I've made a few mistakes. When you asked me the question, I was confused about the answer, so I used ChatGPT for help. I'm still learning about open source. Please let me know what changes I need to make for this issue, and I'll do them. Could you also suggest some beginner-friendly issues that would help me understand the codebase better?

fury/tests/test_stream.py
annapurna-gupta: ok

annapurna-gupta: i have done it, please check!!

annapurna-gupta:
@skoudoro it would be great if you review the changes again.

skoudoro:
Closing for now. Feel free to reopen the PR with new idea. You should look at graphql and github, play with it in your own repository, then, a PR here will be easy to do.
Thank you for your work

PR #948, opened 2024-12-11
CI:  pin vtk<9.4.0

skoudoro:
The new VTK have a lot of changes that make it incompatible with fury for now.

I do not think we will work on this since we are working on fury v2 which does not use VTK. 

So pinning VTK version to vtk<9.4.0


skoudoro:
All good, merging

PR #955, opened 2024-12-20
RF: Flatten the window folder to the file

maharshi-gor:
### RF: Flatten the window folder to the file

**PR Contents**
- Flatten the window folder to the file.
- Added all the pygfx imports from the window file to the `lib.py`.
- Moved `jupyter-rfb` as an optional dependency under the plot.


maharshi-gor:
Hey @skoudoro , I have not added the docstring intentionally as It will change a lot in the upcoming PRs.

I have also figured out some tests, It will create a change in a lot of files and a lot of places. I am breaking this big change in to small changes, that are easy to review.

PS: This also tackles the issue of jupyter-rfb not being the optional dependency.

fury/lib.py
```diff
@@ -1,14 +1,42 @@
...
 Background = gfx.Background
 BackgroundSkyboxMaterial = gfx.BackgroundSkyboxMaterial
-Camera = gfx.Camera
-Controller = gfx.Controller
+
+# Classes that needed to be written as types
+Camera: TypeAlias = gfx.Camera
+Controller: TypeAlias = gfx.Controller
+Scene: TypeAlias = gfx.Scene
+Viewport: TypeAlias = gfx.Viewport
```
skoudoro, line 30: What do you need the TypeAlias here ?

maharshi-gor: It gives pre-commit error as these classes are used to either extend the class or used as type in a dataclass.

skoudoro: strange, ok, I look into it later, merging

skoudoro (commented):
all ok, merging

fury/lib.py
maharshi-gor: @skoudoro  you can take a look at this. The pre-commit points to this resource when it gives the error. https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases

PR #956, opened 2024-12-20
RF: Examples working properly

maharshi-gor:
### RF: Examples working properly

**PR Contents**

- The examples are working as they are supposed to be with the new `actor.py` added.
- Removed the old examples made for trials.
- Cleaned imports, To not have v2 in the import path.
- Updated the lib file to have the `JupyterCanvas` placeholder so the imports do not throw errors.

**PS: Please merge this PR after #955 is merged.**


skoudoro (approved):
All Good Merging ! Thanks @maharshi-gor

ps: small note, we need to get rid of the `interactive` variabele

PR #958, opened 2024-12-27
Adding box actor

m-agour:
Pygfx calculates normals in a specific way, and for the box to appear with the correct edges and lighting I had to change the vertices numbers to 24 for each box. this is unavoidable.

image


skoudoro (commented):
> for the box to appear with the correct edges and lighting I had to change the vertices numbers to 24 for each box. this is unavoidable

ok, make sense. Can you update all the tests due to this change.  As you can see above, tests are failing due to change, especially `test_primitive.py::test_vertices_primitives`.


All tests should be green before we start to review

skoudoro:
Also, I believe we should rename the primitive `box` in `box_8` and `box_24`, just in case someone needs the box_8 and do not care about the lighting, etc....
OR
keeping the name box but add a parameter to select 8 or 24 vertices, a bit like the sphere.

m-agour:
Hi @skoudoro , thanks for the review and yes it is important to keep both versions maybe we come up with a fix later and revert to the old ways! I will do the same for other actors is needed

m-agour:
I have updated the actor and the tests but, I am not sure about the argument name! @skoudoro

skoudoro (approved):
Looks very good, I like it better like that.

Thank @m-agour!

merging

PR #960, opened 2025-01-20
NF: shaders and materials modifiable & added smooth/flat shading

m-agour:
- Added a slightly modified version of `Pygfx` mesh shader `mesh.wgsl` that calculates the normal on the fly and this is the starting point for billboards.
- for any `actor` user can choose to make it `smooth` or not which is a simple argument to choose between `smooth_shading` & `flat_shading` and yet in the material it make sense to use `flat_shading` since it's more commonly used for materials. 

- [x]  I am working on tests (hopefully by tomorrow)

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -15,6 +15,7 @@ def sphere(
     opacity=None,
     material="phong",
     enable_picking=True,
+    smooth=True,
 ):
     """
     Visualize one or many spheres with different colors and radii.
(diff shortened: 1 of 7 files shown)


skoudoro (approved):
Thank you for this @m-agour!

Looks good to me. We might need to spend time on improving the test framework later on but for now it is good to go!

m-agour:
> Thank you for this @m-agour!
> 
> Looks good to me. We might need to spend time on improving the test framework later on but for now it is good to go!

Thank you @skoudoro! yes tests for this part must be bulletproof I agree

PR #961, opened 2025-01-21
Restored actors visual tests

m-agour:
`window.snapshot` does the job!

PR #964, opened 2025-01-25
RF: install wgpu and mesa driver for ubuntu CI's

skoudoro:
this PR just update our CI's by installing some driver. 

we should be able to use `snapshot` function now. So, some tests has been enable to check that.

supersed #961

### DIFF
--- a/fury/tests/test_actor.py
+++ b/fury/tests/test_actor.py
@@ -1,3 +1,4 @@
+from PIL import Image
 import numpy as np
 import numpy.testing as npt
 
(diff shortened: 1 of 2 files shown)


m-agour:
@skoudoro this is great news!

m-agour:
@skoudoro this is good to merge as well

fury/tests/test_actor.py
```diff
@@ -1,3 +1,4 @@
+from PIL import Image
```
m-agour, line 1: Not sure if alphabetical ordering applies to third-party library imports or not

skoudoro: I did that initially but it seems that uppercase have priority over lowercase then it is alphabetical.

skoudoro:
Thank you for the review @m-agour, merging

fury/tests/test_actor.py
m-agour: Yes I remember now that the pre-commit made it like this

PR #968, opened 2025-02-13
NF: Adding point actor and all its requirements.

ManishReddyR:
Hello.
Added the point actor with required geometry, material and test cases.

point_nc

point_normal

Thank you

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -2,7 +2,6 @@ import numpy as np
 
 from fury.geometry import buffer_to_geometry, create_mesh, create_point
 from fury.material import (
-    _create_marker_material,
     _create_mesh_material,
     _create_points_material,
 )
@@ -1136,7 +1135,7 @@ def star(
 def point(
     centers,
     *,
-    size=4,
+    size=4.0,
     colors=(1.0, 0.0, 0.0),
     material="basic",
     map=None,
@@ -1148,18 +1147,19 @@ def point(
 
     Parameters
     ----------
-    centers : ndarray, shape (N, 3), optional
-        The positions of the points. If None, random points will be generated.
+    centers : ndarray, shape (N, 3).
+        The positions of the points.
     size : float
-        The size (diameter) of the points in logical pixels. Default 4.
+        The size (diameter) of the points in logical pixels.
     colors : ndarray (N,3) or (N,4) or tuple (3,) or tuple (4,), optional
         RGB or RGBA values in the range [0, 1].
     material : str, optional
-        The material type for the points. Options are 'basic', 'gaussian'.
+        The material type for the points.
+        Options are 'basic', 'gaussian'.
     map : TextureMap | Texture
         The texture map specifying the color for each texture coordinate.
-    aa : bool
-        Whether or not the points are anti-aliased in the shader. Default True.
+    aa : bool, optional
+        Whether or not the points are anti-aliased in the shader.
     opacity : float, optional
         Takes values from 0 (fully transparent) to 1 (opaque).
     enable_picking : bool, optional
@@ -1206,9 +1206,10 @@ def marker(
     *,
     size=15,
     colors=(1.0, 0.0, 0.0),
+    material="marker",
     marker="circle",
     edge_color="black",
-    edge_width=1,
+    edge_width=1.0,
     opacity=1.0,
     enable_picking=True,
 ):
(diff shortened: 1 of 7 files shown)


ManishReddyR:
Hello, 
Modified point actor and also added the maker actor. Here is the screenshot of the marker actor.


Screenshot 2025-03-15 at 9 13 19 PM

Thanks

fury/geometry.py
```diff
@@ -1,4 +1,4 @@
-from pygfx import Geometry, Mesh
+from pygfx import Geometry, Mesh, Points
```
maharshi-gor, line 1: import the pygfx things in fury.lib and then use it.

fury/material.py
```diff
@@ -99,3 +99,162 @@ def _create_mesh_material(
...
+     enable_picking : bool, optional
+         Whether the material should be pickable in a scene.
+
+     Returns
+     -------
+     MarkerMaterial
+         A marker material object of the specified type with the given properties.
+    """
+
+    if not (0 <= opacity <= 1):
```
maharshi-gor, line 231: These kind of checks can be curated into a separate file called validation and where we can verify the opacity, color and other things.

All the actors and other functions can leverage that validation if they need to check opacity value or colors

fury/actor.py
```diff
@@ -1127,3 +1131,139 @@ def star(
...
+    map=None,
+    aa=True,
+    opacity=1.0,
+    enable_picking=True,
+):
+    """Visualize one or many points with different features.
+
+    Parameters
+    ----------
+    centers : ndarray, shape (N, 3), optional
```
maharshi-gor, line 1151: we can also put the validations for the shape check here.

fury/actor.py
```diff
@@ -1127,3 +1131,139 @@ def star(
...
+    >>> import numpy as np
+    >>> scene = window.Scene()
+    >>> centers = np.random.rand(1000, 3) * 10
+    >>> colors = np.random.rand(1000, 3)
+    >>> point_actor = actor.point(centers=centers, colors=colors)
+    >>> scene.add(point_actor)
+    >>> show_manager = window.ShowManager(scene=scene, size=(600, 600))
+    >>> show_manager.start()
+    """
+    geo = buffer_to_geometry(
```
maharshi-gor, line 1186: No check for the shape of the centers or if they are empty?

fury/actor.py
```diff
@@ -1127,3 +1131,139 @@ def star(
...
+    >>> import numpy as np
+    >>> scene = window.Scene()
+    >>> centers = np.random.rand(1000, 3) * 10
+    >>> colors = np.random.rand(1000, 3)
+    >>> marker_actor = actor.marker(centers=centers, colors=colors)
+    >>> scene.add(marker_actor)
+    >>> show_manager = window.ShowManager(scene=scene, size=(600, 600))
+    >>> show_manager.start()
+    """
+    geo = buffer_to_geometry(
```
maharshi-gor, line 1254: no check for shape of the centers or if they are empty

fury/geometry.py
```diff
@@ -41,3 +41,23 @@ def create_mesh(geometry, material):
...
+        The geometry object.
+    material : Material
+        The material object.
+
+    Returns
+    -------
+    points : Points
+        The point object.
+    """
+    point = Points(geometry=geometry, material=material)
```
maharshi-gor, line 62: Any public method needs validations for the inputs

maharshi-gor (changes requested):
I have left a lot of comments on this

image

Also @ManishReddyR why do we see the borders changing its colors while moving the scroll wheel

fury/tests/test_actor.py
```diff
@@ -17,46 +17,46 @@ def validate_actors(centers, colors, actor_type="actor_name"):
     npt.assert_array_almost_equal(mean_vertex, centers[0])
 
     assert get_actor.prim_count == 1
-    fname = f"{actor_type}_test.png"
-    window.snapshot(scene=scene, fname=fname)
+    # fname = f"{actor_type}_test.png"
```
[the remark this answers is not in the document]
ManishReddyR: Due to the new pygfx update, all the commented test cases are getting failed because the actor created is looking like this.

arrow_test_1

I not sure whats happening, but I think its due the change of shader form the recent update.

ManishReddyR:
> I have left a lot of comments on this
> 
> image
> 
> Also @ManishReddyR why do we see the borders changing its colors while moving the scroll wheel



Its due to the over lapping of multiple actors at a time.

fury/actor.py
ManishReddyR: I have added a check for positions, as a validation for buffer_to_geometry in fury/geometry.py

fury/actor.py
ManishReddyR: I have added a check for positions, as a validation for buffer_to_geometry in fury/geometry.py

fury/material.py
```diff
@@ -99,3 +113,130 @@ def _create_mesh_material(
         )
     else:
         raise ValueError(f"Unsupported material type: {material}")
+
+
+def _create_points_material(
+    *,
+    material="basic",
+    color=(1.0, 1.0, 1.0),
+    size=4,
```
maharshi-gor, line 122:
```suggestion
    size=4.0,
```

fury/material.py
```diff
@@ -99,3 +113,130 @@ def _create_mesh_material(
...
+    opacity=1.0,
+    enable_picking=True,
+):
+    """
+     Create a points material.
+
+     Parameters
+     ----------
+     material : str, optional
+         The type of material to create. Options are 'baisc' (default),
```
maharshi-gor, line 135: Let's not create two different materials for the points and have one more if with marker and use the same method for both actors internally. Also the spelling for basic is incorrect in the doc.

maharshi-gor (changes requested):
The PR looks good but there are some docstring changes required and some methods needs refractoring.

ManishReddyR:
Thanks for review @maharshi-gor 

I made the changes which you have suggested.

fury/actor.py
```diff
@@ -1127,3 +1130,146 @@ def star(
...
+    obj = create_point(geo, mat)
+    return obj
+
+
+def marker(
+    centers,
+    *,
+    size=15,
+    colors=(1.0, 0.0, 0.0),
+    material="marker",
```
maharshi-gor, line 1209: Remove material option as it can be marker in marker actor's case

maharshi-gor (changes requested):
Just fix one small comment on the marker actor method parameters.

maharshi-gor (approved):

maharshi-gor:
@ManishReddyR Thank you for the work. LGTM! Merging.

PR #976, opened 2025-04-01
fix: shader compatibility issue temp fix.

m-agour:
A fix for the `mesh.wgsl` shader file producing error every new update of `pygfx` this will hold for a couple of future versions until the new versions of Python doesn't support `pygfx 0.9.0` anymore! Or until we find another more efficient way to do it.


maharshi-gor:
Let's merge this as it will be a temp fix for the shader and the v2 branch will be working fine atleast. we need to create a permanent solution for this.

PR #977, opened 2025-04-09
BF: Fixed package name for jupyter_rfb

maharshi-gor:
- The name was copied from the pip show list and had `jupyter-rfb`.
- While using it with optional_package the `jupyter_rfb works as how they register the base name with python.`


skoudoro (approved):
All good, merging. Thanks @maharshi-gor

PR #978, opened 2025-04-11
Relying on `pygfx` shaders for non-shader geometries

m-agour:
Relying on `pygfx` shaders for non-shader geometries.


skoudoro (approved):
All good, merging

PR #980, opened 2025-04-11
NF: Slicer Actor

maharshi-gor:
- Slicer actor works on 3D volumes, respresenting it in XY, YZ, ZX slices.
- Test cases for slicer actor introduced.
- Rgb support enabled for slicer actor.
- Example written for Slicer Actor.

Screenshot from 2025-04-11 15-18-49
Screenshot from 2025-04-11 15-18-40

The example covers the picking and scrolling to check all the slices in the template.

### DIFF
--- a/docs/examples/viz_slicer.py
+++ b/docs/examples/viz_slicer.py
@@ -1,0 +1,39 @@
+import numpy as np
+from fury import actor, window
+from dipy.data import read_mni_template
+
+###############################################################################
+# Let's read the 3D data from dipy
+
+nifti = read_mni_template()
+data = np.asarray(nifti.dataobj)
+print(data.shape)
+
+###############################################################################
+# Create slice actor to visualize the 3D data as XY, YZ and XZ slices.
+
+slicer_actor = actor.slicer(data)
+scene = window.Scene()
+scene.add(slicer_actor)
+
+
+def handle_pick_event(event):
+    info = event.pick_info
+    intensity = np.asarray(info["rgba"].rgb).mean()
+    print(f"Voxel {info['index']}: {intensity:.2f}")
+
+
+def handle_wheel_event(event):
+    position = slicer_actor.get_slices()
+    position += event.dy // 20
+    position = np.maximum(np.zeros((3,)), position)
+    position = np.minimum(np.asarray(data.shape), position)
+    slicer_actor.show_slices(position)
+
+
+slicer_actor.add_event_handler(handle_pick_event, "pointer_down")
+slicer_actor.add_event_handler(handle_wheel_event, "wheel")
+
+if __name__ == "__main__":
+    show_m = window.ShowManager(scene=scene, title="FURY 2.0: Slicer Example")
+    show_m.start()
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1,9 +1,11 @@
 import numpy as np
 
 from fury.geometry import buffer_to_geometry, create_mesh, create_point
+from fury.lib import Geometry, Group, Texture, Volume, VolumeSliceMaterial
 from fury.material import (
     _create_mesh_material,
     _create_points_material,
+    validate_opacity,
 )
 import fury.primitive as fp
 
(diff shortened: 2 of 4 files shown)


fury/actor.py
```diff
@@ -1270,3 +1272,145 @@ def marker(
...
+            data_shape[2] // 2,
+            data_shape[1] // 2,
+            data_shape[0] // 2,
+        )
+
+    texture = Texture(data, dim=3)
+
+    slices = []
+    for dim in [0, 1, 2]:  # XYZ
+        abcd = [0, 0, 0, 0]
```
skoudoro, line 1343: abcd ?

fury/actor.py
```diff
@@ -1270,3 +1272,145 @@ def marker(
...
+            clim=value_range,
+            interpolation=interpolation,
+            opacity=opacity,
+            pick_write=True,
+        )
+        geo = Geometry(grid=texture)
+        plane = Volume(geo, mat)
+        slices.append(plane)
+
+    class Slicer(Group):
```
skoudoro, line 1357: I do not see the benefit of this class. 

Please create utils function that can be applied to a group. 

for example. `set_visibility` is not specific to slicer. it could be a simple function `set_group_visibility` which will work on any type of group later on. Same logic for all function

fury/actor.py
```diff
@@ -1270,3 +1272,145 @@ def marker(
...
+    opacity=1.0,
+    interpolation="linear",
+    visibility=(True, True, True),
+    initial_slices=None,
+):
+    """Visualize a 3D volume data as a slice.
+
+    Parameters
+    ----------
+    data : ndarray, shape (X, Y, Z) or (X, Y, Z, 3)
```
skoudoro, line 1290: do you plan to create slicer for 4D data later on ? (e.g diffusion data)

skoudoro (changes requested):
Hi @maharshi-gor,

thank you for this. See below my comment

skoudoro:
Also, Do not forget to rebase

fury/actor.py
maharshi-gor: its the coefficients of the plane, they are usually written denoting that but I can say `plane_coeffs`?

fury/actor.py
maharshi-gor: The current slicer always works with 3D data only. In Horizon, we create a new slicer for each volume. As we should not load and create entire data.

If we want to do it in FURY I can incorporate that in medical actor for the slicer.

fury/actor.py
skoudoro: I still do not understand. What do you mean by plane coefficients? 

I will search on the web

skoudoro: ah ok, equation of the plane, ok, fine

fury/actor.py
skoudoro: ok sounds good. thanks for clarification !

fury/actor.py
maharshi-gor: The general form of a plane equation is `ax + by + cz = d`

fury/actor.py
maharshi-gor: The problem is the slice functions are specific to planes and not all groups will have planes in them they can have something different. For example, gfx.Scene is also a subclass of Group.

skoudoro (changes requested):
Can you rebase and fix the new docstring standard and requirements. 

thank you

docs/examples/viz_slicer.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
```
skoudoro, line 1: There is a high probability that we will never come back to this Tutorial. Or in a very long time....

So, this tutorial needs much more explanation. It needs a header + title, like in DIPY.

docs/examples/viz_slicer.py
```diff
@@ -0,0 +1,40 @@
...
+from fury import actor, window
+from fury.utils import get_slices, show_slices
+from dipy.data import read_mni_template
+
+###############################################################################
+# Let's read the 3D data from dipy
+
+nifti = read_mni_template()
+data = np.asarray(nifti.dataobj)
```
skoudoro, line 10: not sure an user in mechanical fluid understand this. You have to explain more or use a more general data.

Why `nifti.dataobj`? etc

docs/examples/viz_slicer.py
```diff
@@ -0,0 +1,40 @@
...
+
+slicer_actor = actor.slicer(data)
+scene = window.Scene()
+scene.add(slicer_actor)
+
+
+def handle_pick_event(event):
+    info = event.pick_info
+    intensity = np.asarray(info["rgba"].rgb).mean()
+    print(f"Voxel {info['index']}: {intensity:.2f}")
```
skoudoro, line 24: is it always needed or just and example ?

docs/examples/viz_slicer.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def handle_wheel_event(event):
```
skoudoro, line 27: Does it means wheel will not work if we do not add that ?

fury/utils.py
```diff
@@ -359,3 +362,111 @@ def fix_winding_order(vertices, triangles, *, clockwise=False):
...
+    if not isinstance(group, Group):
+        raise TypeError("group must be an instance of Group.")
+
+    opacity = validate_opacity(opacity)
+
+    for child in group.children:
+        child.material.opacity = opacity
+
+
+def _valid_slices(group):
```
skoudoro, line 410: I would rename this `validate_slices_group` and no need to be private

docs/examples/viz_slicer.py
skoudoro: Also, it does not work well, the camera is scrolling as the same time of the slice. it is confusing.

docs/examples/viz_slicer.py
skoudoro: voxel intensity seems wrong or is it between 0 and 1 ?

skoudoro (changes requested):
Hi @maharshi-gor,

Overall, Looks very good. it is almost ready to go. 

See my comments below. A little effort need to be done concerning the tutorial. 

Thanks!

docs/examples/viz_slicer.py
maharshi-gor: Yes this is not the default behavior out of the box I created to confirm that we can scroll through slices.

I can update the tutorial to handle it with arrow keys or I can add this out of the box in the actor.

I refrain from adding to the actor as it limits the customizability.

docs/examples/viz_slicer.py
maharshi-gor: Between 0 and 1. Which is the case in Horizon as well

Issue #982, opened 2025-04-17
Line Actor creates issue if not pass the colors


maharshi-gor:
```
import numpy as np

from fury import actor, window

scene = window.Scene()
lines = [np.random.rand(10, 3) for _ in range(5)]
# colors = np.random.rand(5, 3)
line_actor = actor.line(lines=lines)
scene.add(line_actor)
show_manager = window.ShowManager(scene=scene, size=(600, 600))
show_manager.start()
```

Error: `ValueError: Color array size doesn't match either vertex count or segment count`

m-agour:
Solved in #983.

m-agour:
Solved in #983.

PR #983, opened 2025-04-17
Fix `line` actor colors 

m-agour:
Fix line triggering error when "color per whole set of lines" case

issue: #982


maharshi-gor:
Thank you @m-agour ! It fixes the issue.

PR #984, opened 2025-04-17
NF: added `ellipsoid` actor 

m-agour:
image

``` python
import numpy as np
from fury.actor import ellipsoid
from fury.window import ShowManager, Scene
from scipy.spatial.transform import Rotation as R

centers = np.array([[0, 0, 0], [7, 0, 0], [0, 7, 0]])
lengths = np.array([[3, 1.5, 1], [3, 1, 1], [2, 3, 1]])

axes = np.array([
    np.eye(3),
    R.from_euler('xyz', [45, 45, 45], degrees=True).as_matrix(),
    R.from_euler('xyz', [0, 90, 0], degrees=True).as_matrix()
])

actor = ellipsoid(centers, orientation_matrices=axes, lengths=lengths, colors=colors)

scene = Scene()
scene.add(actor)

show_m = ShowManager(scene=scene, title="FURY 2.0: Ellipsoid")
show_m.start()

```

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -272,7 +272,7 @@ def ellipsoid(
     centers,
     *,
     orientation_matrices=None,
-    lengths=(4, 3, 2),
+    lengths=(4, 2, 2),
     colors=(1, 0, 0),
     opacity=None,
     phi=16,
@@ -335,6 +335,7 @@ def ellipsoid(
 
     orientation_matrices = np.asarray(orientation_matrices)
     lengths = np.asarray(lengths)
+    colors = np.asarray(colors)
 
     if centers.ndim == 1:
         centers = centers.reshape(1, 3)
@@ -348,8 +349,12 @@ def ellipsoid(
         lengths = lengths.reshape(1, 3)
     if lengths.ndim != 2 or lengths.shape[1] != 3:
         raise ValueError("Lengths must be (N, 3) array")
+    if lengths.size == 3:
+        lengths = np.tile(lengths.reshape(1, -1), (centers.shape[0], 1))
     if lengths.shape != centers.shape:
         raise ValueError("Lengths must match centers shape")
+    if colors.size == 3 or colors.size == 4:
+        colors = np.tile(colors.reshape(1, -1), (centers.shape[0], 1))
 
     base_verts, base_faces = fp.prim_sphere(phi=phi, theta=theta)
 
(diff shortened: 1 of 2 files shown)


fury/actor.py
```diff
@@ -247,6 +258,138 @@ def sphere(
...
+        Smooth shading
+
+    Returns
+    -------
+    actor : fury.actor.Actor
+        Ellipsoid actor with transformations applied
+
+    Examples
+    --------
+    >>> from fury import window, actor
```
maharshi-gor, line 310: Example uses the show method incorrectly.

We use show with list of actors now.

Correct version
```
import numpy as np

centers = np.array([[0, 0, 0]])
lengths = np.array([[2, 1, 1]])
colors = np.array([[1, 0, 0]])
ellipsoid = actor.ellipsoid(centers=centers, lengths=lengths, colors=colors)
window.show([ellipsoid])
```

fury/actor.py
```diff
@@ -247,6 +258,138 @@ def sphere(
...
+    material : str, optional
+        Surface material type
+    enable_picking : bool, optional
+        Allow mouse selection
+    smooth : bool, optional
+        Smooth shading
+
+    Returns
+    -------
+    actor : fury.actor.Actor
```
maharshi-gor, line 305: Update this as we do not have an actor class

fury/actor.py
```diff
@@ -247,6 +258,138 @@ def sphere(
...
+    smooth=True,
+):
+    """
+    Create ellipsoid actor(s) with specified orientation and scaling.
+
+    Parameters
+    ----------
+    centers : ndarray (N, 3)
+        Centers of the ellipsoids
+    axes : ndarray, shape (N, 3, 3) or (3, 3), optional
```
maharshi-gor, line 281: Should we call it R?

fury/actor.py
```diff
@@ -65,15 +69,22 @@ def actor_from_primitive(
...
+
+    if repeat_primitive:
+        res = fp.repeat_primitive(
+            vertices,
+            faces,
+            directions=directions,
+            centers=centers,
+            colors=colors,
+            scales=scales,
+        )
```
maharshi-gor, line 81:
```suggestion
        res = fp.repeat_primitive(
            vertices,
            faces,
            centers,
            directions=directions,
            colors=colors,
            scales=scales,
        )
```

This allows everyone by looking why the if repeat_primitive is required.

maharshi-gor (changes requested):
Hello @m-agour Thank you for this work!

Please look at the mentioned comments.

fury/actor.py
m-agour: `Orientations` would be more suitable here. `axes` (referring to basis_axes) is also fine. `rotation_matrices` is also goos but very verbose same as `orientation_matrices`  but it's the best option here (orientation_matrices) . will change to `orientation_matrices` .

skoudoro (changes requested):
Hi @m-agour,

this code below faile
```python
actor = ellipsoid(centers, orientation_matrices=axes)
# or 
actor = ellipsoid(centers)
```
with 

```
File ~/devel/fury/fury/actor.py:342, in ellipsoid(centers, orientation_matrices, lengths, colors, opacity, phi, theta, material, enable_picking, smooth)
    340     raise ValueError("Lengths must be (N, 3) array")
    341 if lengths.shape != centers.shape:
--> 342     raise ValueError("Lengths must match centers shape")
```

however `lengths` is optional

Can you have a look? Please, update the tests also

skoudoro:
Thank you for the update.

it seems the new docstring checker work as expected. please, check the failing ci: https://github.com/fury-gl/fury/actions/runs/14580809257/job/40896909086?pr=984#step:4:128

PR #987, opened 2025-04-18
[Draft] NF: Spline tube actor

m-agour:
Streamtube actor with adaptive smoothing (big performance improvement for the non-shader-based version)

```python
import numpy as np
scene = window.Scene()
points = np.array([[4, 1, 3], [0, 1, 0], [1, 0, 1], [2, 9, 2], [3, 0, 0], [4, 10, 3]])
tube_actor = actor.tube(points=points, colors=(1, 0, 1))
scene.add(tube_actor)
show_manager = window.ShowManager(scene=scene, size=(600, 600))
show_manager.start()
```

image

image

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1,6 +1,8 @@
 import numpy as np
 
 from fury.geometry import (
+    _generate_smooth_points,
+    _tube_frame,
     buffer_to_geometry,
     create_line,
     create_mesh,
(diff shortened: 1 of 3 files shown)


skoudoro (changes requested):
Hi @m-agour,

That's great! thank you for this first version. 

please, fix the current conflict with the branch. Also rebase to have the docstring automatically checked.

Also, use the power of numpy to avoid your loop and `range(len`.  this will naturally improve performance. 

Thank you in advance.

m-agour:
> Hi @m-agour,
> 
> That's great! thank you for this first version.
> 
> please, fix the current conflict with the branch. Also rebase to have the docstring automatically checked.
> 
> Also, use the power of numpy to avoid your loop and `range(len`. this will naturally improve performance.
> 
> Thank you in advance.

Thank you @skoudoro, I tried to think of a way to do this without loops but it's a tough problem already and couldn't think of a way to do this. this can be parallelized but for a working version I have just created a new PR for the streamtube and as for this PR it might be called spline_tube or something but I will draft it for now.

PR #988, opened 2025-04-18
NF: Adding disk actor

ManishReddyR:
Added disk actor and its test cases. Here is demo of disk actor:

Screenshot 2025-04-18 at 9 00 46 AM

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1364,6 +1364,79 @@ def star(
     )
 
 
+def disk(
+    centers,
+    *,
+    colors=(1.0, 1.0, 1.0),
+    radii=0.5,
+    sectors=36,
+    scales=(1.0, 1.0, 1.0),
+    directions=(0.0, 0.0, 0.0),
+    opacity=None,
+    material="phong",
+    enable_picking=True,
+):
+    """Visualize one or many disks with different features.
+
+    Parameters
+    ----------
+    centers : ndarray, shape (N, 3)
+        Disk positions.
+    colors : ndarray (N,3) or (N, 4) or tuple (3,) or tuple (4,), optional
+        RGB or RGBA (for opacity) R, G, B and A should be at the range [0, 1].
+    radii : float or ndarray (N,) or tuple, optional
+        The radius of the disks, single value applies to all disks,
+        while an array specifies a radius for each disk individually.
+    sectors : int, optional
+        The number of divisions around the disk's circumference .
+        Higher values produce smoother disk.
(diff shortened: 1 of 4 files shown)


m-agour (commented):
Hi @ManishReddyR ,

Thanks for the PR. Please address comments below.

fury/actor.py
```diff
@@ -1210,6 +1210,79 @@ def star(
     )
 
 
+def disk(
```
skoudoro, line 1213: maybe we should rename it `flat_disk`

skoudoro (commented):
please, rebase or merge on v2

fury/actor.py
m-agour: @skoudoro in my opinion `disk` is the best name for this. flat makes it confusing since we have other 2d actors without flat.

skoudoro: I understand your point and I finally agree.

Confused, with disk which look like  circle. or it could be a plate.   

Do you agree that 3D disk == cylinder ? 

I need to check the doc to see if we explicity say it is a 2D shape. otherwise, it can be confusing

m-agour: Yes, I think so as well a 3D disk is a cylinder. disk is flat generally and a circle is optimally a round line.

skoudoro: ok, sorry for that @ManishReddyR, put it back to disk. thanks

fury/primitive.py
```diff
@@ -1299,3 +1299,46 @@ def prim_cone(*, radius=0.5, height=1, sectors=10):
...
 
     return vertices, triangles
+
+
+def prim_disk(*, radius=0.5, sectors=36):
+    """Return vertices and triangles for a disk.
+
+    Parameters
+    ----------
+    radius : float
```
skoudoro, line 1309: `, optional`

fury/primitive.py
```diff
@@ -1299,3 +1299,46 @@ def prim_cone(*, radius=0.5, height=1, sectors=10):
...
+        Radius of the disk.
+    sectors : int
```
skoudoro, line 1311: `, optional`

skoudoro (commented):
Hi @ManishReddyR,

Overall, looks good.

Please, rebase onto v2 branch. This rebase will add the automatic check of the docstring.  I will wait for this rebase before any merging

ManishReddyR:
Hello @skoudoro 

I have completed rebase, but with lot of commits.

PR #990, opened 2025-04-29
RF: Added position argument for text actor and name fix.

ManishReddyR:
- Added position argument for text actor.
- Name fix form test_flat_disk to test_disk.

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1581,6 +1581,7 @@ def text(
     text,
     *,
     colors=(1.0, 1.0, 1.0),
+    position=(0.0, 0.0, 0.0),
     font_size=1.0,
     family="Arial",
     anchor="middle-center",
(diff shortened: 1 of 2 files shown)


skoudoro (changes requested):
Good catch and Thank you for this @ManishReddyR.

it means that position is not tested. 

Can you updated the test `validated_actors` with different positions.

We can merge after that.

Thank you

ManishReddyR:
> Good catch and Thank you for this @ManishReddyR.
> 
> it means that position is not tested.
> 
> Can you updated the test `validated_actors` with different positions.
> 
> We can merge after that.
> 
> Thank you




Hello @skoudoro 

- For the case of text, we are using the test_text function, which already consists a test for position.
- And also adding few more cases to it.

skoudoro (approved):
All good, Thanks @ManishReddyR, merging

PR #991, opened 2025-04-29
NF: Adding Triangle actor.

ManishReddyR:
- Added triangle actor and its primitive.
- Added test cases for actor and primitive.
- Example:

image


skoudoro (approved):
All good, Merging,

Thanks @ManishReddyR

PR #1004, opened 2025-06-10
BF: Long key press event None fixed

maharshi-gor:
**PR Content**

Fix for the below error when clicked outside the canvas.

```
Error during handling key_up event
  File "/Users/maharshigor/miniforge3/envs/fury/lib/python3.11/site-packages/rendercanvas/_coreutils.py", line 41, in log_exception
    yield
  File "/Users/maharshigor/miniforge3/envs/fury/lib/python3.11/site-packages/pygfx/objects/_events.py", line 395, in handle_event
    callback(event)
  File "/Users/maharshigor/Desktop/GRG/fury/fury/window.py", line 637, in _set_key_long_press_event
    self._key_long_press.cancel()
```

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -633,7 +633,7 @@ class ShowManager:
             self._key_long_press = asyncio.create_task(
                 self._handle_key_long_press(event)
             )
-        else:
+        elif self._key_long_press is not None:
             self._key_long_press.cancel()
             self._key_long_press = None
 


skoudoro (changes requested):
Thanks @maharshi-gor !

tests that show the fix works ?

maharshi-gor:
Not really @maharshi-gor , we need to click outside the window and need to be interactive to even see the error.

That was the reason we were not able to catch it before.

skoudoro:
I am pretty sure we can reproduce it programmatically, but it might take time. merging for now but we needs to find a fast and good way.
`interactive` == replay to implement.

PR #1013, opened 2025-06-24
CI: Activate Build doc for v2 branch

skoudoro:
all in the title. Activate build doc CI for the v2 branch


skoudoro:
This initial version is working, we need to check why it fails with `make -C . html`

PR #1014, opened 2025-06-25
RF: `actor.py` file refractored into multiple files

maharshi-gor:
### RF: `actor.py` file refractored into multiple files

**PR Contents**

- Files include init, curved, polyhedron, planar, core, slicer, surface, bio.
- Reduces the load on low RAM machines, with increasing extension for productivity.
- Slicer actor renamed to `data_slicer`


fury/__init__.pyi
```diff
@@ -6,7 +6,6 @@
 
 __all__ = [
     "actor",
-    "bio",
     #     "actors",
```
skoudoro: please, remove this one also

fury/actor/__init__.py
```diff
@@ -0,0 +1,62 @@
+"""Actor module for FURY."""
```
skoudoro: we try to follow the lazy loading technique.  

So, you need to create a `init.pyi` and this init should be updated

More information here: https://scientific-python.org/specs/spec-0001/

fury/actor/bio.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
 
-from fury.actor import slicer
+from fury.actor.slicer import slicer
```
skoudoro: this is strange, we would like to avoid that, repeating twice slicer

skoudoro (changes requested):
Hi @maharshi-gor,

looks good overall. just make sure to add the lazy loading. see my comment below

fury/actor/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+import warnings
+
+import lazy_loader as lazy
+
+from fury.optpkg import optional_package
+from fury.pkg_info import __version__, pkg_commit_hash
+
+__getattr__, __dir__, __all__ = lazy.attach_stub(__name__, __file__)
+
+__all__ += [
```
skoudoro: everything below that to remove. Not sure why you copy that....

make sure to understand the lazy loading module

fury/actor/slicer.py
```diff
@@ -0,0 +1,677 @@
...
+from fury.utils import (
+    create_sh_basis_matrix,
+    get_lmax,
+    set_group_opacity,
+    set_group_visibility,
+    show_slices,
+)
+
+
+def im_slicer(
```
skoudoro: to rename `data_slicer`

skoudoro (changes requested):
Almost done, see my last comment below @maharshi-gor.

make sure to understand lazy loading

skoudoro (approved):
All good, merging. Thanks @maharshi-gor

PR #1015, opened 2025-07-15
ENH: Update to use texture coordinates in surface

maharshi-gor:
- Added texture coordinates attribute in the surface actor.
- Provided validation for the texcoords.
- Provided test cases for the same.

### DIFF
--- a/fury/actor/surface.py
+++ b/fury/actor/surface.py
@@ -20,6 +20,7 @@ def surface(
     colors=None,
     texture=None,
     texture_axis="xy",
+    texture_coords=None,
     opacity=1.0,
 ):
     """Create a surface mesh actor from vertices and faces.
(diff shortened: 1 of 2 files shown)


fury/actor/surface.py
```diff
@@ -87,11 +91,12 @@ def surface(
         )
 
         tex = load_image_texture(texture)
-        texcoords = generate_planar_uvs(vertices, axis=texture_axis)
+        if texture_coords is None:
```
skoudoro, line 94: if it is not none, is there a way to check if it is in the correct format ?

skoudoro (commented):
All good @maharshi-gor , Can you address my small comment, then we can merge

fury/actor/surface.py
maharshi-gor: We can just check if it is of size (num_vertices, 2)

Let me add that check

fury/actor/surface.py
```diff
@@ -87,11 +91,21 @@ def surface(
...
+            texture_coords = generate_planar_uvs(vertices, axis=texture_axis)
+        else:
+            if (
```
skoudoro: let's be pythonic, Can you use `elif` instead. Otherwise it is good to go

skoudoro (commented):
Last comment, then we can merge

fury/actor/surface.py
skoudoro: Can you address this that we can merge ? Also, a rebase is needed to fix the conflict. 

Thanks

skoudoro (approved):
Ok, I fixed it, all good, merging

PR #1017, opened 2025-07-17
NF: Adding Ring Actor

ganimtron-10:
Adding Ring actor from primitive with relevant primitive and actor tests

image

### DIFF
--- a/fury/actor/planar.py
+++ b/fury/actor/planar.py
@@ -611,3 +611,84 @@ def image(
 
     obj.local.rotation = rot.as_quat()
     return obj
+
+
+def ring(
+    centers,
+    *,
+    inner_radius=0.5,
+    outer_radius=1.0,
+    radial_segments=1,
+    circumferential_segments=32,
+    directions=(0, 0, 0),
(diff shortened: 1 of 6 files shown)


fury/actor/planar.py
```diff
@@ -611,3 +611,84 @@ def image(
...
     return obj
+
+
+def ring(
+    centers,
+    *,
+    inner_radius=0.5,
+    outer_radius=1.0,
+    radial_segments=1,
+    circumferential_segments=32,
```
skoudoro, line 622: All these inputs variables should follow the same principle of `colors, directions, scale`.

Maybe, we want ring, with different `inner_radius` and `outer_radius`. So these variables should accept a float or a  ndarray with a size of `centers`

skoudoro (approved):
Hi @ganimtron-10,

Thank you for this. it looks good to me. See my comment below, but it can be addressed in a new PR. Merging

fury/actor/planar.py
ganimtron-10: Sure will raise a PR updating the actor accordingly.

PR #1022, opened 2025-08-18
ENH: Update the odf parameters

maharshi-gor:
### ENH: Update the ODF parameters

**PR Contents**

- Scale and l_max parameters can be set and updated.
- These can reflect changes in real-time.

### DIFF
--- a/fury/material.py
+++ b/fury/material.py
@@ -600,7 +600,7 @@ class SphGlyphMaterial(MeshPhongMaterial):
 
     Parameters
     ----------
-    l_max : int, optional
+    n_coeffs : int, optional
         The maximum spherical harmonic degree.
     scale : int, optional
         The scale factor.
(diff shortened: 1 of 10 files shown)


fury/material.py
```diff
@@ -616,14 +616,14 @@ class SphGlyphMaterial(MeshPhongMaterial):
...
-        l_max="i4",
+        n_coeffs="i4",
         scale="f4",
     )
 
     def __init__(
         self,
-        l_max=4,
-        scale=2,
+        n_coeffs=-1,
```
skoudoro, line 625: what does it means `-1`? `n_coeffs=None`, then, call your function `get_n_coeff()` if none

skoudoro (commented):
`l_max`seems more explicit than `n_coeffs`.  when you check the literrature or other library like scipy or dipy, they all use `l`.  
why 36 n_coeffs? how someone will guess this number?  Can you explain your logic here?

fury/material.py
maharshi-gor: If it is -1 it will accept what ever number of coefficient present in the given data. This variable overrides if you would only like to use certain limit of n_coeffs and not all coeffs to get faster render. time on large data

Let me write description so it becomes clear.

maharshi-gor:
@skoudoro  we still use `l_max` as the user facing parameter if you can check. the only problem in material is that it does not hold the `basis_type` details to calculate the `n_coeffs` from `l_max` so user can set the value in the actor it will internally calculate `n_coeffs` and set in the material so we can update the values on the fly in the shader code.

skoudoro (approved):
ok, got it merging

PR #1024, opened 2025-08-20
BF: Updated the line_buffer_separator

maharshi-gor:
### BF: Updated the line_buffer_separator

**PR Contents**

- removed redundant `color_mode` parameter from the function
- updated it to expected behavior on single color.
- updated test cases to the expected behavior.
- updated line actor to use the latest updated function.

**Code to try out: same as number of lines**
```
def simulated_bundle(no_streamlines=10, waves=False):
    t = np.linspace(20, 80, 200)
    # parallel waves or parallel lines
    bundle = []
    for i in np.linspace(-5, 5, no_streamlines):
        if waves:
            pts = np.vstack((np.cos(t), t, i * np.ones(t.shape))).T
        else:
            pts = np.vstack((np.zeros(t.shape), t, i * np.ones(t.shape))).T
        bundle.append(pts)

    return np.asarray(bundle)


scene = window.Scene()
bundle = simulated_bundle(no_streamlines=3, waves=False)
print(f"Number of streamlines: {bundle.shape}")
slines_actor = actor.line(np.array(bundle), colors=[(1, 0, 0), (0, 1, 0), (0, 0, 1)])
scene.add(slines_actor)
window.show(actors=[slines_actor])
```

**Code to try out: single color**
```
slines_actor = actor.line(np.array(bundle), colors=(1, 0, 0))
```

**Code to try out: No color**
```
slines_actor = actor.line(np.array(bundle))
```

**Code to try out: same as all vertices but in the fashion that lines are provided**
```
slines_actor = actor.line(np.array(bundle), colors=np.random.rand(*bundle.shape))
```

**Code to try out: same as all vertices in flatten format**
```
slines_actor = actor.line(
    np.array(bundle), colors=np.random.rand(np.prod(bundle.shape[:2]), 3)
)
```

### DIFF
--- a/fury/actor/core.py
+++ b/fury/actor/core.py
@@ -290,30 +290,18 @@ def line(
     >>> show_manager.start()
     """
 
-    lines_positions, lines_colors = line_buffer_separator(
-        lines, color=colors, color_mode="auto"
-    )
+    lines_positions, lines_colors = line_buffer_separator(lines, color=colors)
 
     geo = buffer_to_geometry(
         positions=lines_positions.astype("float32"),
-        colors=lines_colors.astype("float32")
-        if lines_colors is not None
-        else np.empty_like(lines_positions),
+        colors=lines_colors.astype("float32"),
     )
-
-    if lines_colors is None:
-        material_mode = "auto"
-        material_colors = None
-    else:
-        material_mode = "vertex"
-        material_colors = lines_colors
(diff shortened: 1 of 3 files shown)


skoudoro (commented):
Thank you for this @maharshi-gor! 

It looks good, let me change branch to test this

PR #1025, opened 2025-08-20
BF: Handle empty scene case in update camera

maharshi-gor:
- Handled when scene is empty and size is not passed


skoudoro (approved):
Thanks! all good, merging

PR #1028, opened 2025-08-29
BF: Allows the lines to have different size

maharshi-gor:


skoudoro (approved):
LGTM, merging, thanks @maharshi-gor

PR #1029, opened 2025-09-04
NF: Plannar projection of lines can be captured

maharshi-gor:
### NF: Plannar projection of lines can be captured

**PR Contents**

- Introduces a new actor called `line_projection`.
- Compute shader to calculate the intersection points.
- Fixes opacity issue in points material method.
- Adds functionality to update slices for groups apart from img slicer.
- Added Test cases.
- Added Tutorial for the same.

**NOTE: CAUTION**
I have freezed the pygfx version as it has created some breaking changes in the latest release. I have created an issue to track the same as well.

**Code to try**

Please use the tutorial `viz_line_projections.py` in the PR

**Screenshots**
Screenshot 2025-09-16 at 12 10 28 AM
Screenshot 2025-09-16 at 12 11 53 AM

### DIFF
--- a/docs/examples/viz_line_projections.py
+++ b/docs/examples/viz_line_projections.py
@@ -1,0 +1,73 @@
+"""
+============================
+Line Projection Visualization
+============================
+
+This tutorial demonstrates how to use FURY's `actor.line_projection` to visualize the
+planar projection of 3D lines onto a specified plane. This is useful for understanding
+how streamlines or trajectories intersect or relate to a given plane in 3D space.
+
+We will create a scene with several colored lines, a square plane, and their projections
+onto the plane. The projected lines are rendered with customizable thickness, opacity,
+and outline for better visualization.
+"""
+
+import numpy as np
+from fury import actor, window
+
+###############################################################################
+# Define 3D lines and their colors.
+# Each line is represented by two endpoints in 3D space.
+
+lines = [
+    np.asarray([[20, 20, -2], [20, 0, 3]]),
+    np.asarray([[10, 0, -2], [0, 0, 3]]),
+    np.asarray([[0, -20, -2], [0, -20, 3]]),
+]
+
+colors = [
+    (1, 0, 0),  # Red
+    (0, 1, 0),  # Green
+    (0, 0, 1),  # Blue
+]
+
+###############################################################################
+# Create a square plane actor.
+# The plane will serve as the target for projecting the lines.
+
+plane = actor.square(
+    np.zeros((1, 3)),
+    directions=(1, 0, 0),
+    scales=(50, 50, 1),
+)
+
+###############################################################################
+# Create the original line actor and the projected line actor.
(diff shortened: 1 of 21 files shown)


fury/actor/planar.py
```diff
@@ -692,3 +709,326 @@ def ring(
...
         enable_picking=enable_picking,
     )
+
+
+class LineProjection(Points):
+    """Initialize the line projection object.
+
+    Parameters
+    ----------
+    lines : list
```
Garyfallidis: sequence

docs/examples/viz_line_projections.py
```diff
@@ -0,0 +1,73 @@
...
+
+line_actor = actor.streamlines(lines, colors=colors, thickness=10, opacity=0.25)
+
+projection = actor.line_projection(
+    lines,
+    plane=(0, 0, -1, 0),  # Plane equation: z = 0 or XY plane
+    colors=colors,
+    thickness=10,
+    outline_thickness=3,
+    outline_color=(0, 0, 0),
```
Garyfallidis, line 56: Default lift should be 0.

Garyfallidis:
See if you can support the latest version before merging this.

docs/examples/viz_line_projections.py
```diff
@@ -0,0 +1,73 @@
+"""
+============================
```
skoudoro: too short, it generates a warning. 

I think I will activate warnings as errors in general. we start to accumulate warnings which is not good

docs/examples/viz_line_projections.py
```diff
@@ -0,0 +1,73 @@
...
+
+###############################################################################
+# Create the original line actor and the projected line actor.
+# The projected lines are visualized using `actor.line_projection`.
+
+line_actor = actor.streamlines(lines, colors=colors, thickness=10, opacity=0.25)
+
+projection = actor.line_projection(
+    lines,
+    plane=(0, 0, -1, 0),  # Plane equation: z = 0 or XY plane
```
skoudoro: This does not make sense to me, and it is not user friendly. 
Why not using directly `plane`  which is `actor.square`. Please, explain the link between both...

skoudoro (changes requested):
Hi @maharshi-gor,

I will continue my review, but here, 2 issues that needs to be fixed

docs/examples/viz_line_projections.py
maharshi-gor: This will not draw a plane, this is just to tell the the actor which plane to use to intersect a line. 

The tutorial itself creates a square actor to draw the plane. But the projection actor only shows the points.

Let me know if you need more information on this.

docs/examples/viz_line_projections.py
maharshi-gor: Let me create a fix PR for the usual test case warning we get and we can enable the warning to error option

PR #1040, opened 2025-09-30
ENH: Surface now uses normals and phong material

maharshi-gor:
### ENH: Surface now uses normals and phong material

**PR Contents**
- Simplified `_create_mesh_material` method to have single dict of parameters.
- Updated test cases for the same.

**Preveiw**

image


skoudoro (approved):
ok, merging

PR #1043, opened 2025-10-14
UI: Cleaning UI Backward Compatibility Code

ganimtron-10:
As per out last discussion, we will be dropping the v1 UI backward compatibility as it introduces complexity and brings a high risk of error. This PR cleans that compatibility code.
Will add the migration guide along with the Panel2D PR.

PR #1045, opened 2025-10-18
BF: Rectangle2D warning for np.divide fixed and compatibility with 0.14.0 Pygfx version

maharshi-gor:
## BF: Rectangle2D warning for np.divide fixed and compatibility with Pygfx 0.14.0.

**PR Contents**
- Fixes np.divide warning for Rectangle2D ui.
- Defaults to have (100, 100) size for Rectangle.
- Updates test cases.

- Pygfx has introduced a wrapper dependency called render canvas (Maintained by them only) which introduces a layer between direct canvas and window system to maintain similar apis across.
- We are not required to add this explicitly in the dependencies as this is a default dependency of wgpu-py.

### DIFF
--- a/fury/optpkg.py
+++ b/fury/optpkg.py
@@ -2,6 +2,8 @@
 
 import importlib
 
+from packaging.version import Version
+
 try:
     import pytest
 except ImportError:
@@ -123,22 +125,25 @@ class TripWire:
         raise TripWireError(self._msg)
 
 
-def optional_package(name, *, trip_msg=None):
+def optional_package(name, *, trip_msg=None, min_version=None):
     """Return package-like thing and module setup for package `name`.
 
     Parameters
     ----------
     name : str
-        Package name to be imported.
-    trip_msg : None or str, optional
+        Package name.
+    trip_msg : None or str
         Message to give when someone tries to use the return package, but we
         could not import it, and have returned a TripWire object instead.
-        Default message if None.
+    min_version : None or str
+        If not None, require that the imported package be at least this
+        version.  If the package has no ``__version__`` attribute, or if the
+        version is not parseable, raise an error.
 
     Returns
     -------
     pkg_like : module or ``TripWire`` instance
-        If we can import the package, return it. Otherwise return an object
+        If we can import the package, return it.  Otherwise return an object
         raising an error when accessed.
     have_pkg : bool
         True if import for package was successful, false otherwise.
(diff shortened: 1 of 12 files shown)


fury/ui/core.py
```diff
@@ -656,7 +656,9 @@ class Rectangle2D(UI):
     """
 
     @warn_on_args_to_kwargs()
-    def __init__(self, *, size=(0, 0), position=(0, 0), color=(1, 1, 1), opacity=1.0):
+    def __init__(
+        self, *, size=(100, 100), position=(0, 0), color=(1, 1, 1), opacity=1.0
```
ganimtron-10, line 660: I was thinking why don't we make this `size` property a required param?
Similar to the `outer_radius` in `Disk2D`.

I have added this required size feature for Rect in this PR #1043, Let me know which approach we should move ahead so I can update it accordingly

maharshi-gor: I will keep it as it is and not make it required param. As people might confuse the values with position if not passed as key word argument.

And if we even feel like doing it. I think let's create a separate PR to fix the other things.

fury/optpkg.py
```diff
@@ -150,21 +155,31 @@ def optional_package(name, *, trip_msg=None):
     --------
     Typical use would be something like this at the top of a module using an
     optional package:
-    >>> from fury.optpkg import optional_package
+
+    >>> from dipy.utils.optpkg import optional_package
```
skoudoro: Hi @maharshi-gor, 

I was going to merge but can you fix this issue on the CI:

```
150     module_setup : function
151         Callable usually set as ``setup_module`` in calling namespace, to allow
152         skipping tests.
153 
154     Examples
155     --------
156     Typical use would be something like this at the top of a module using an
157     optional package:
158 
159     >>> from dipy.utils.optpkg import optional_package
UNEXPECTED EXCEPTION: ModuleNotFoundError("No module named 'dipy'")
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/doctest.py", line 1350, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest fury.optpkg.optional_package[0]>", line 1, in <module>
ModuleNotFoundError: No module named 'dipy'
/Users/runner/work/fury/fury/venv/lib/python3.10/site-packages/fury/optpkg.py:159: UnexpectedException
```

skoudoro (commented):
Almost ready to go, can you fix the comment below

skoudoro:
All good, merging, Thanks @maharshi-gor

PR #1046, opened 2025-10-27
NF: Event type drag created.

maharshi-gor:
### NF: Event type drag created.

**PR Contents**
- Supports dragging from the renderer level.
- Will help drag and draw things.
- Updates UI based element dragging when go out of bounds.

**Fixes**
- Fixes the viz_shapes.py issue for dragging.
- Try `viz_shapes.py` example to test the workings.

### DIFF
--- a/docs/examples/viz_slicer.py
+++ b/docs/examples/viz_slicer.py
@@ -22,6 +22,7 @@ intensity value and use arrow keys to navigate through the slices.
 import numpy as np
 from fury import actor, window
 from fury.utils import get_slices, show_slices
+from fury.lib import EventType
 from dipy.data import read_mni_template
 
 ###############################################################################
@@ -80,7 +81,7 @@ def handle_key_event(event):
 # Add event handlers to the slice actor for key events.
 
 show_m = window.ShowManager(scene=scene, title="FURY 2.0: Cube Slicer Example")
-show_m.renderer.add_event_handler(handle_key_event, "key_down")
+show_m.renderer.add_event_handler(handle_key_event, EventType.KEY_DOWN)
 
 
 ################################################################################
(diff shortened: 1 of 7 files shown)


fury/ui/core.py
```diff
@@ -315,7 +315,7 @@ def handle_events(self, actor):
         """
         actor.add_event_handler(self.mouse_button_down_callback, "pointer_down")
         actor.add_event_handler(self.mouse_button_up_callback, "pointer_up")
-        actor.add_event_handler(self.mouse_move_callback, "pointer_move")
+        actor.add_event_handler(self.mouse_move_callback, "pointer_drag")
```
ganimtron-10: Lets use the `EventType.POINTER_DOWN` and other eventtype instead of hard coding as we did in `window.py`

ganimtron-10 (changes requested):
Hi @maharshi-gor ,

I tested changes locally and the drag feels a lot better now.

Just had a quick question
Will we be using this drag only for UI elements or do we have plan for something else too?
I was just asking to plan and see how `hot_ui` and `active_ui` work with this later on.

fury/ui/core.py
maharshi-gor: You mean `EventType.POINTER_DRAG` and not `EventType.POINTER_DOWN`.

maharshi-gor:
> Hi @maharshi-gor ,
> 
> I tested changes locally and the drag feels a lot better now.
> 
> Just had a quick question Will we be using this drag only for UI elements or do we have plan for something else too? I was just asking to plan and see how `hot_ui` and `active_ui` work with this later on.

We do plan to create new actor based on dragging for other tools like dipy. So, we need it for all the actor and renderer level to understand those kind of callbacks.

fury/ui/core.py
ganimtron-10: I meant using the below things instead of harcoded values

```
actor.add_event_handler(self.mouse_button_down_callback, EventType.POINTER_DOWN)
 actor.add_event_handler(self.mouse_button_up_callback, EventType.POINTER_UP)
 actor.add_event_handler(self.mouse_move_callback, EventType.POINTER_DRAG)
```

ganimtron-10:
As you are updating all the EventType harcoded values, Please also update the EventType at the below mentioned lines too for consistency.

windows.py:776, 581
viz_vector_field:121
viz_slicer:129, 132, 83

docs/examples/viz_slicer.py
```diff
@@ -80,7 +81,7 @@ def handle_key_event(event):
 # Add event handlers to the slice actor for key events.
 
 show_m = window.ShowManager(scene=scene, title="FURY 2.0: Cube Slicer Example")
-show_m.renderer.add_event_handler(handle_key_event, "key_down")
+show_m.renderer.add_event_handler(handle_key_event, EventType.KEY_DOWN)
```
skoudoro: it would be great to handle both

skoudoro (approved):
All good, merging

PR #1048, opened 2025-10-31
BF: Jupyter integration working.

maharshi-gor:
### BF: Jupyter integration working.

**PR Contents**
- Jupyter required the canvas separately to display.
- Uses Ipython display function to not requiring to specify the canvas.

**NOTE**

- This allows to run the fury in jupyter without explicit code writing.
- The only thing required is to provide the `window_type="jupyter"` parameter to `show` method of `ShowManager`

### DIFF
--- a/fury/lib.py
+++ b/fury/lib.py
@@ -33,6 +33,12 @@ jupyter_rfb, have_jupyter_rfb, _ = optional_package(
 if have_jupyter_rfb:
     from rendercanvas.jupyter import RenderCanvas as JupyterWgpuCanvas
 
+IPython, have_ipython, _ = optional_package("IPython")
+if have_ipython:
+    from IPython.display import display as display_jupyter_widget
+else:
+    display_jupyter_widget = IPython
+
 qt_pckg_msg = (
     "You do not have any qt package installed. The qt window will not work for "
     "you. Please install or upgrade any of PySide6, PyQt6, PyQt5, PySide2 "
@@ -145,7 +151,6 @@ Clock = gfx.Clock
 Stats = gfx.Stats
 
 plane_geometry = gfx.plane_geometry
-
 if have_jupyter_rfb:
     JupyterCanvas = JupyterWgpuCanvas
 else:
--- a/fury/window.py
+++ b/fury/window.py
@@ -38,6 +38,7 @@ from fury.lib import (
     UIRenderer,
     Viewport,
     call_later,
+    display_jupyter_widget,
     get_app,
     qcall_later,
     run,
@@ -1106,6 +1107,9 @@ class ShowManager:
             self._qt_app.exec()
         else:
             run()
+
+        if self._window_type == "jupyter":
+            display_jupyter_widget(self.window)
 
     def close(self):
         """Close the rendering window and terminate the application if necessary."""


maharshi-gor:
Hello @m-agour and @ganimtron-10 Please do provide your thoughts on this PR.

fury/lib.py
```diff
@@ -21,6 +21,12 @@
 if have_jupyter_rfb:
     from rendercanvas.jupyter import RenderCanvas as JupyterWgpuCanvas
 
+IPython, have_ipython, _ = optional_package("IPython")
+if have_ipython:
+    from IPython.display import display as display_jupyter_widget
+else:
+    display_jupyter_widget = IPython
```
m-agour, line 40: This will cause runtime error if IPython (Optional dependency) is not installed, please handle the error here.

maybe implement a shell: 
```python
else:
    def display_jupyter_widget(*args, **kwargs):
        pass
```
and handle the error here it's used.

fury/window.py
```diff
@@ -958,6 +959,9 @@ def start(self):
         else:
             run()
 
+        if self._window_type == "jupyter":
+            display_jupyter_widget(self.window)
```
m-agour: Hnadle Error here.

m-agour: YOu can also import have_jupyter_rfb from fury.lib

m-agour (changes requested):
Thanks @maharshi-gor for this PR, Can you please add tests for this so that we know if something breaks again. Also, I left you some comments below:

fury/lib.py
maharshi-gor: It is exactly what we want, as it will create a Tripwire Error Due to The line no 24. Which will tell them to install Ipython. 

Also if someone is trying to use jupyter, IPython is installed by default.

maharshi-gor:
Hello @m-agour ,

Thank you for the review. I have update the changes as requested.

fury/lib.py
skoudoro: this looks good to me

fury/tests/test_window.py
```diff
@@ -340,12 +374,11 @@ def test_show_manager_with_empty_config():
...
 
 
-def test_display_default(sample_actor):
+def test_display_default(sample_actor, monkeypatch):
     """Test the display function with default parameters."""
-
-    with patch("fury.window.ShowManager") as mock_show_manager:
-        show([sample_actor])
-        mock_show_manager.assert_called_once()
+    monkeypatch.setattr("fury.window.ShowManager", MockShowManager)
```
skoudoro: Not fan of monkeypatch. Most of the time, they hide issue since too flexible. Can you find another alternative to test this ?

fury/window.py
skoudoro: agree

skoudoro (commented):
Hi @maharshi-gor,

Thank you for this. Tests are failing, why is it not addressed ?
`FAILED ../venv/lib/python3.11/site-packages/fury/tests/test_window.py::test_display_jupyter_widget - ValueError: renderer.pixel_scale must be bwteen 0.1 and 10.`

skoudoro:
All good, merging

PR #1050, opened 2025-11-04
MTN: Analyze snapshot

maharshi-gor:
### MTN: Analyze Snapshot enabled.

**PR Contents**
- Can be used to verify what is displayed in the test.
- Updated the snapshot test cases to analyze.
- Added a test case for the method.

PR #1051, opened 2025-11-05
TEST: added visual testing system

m-agour:
Added Snapshots analysis system.

Does this mean everything tested is working correctly? No, tests does not cover everything yet, this is just a starting point. 
How to proceed from here? whenever we see a bug we apply the visual test and see if the test can't detect the bug we can change the functionality of the snapshot tester and/or add more methods for visual analysis. So, we need to see if everything is workign as expected and before we fix we make this testing syste able to detect it first to prevent future bugs.

### DIFF
--- a/fury/testing.py
+++ b/fury/testing.py
@@ -2,7 +2,6 @@
 
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from distutils.version import LooseVersion
 from functools import partial
 import io
 import json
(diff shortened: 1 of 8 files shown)


skoudoro:
Are @maharshi-gor and you working on the same thing without communication ?

see #1050

I am lost

m-agour:
@skoudoro you are right Serge I didn't see maharshi's PR. they are almost the same. but here I try to use it to detect the errors we already have in fury and we test shading here which is crucial for testing the shaders parts. @maharshi-gor lets talk about this tomorrow.

skoudoro:
Did you and @maharshi-gor  talked ? 

What is the final conclusion ?

PR #1052, opened 2025-11-09
UI: Adding TextBlock2D

ganimtron-10:
Adding TextBlock2D ie. Text with a background (Rectangle2D) 

image


fury/ui/core.py
```diff
@@ -849,570 +845,510 @@ def inner_radius(self, radius):
...
+            left, right or center.
+        vertical_justification : str
+            bottom, middle or top.
+        bold : bool
+            Makes text bold.
+        italic : bool
+            Makes text italicised.
+        shadow : bool
+            Adds text shadow.
+        size : (int, int)
```
skoudoro: missing, all the `, optional`

fury/ui/core.py
```diff
@@ -849,570 +845,510 @@ def inner_radius(self, radius):
...
+        Parameters
+        ----------
+        text : str
+            The initial text while building the actor.
+        position : (float, float)
+            (x, y) in pixels.
+        color : (float, float, float)
+            RGB: Values must be between 0-1.
+        bg_color : (float, float, float)
+            RGB: Values must be between 0-1.
```
skoudoro: I forgot, do we do 0-1 for the whole codebase or 0-255. something to check

0-1 can be hard for the user

skoudoro (changes requested):
Overall, it looks good, I wonder why bold, italic, shadow, color, are still under comment ?

Do you plan to add those feature in this PR ? if not, you need to remove them from init, it can be confusing

ganimtron-10:
Hi @skoudoro,

> Overall, it looks good, I wonder why bold, italic, shadow, color, are still under comment ?
> Do you plan to add those feature in this PR ? if not, you need to remove them from init, it can be confusing

Yes I plan to add those, just I am exploring how to do it efficiently so that partial and full texts can be formatted as pygfx supports these formatting only through markdown format.

fury/ui/core.py
ganimtron-10: Will look and update it accordingly to use 0-255 range

maharshi-gor:
Hello @ganimtron-10 

Thanks for the work.

I tried the example. On macOS when I load the tutorial for the first time the text goes out of the boxes. But the the moment I resize, it becomes correct.

Screenshot 2025-12-26 at 1 15 15 PM

This is the screenshot of the first load.

On understanding more the text size creates issue in the mac because of the screen resolution as it is not hd the fonts and size goes beyond the size of the window opened.

ganimtron-10:
Hi @maharshi-gor ,
Thanks for testing this.

I am still not clear on what might be happening, can you answer these below questions? 

On resize, does the text shrinks to match background size or the background scales to match text ?

If you replace https://github.com/ganimtron-10/fury/blob/07951d82f8d52b1fc6b3ac77b075c6d8cc2f7102/fury/ui/core.py#L981-L984 these four line with just `self.update_bounding_box(size=size)` does it resolves?

maharshi-gor:
Hi @ganimtron-10 

> On resize, does the text shrinks to match background size or the background scales to match text ?

There are multiple things happening. 
1. For the top sections the pink background extends to cover most of the things.
2. Left side, the text is not resizing but moving to the box.
3. Right side, the box is adjusting to the size of the text.

> If you replace https://github.com/ganimtron-10/fury/blob/07951d82f8d52b1fc6b3ac77b075c6d8cc2f7102/fury/ui/core.py#L981-L984 these four line with just self.update_bounding_box(size=size) does it resolves?

No it does not resolved the issue.

Thanks

ganimtron-10:
Hi @maharshi-gor,
Please test and let me know if there's anything pending to go, I have tested it on Windows and it works fine.

Thanks!

maharshi-gor:
Hi @ganimtron-10 I tried it on the macbook which giving me issues. It works fine now. I will review the PR and let you know.

docs/examples/viz_textblock.py
```diff
@@ -0,0 +1,104 @@
+"""
+===========
+TextBlock2D
```
maharshi-gor: Provide a better title for the tutorial.

fury/ui/core.py
```diff
@@ -887,570 +884,500 @@ def inner_radius(self, radius):
...
+        (float, float, float)
+            Returns text color in RGB.
+        """
+        return self.actor.material.color[:3]
 
-#     @background_color.setter
-#     def background_color(self, color):
-#         """Set text color.
+    @color.setter
+    def color(self, color=(1, 0, 0)):
```
maharshi-gor: Please update this color should not have a default value.

Put a check below. If passed None make it (1, 0, 0)

fury/window.py
```diff
@@ -1078,7 +1087,11 @@ def _draw_function(self):
             self._stats = Stats(self.renderer)
             self._stats_initialized = True
 
-        render_screens(self.renderer, self.screens, stats=self._stats)
+        self._frame_count += 1
```
maharshi-gor, line 1090: Can you explain how the _frame_count works to decide if the layout needs adjustment?

maharshi-gor: Also we never set it back to 0? Does that mean it needs to happen only once?

maharshi-gor (commented):
Hi @ganimtron-10  Thanks for the work. Please look at some of the questions and changes requested.

fury/lib.py
```diff
@@ -142,6 +142,7 @@ def get_app():
 call_later = loop.call_later
 Clock = gfx.Clock
 Stats = gfx.Stats
+Text = gfx.Text
```
maharshi-gor: remove this

fury/ui/core.py
```diff
@@ -9,6 +9,7 @@
 from fury.geometry import buffer_to_geometry
 from fury.lib import (
     EventType,
+    Text,
```
maharshi-gor: import it from fury.actor

fury/window.py
ganimtron-10: To recalculate the layout on the 2nd frame, we use this `_frame_count`  as of now as a workaround.

fury/ui/core.py
ganimtron-10: Why are we tightly integrating with pygfx here and using it directly in our core actors?
If we might plan to add support to more backend wouldnt this cause an issue?

ganimtron-10:
Hi @maharshi-gor ,
I have updated the PR according to the review PTAL.
Thanks!

fury/ui/core.py
maharshi-gor: Hey @ganimtron-10 

Yes and No, so `actor.core` is the only file. Also the Text you will be importing is our Text not pygfx Text. We wanted to support a flexible way to use all the apis separately available in material, geometry. Also we need a smarter api for handling animation and events like we do in UI will just `actor.<API>`.

Hope that makes sense.

fury/window.py
maharshi-gor: Let's create an issue to keep the reminder to the same?

maharshi-gor:
Thanks for all the work @ganimtron-10 🚀 

LGTM, Merging!

Issue #469, opened 2021-07-24
Mismatch in parameter and docstring in manifest_standard() in material module


SunTzunami:
Parameter `interpolation` is set to string [here](https://github.com/fury-gl/fury/blob/master/fury/material.py#L34) but in the [docstring](https://github.com/fury-gl/fury/blob/master/fury/material.py#L52-L54), it says that it's a float.

SunTzunami:
Also, should we have option of more types of interpolations other than Phong too? Other types of interpolation supported by vtk can be found [here](https://vtk.org/doc/nightly/html/classvtkProperty.html#a5c95702d226c4f4783957b7f3a98c069).

SunTzunami:
Another thing, parameter `specular_power` is set to 10 [here](https://github.com/fury-gl/fury/blob/master/fury/material.py#L33) but the according to the [docstring](https://github.com/fury-gl/fury/blob/master/fury/material.py#L51) it's supposed to be between 0 and 1.

skoudoro:
Thank you for the feedback @SunTzunami!

Can you address this @guaje in a quick PR?

guaje:
@SunTzunami thanks for catching this. I'm already working on the corrections on the documentation.

Regarding other types of interpolation, that is also a work in progress, and those features will come in other PRs.

skoudoro:
Fixed by #488, closing

PR #1053, opened 2025-11-10
NF: Contour from ROI and Contour from Volume Introduced.

maharshi-gor:
### NF: Contour from ROI and Contour from Volume Introduced.

**PR Contents**
 Added function in utils to create a mesh from volume.
- Supports multiple objects in the volume.
- Requires an implementation bio-medical specific actor.
- Test cases for all the actors and utils function.
- Tutorial to try various combinations from volume.
- Fixed small bugs from surface actor.

### DIFF
--- a/fury/actor/bio.py
+++ b/fury/actor/bio.py
@@ -2,14 +2,17 @@
 
 import numpy as np
 
-from fury.actor import (
+from fury.actor.slicer import data_slicer, vector_field_slicer
+from fury.actor.surface import contour_from_volume
+from fury.lib import Group
+from fury.utils import (
     apply_affine_to_actor,
     apply_affine_to_group,
     data_slicer,
+    get_transformed_cube_bounds,
     show_slices,
     vector_field_slicer,
 )
-from fury.utils import get_transformed_cube_bounds
 
 
 def volume_slicer(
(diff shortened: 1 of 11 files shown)


fury/actor/bio.py
```diff
@@ -129,3 +130,38 @@ def peaks_slicer(
...
+):
+    """Create contour for region of interest in a volume data.
+
+    Parameters
+    ----------
+    data : ndarray
+        The 3D volume data of shape (X, Y, Z).
+    affine : ndarray, optional
+        The affine transformation matrix to apply to the contour.
+    color : tuple, optional
```
skoudoro, line 148: the output color ? not clear

fury/actor/bio.py
```diff
@@ -129,3 +130,38 @@ def peaks_slicer(
         obj.cross_section = np.asarray(bounds).mean(axis=0)
 
     return obj
+
+
+def contour_from_roi(
```
skoudoro, line 137: function name can be confusing since the input data does not seems to be a roi.

skoudoro (changes requested):
Hi @maharshi-gor,

I need to play a bit with this but here the initial thoughts:

- the function inside `voxel_mesh_by_object` should be out, because they can be util.
- please, use more numpy, some operations can be avoided instead of a for loop

fury/actor/bio.py
maharshi-gor: I kept the old name. Should I update the variable name or the function name @skoudoro ?

skoudoro: when you look at the old name and old function, look at this line:

https://github.com/fury-gl/fury/blob/792f89c8f09b840839a1a9137897b16b3d7678bd/fury/actor.py#L444

and look at this line : https://github.com/fury-gl/fury/blob/792f89c8f09b840839a1a9137897b16b3d7678bd/fury/actor.py#L468-L470

your function have the same name but seems to do something different and expect a different input. 
I need to introspect. Please, look at the old function also

maharshi-gor: Hey @skoudoro 

I looked at the old function before, it was assuming that there is only one region of interest in the ROI. But the new object finds different regions not connected with each other.

The line #469 is scaling the labels to color range so the vtk can process them for resliceing.

The object relies mostly on vtk to do things for on the contrary we are doing things in pythonic way.

I can change the docstrings as they might be confusing.

Let me know if this makes sense.

fury/utils.py
```diff
@@ -740,3 +772,207 @@ def get_transformed_cube_bounds(affine_matrix, vertex1, vertex2):
     max_vals = np.max(transformed_vertices, axis=0)
 
     return [min_vals, max_vals]
+
+
+def extract_surface_voxels(volume, label_value, *, structuring_element):
```
skoudoro: `=None`

fury/utils.py
```diff
@@ -740,3 +772,207 @@ def get_transformed_cube_bounds(affine_matrix, vertex1, vertex2):
...
+    """Extract boundary voxel coordinates for a label within a volume.
+
+    Parameters
+    ----------
+    volume : ndarray
+        Labeled volume containing the objects of interest; can be a full volume
+        or a cropped sub-section.
+    label_value : int
+        The label identifying the object whose surface voxels should be returned.
+    structuring_element : ndarray
```
skoudoro: `, optional`

fury/utils.py
```diff
@@ -740,3 +772,207 @@ def get_transformed_cube_bounds(affine_matrix, vertex1, vertex2):
...
+    Returns
+    -------
+    tuple or None
+        Returns (surface_coords, object_mask) when a surface is found, where
+        surface_coords has shape (N, 3) ordered as (x, y, z). Returns None when
+        the label does not have exposed voxels.
+    """
+    object_mask = volume == label_value
+
+    surface_mask = object_mask & ~binary_erosion(
```
skoudoro: - `not`
- `&` vs`&&`
- clarify please

fury/utils.py
```diff
@@ -740,3 +772,207 @@ def get_transformed_cube_bounds(affine_matrix, vertex1, vertex2):
...
+        raise ValueError("spacing must be a tuple of 3 elements.")
+
+    if triangulate not in (True, False):
+        raise ValueError("triangulate must be a boolean value.")
+
+    struct = generate_binary_structure(rank=3, connectivity=connectivity)
+    labels, n = label(volume > 0, structure=struct)
+
+    if n == 0:
+        logging.info("No objects found in the volume.")
```
skoudoro, line 755: Todo later, uniform logger. sometimes print sometimes log

fury/utils.py
```diff
@@ -740,3 +772,207 @@ def get_transformed_cube_bounds(affine_matrix, vertex1, vertex2):
...
+        ny = neighbors[..., 1]
+        nz = neighbors[..., 2]
+
+        Z, Y, X = object_mask.shape
+        valid = (nx >= 0) & (nx < X) & (ny >= 0) & (ny < Y) & (nz >= 0) & (nz < Z)
+
+        occupied = np.zeros_like(valid, dtype=bool)
+        valid_idx = np.where(valid)
+        occupied[valid_idx] = object_mask[nz[valid_idx], ny[valid_idx], nx[valid_idx]]
+        exposed = ~occupied
```
skoudoro: same

skoudoro:
Can you fix the conflict here ? Thanks !

skoudoro (approved):
Thank you for this @maharshi-gor,

merging

PR #1054, opened 2025-11-10
FPS support (setting and displaying)

m-agour:
FPS was limited to 30 all this time! now we set it to 60 and added the ability to show the FPS (in the console). Both arguments are added to the ShowManager.

```python
ShowManager(scene=scene, title="FPS", show_fps=True, max_fps=60)
```

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -398,8 +398,7 @@ def render_screens(renderer, screens, stats=None):
     screens : list of Screen
         The list of Screen objects to render.
     stats : Stats, optional
-        Optional pygfx Stats helper to time rendering. If provided, it will
-        wrap the rendering cycle and display FPS overlay after rendering all screens."""
+        Stats helper to display FPS overlay."""
     if stats is not None:
         stats.start()
 
@@ -510,7 +509,7 @@ class ShowManager:
     qt_parent : QWidget
         An existing QWidget to embed the QtCanvas within (if `window_type` is 'qt').
     show_fps : bool
-        Whether to print FPS statistics in the console.
+        Whether to display FPS statistics using an on-screen overlay.
     max_fps : int
         Maximum frames per second for the canvas.
     """
@@ -595,12 +594,8 @@ class ShowManager:
         self._max_fps = max_fps
         self._window_type = self._setup_window(window_type)
 
-        # For offscreen rendering, enable console FPS printing in the renderer
-        # For interactive windows, we'll use visual Stats overlay instead
-        use_renderer_fps = self._show_fps and self._window_type == "offscreen"
-
         if renderer is None:
-            renderer = Renderer(self.window, show_fps=use_renderer_fps)
+            renderer = Renderer(self.window)
         self.renderer = renderer
         self.renderer.pixel_ratio = pixel_ratio
         self.renderer.add_event_handler(
@@ -621,7 +616,6 @@ class ShowManager:
         )
         self._callbacks = {}
 
-        # Stats will be created lazily on first render to avoid initialization issues
         self._stats = None
         self._stats_initialized = False
 
@@ -979,14 +973,7 @@ class ShowManager:
 
     def _draw_function(self):
         """Draw all screens and request a window redraw."""
-        # Lazy initialize stats on first render if requested
-        # Skip stats for offscreen (crashes)
-        should_init_stats = (
-            self._show_fps
-            and not self._stats_initialized
-            and self._window_type != "offscreen"
-        )
-        if should_init_stats:
+        if self._show_fps and not self._stats_initialized:
             self._stats = Stats(self.renderer)
             self._stats_initialized = True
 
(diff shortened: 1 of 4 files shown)


fury/window.py
```diff
@@ -563,17 +569,23 @@ def __init__(
         qt_parent : QWidget, optional
             An existing QWidget to embed the QtCanvas within (if `window_type`
             is 'qt').
+        show_fps : bool, optional
+            Whether to display FPS statistics in the renderer. Defaults to False.
+        max_fps : int, optional
+            Maximum frames per second for the canvas. Defaults to 60.
```
maharshi-gor: Remove Default value

fury/window.py
```diff
@@ -656,16 +668,25 @@ def _setup_window(self, window_type):
             )
```
maharshi-gor, line 682: Let's add a check if more than 60 reset to 60?

maharshi-gor (changes requested):
Thanks for the work @m-agour 

Please check the comments and CIs

m-agour:
Thanks @maharshi-gor, will address the review, but, the CI is a global problem with the CI itself. This PR is not the reason. Can you check the reason please?

fury/window.py
m-agour: Sometimes we want it more than 60 if the monitor support higher framerate.

maharshi-gor:
Hi @m-agour 

I tested on my machine, I was never able to get more than 50 FPS. 

I was able to bound it below the max_fps number which was great.

Do you mind adding the stats box https://docs.pygfx.org/stable/_gallery/feature_demo/gltf_animations.html#sphx-glr-gallery-feature-demo-gltf-animations-py

You can take inspiration from here!

m-agour:
Hi @maharshi-gor I was already working on showing the stats box and it's working!  nOW for the 50 FPS I tested it on Wndows and Linux, Nvidia and intel, they both show 59 which is perfect , it's hard to achieve 60 exact wihout v-sync enabled so 59 is same as 60 here.

m-agour:
it causes crashes in the off-screen showm so, I disabled it for offscreen.  but, it's not a big deal since it's more useful in interactive debugging.

fury/window.py
```diff
@@ -563,17 +580,27 @@ def __init__(
...
+        self._max_fps = max_fps
         self._window_type = self._setup_window(window_type)
 
+        # For offscreen rendering, enable console FPS printing in the renderer
+        # For interactive windows, we'll use visual Stats overlay instead
+        use_renderer_fps = self._show_fps and self._window_type == "offscreen"
+
         if renderer is None:
-            renderer = Renderer(self.window)
+            renderer = Renderer(self.window, show_fps=use_renderer_fps)
```
maharshi-gor: `use_renderer_fps` does not make sense as we will not be able to see it. The moment I move it offscreen it will open and shut. You will not see any print on the terminal.

Let's not do anything in the case of `offscreen window_type`.

fury/window.py
```diff
@@ -387,20 +388,30 @@ def update_viewports(screens, screen_bbs):
...
+
     for screen in screens:
         scene_root = screen.scene
         screen.viewport.render(scene_root.main_scene, screen.camera, flush=False)
         screen.viewport.render(scene_root.ui_scene, scene_root.ui_camera, flush=False)
 
+    if stats is not None:
+        stats.stop()
+        stats.render(flush=False)
+
```
maharshi-gor, line 413: Can you help me understand why we need to start and stop on every render?

fury/window.py
```diff
@@ -594,6 +621,10 @@ def __init__(
         )
         self._callbacks = {}
 
+        # Stats will be created lazily on first render to avoid initialization issues
```
maharshi-gor: Remove comments

fury/window.py
```diff
@@ -927,7 +979,18 @@ def snapshot(self, fname):
 
     def _draw_function(self):
         """Draw all screens and request a window redraw."""
-        render_screens(self.renderer, self.screens)
+        # Lazy initialize stats on first render if requested
```
maharshi-gor: remove comments

fury/window.py
```diff
@@ -927,7 +979,18 @@ def snapshot(self, fname):
...
+        # Skip stats for offscreen (crashes)
```
maharshi-gor: Here as well.

maharshi-gor (commented):
Hello @m-agour, Thanks for the quick update.

Please find the some small changes in the code base. Let's make a habit of not including comments in the code before creating a PR.

fury/window.py
m-agour: Good question @maharshi-gor , start and stop here are how the actual frame time measurement. The Stats object here is just a UI element it doesnt internally measure the time start here is when the frame starts rendering and stop is when the frame finishes! this is also how we used to get the fps back in the old days to measure everything in the rendering loop like multiple screens we have.

m-agour:
@maharshi-gor Thanks for the quick review. Addressed all!

maharshi-gor:
Thank you for the work @m-agour, LGTM, merging!

PR #1060, opened 2025-12-11
ImGui introduced in FURY

maharshi-gor:
### NF: Imgui introduced.

- Provides APIs to enable and disable imgui.
- It can be activated from the init or using the API at later point.
- Checks different apis to enable/disable imgui.
- Checks setting the callback for drawing the imgui.
- Introduced a tutorial.


Garyfallidis:
Docs to be fixed in an upcoming PR.

PR #1061, opened 2025-12-12
MTN: Pindown the sphinx version below 9.0.0.

maharshi-gor:
- Ablog has not release the updated version supporting the sphinx v9.
- Once that is out we can lift the pinning of sphinx.


skoudoro (approved):
Thank you @maharshi-gor, merging

PR #1064, opened 2025-12-17
BF: Blend mode and depth updated in material

maharshi-gor:
### Opacity Issue Fix

This PR solves the issue of elements hiding in the background and opacity just making them darker rather than making them translucent.


skoudoro (approved):
Looks good to me. Thank you @maharshi-gor. merging

PR #1065, opened 2025-12-19
REL: Release preparation for v2.0.0a4

maharshi-gor:
- Updated examples to show docs properly.
- Updated `prepare_release.py` to work with windows.

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -18,7 +18,7 @@ jobs:
 
     steps:
     - name: Check out repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@v6
 
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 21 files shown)


skoudoro (changes requested):
Can you add `sphinx_gallery`, `sphinx_copybutton`, `pydata_sphinx_theme` for doc in
the pyproject file

`aiohttp`  is required also in fetcher, When did we add this dependency ? we might need to remove it. I do not think we need it as a core dependency

`imgui_bundle` works for python 3.12+ for me. python 3.12 might be our minimal version. We need to discuss that and think about it

Some examples like `viz_contour.py` and `viz_slicer.py` needs vtk. So we might need to add it to doc in pyproject also.

skoudoro (approved):
All good, merging

PR #1069, opened 2026-01-08
BF: Screen Config can be bounding box.

maharshi-gor:
- Fixed the calculate screen size to respect bbs.
- Added resize_callback for downstream.
- callback will be executed first and then regular resize functionality.

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -11,6 +11,7 @@ from dataclasses import dataclass
 from functools import reduce
 import logging
 import os
+import sys
 
 from PIL.Image import fromarray as image_from_array
 import numpy as np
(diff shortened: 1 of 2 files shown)


fury/window.py
```diff
@@ -457,6 +458,13 @@ def calculate_screen_sizes(screens, size):
     if screens is None or not screens:
         return [(0, 0, *size)]
 
+    if isinstance(screens[0], (tuple, list)):
```
skoudoro: I did not look at the rest of the codebase, I need to check
 but why `screens[0]` and not `screens[1]`. I am sure there is a logic but I need to go deeper

maharshi-gor: The screens[0] can be int in the case of basic configuration like 
screens = [1, 2, 4]
Here each element means how many sections along the width and the number itself means how many sub-sections along the height in that particular section.

The more not used way is to provide the bounding boxes in which

screen = [(x0, y0 w0, h0), (x1, y1, w1, h1)]

Given that we can check all of the screen configuration. That will be bulletproof.

fury/window.py
```diff
@@ -457,6 +458,13 @@ def calculate_screen_sizes(screens, size):
     if screens is None or not screens:
         return [(0, 0, *size)]
 
+    if all(isinstance(screen, (tuple, list)) for screen in screens):
```
skoudoro, line 465: ok, great!

I think it is ready to go but one last comment:

can you update the docstring of screens. It will help to understand. Thank you

skoudoro (approved):
All good, Thanks @maharshi-gor, merging

PR #1070, opened 2026-01-13
NF: Streamline based ROI filtering

maharshi-gor:
### NF: Streamline based ROI filtering

**PR Contents**
- Added the baking strategy to pre-compute filtering.
- Later render will be normal as the regular streamline render.
- Added `StreamlineBakingMaterial` to distinguish between baking and
  regular rendering pass.
- Added the tutorial in valid_examples.
- Showcase roi_origin and mask based coverage.
- Added test cases for actor, material, shader

**Screenshots**
image

image

### DIFF
--- a/fury/actor/curved.py
+++ b/fury/actor/curved.py
@@ -618,7 +618,7 @@ class Streamlines(Line):
         else:
             colors_arr = np.asarray(colors, dtype=np.float32)
 
-        self._roi_origin = self._resolve_roi_origin()
+        self._roi_origin = self._resolve_roi_origin(roi_origin)
 
         self._input_positions_array = positions_arr
         self._input_colors_array = colors_arr
@@ -849,8 +849,8 @@ def streamlines(
 
     Returns
     -------
-        Streamline
-            The created streamline object.
+    Streamline
+        The created streamline object.
     """
     lines_list = (
         [np.asarray(seg) for seg in lines]
--- a/fury/actor/curved.py
+++ b/fury/actor/curved.py
@@ -12,6 +12,7 @@ from fury.lib import (
 )
 from fury.material import (
     StreamlinesMaterial,
+    _StreamlineBakedMaterial,
     _StreamtubeBakedMaterial,
     _create_mesh_material,
     validate_opacity,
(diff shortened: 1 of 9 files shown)


fury/wgsl/streamline_compute.wgsl
```diff
@@ -0,0 +1,152 @@
...
+    let base = out_index * 3u;
+    if (base + 2u >= capacity) {
+        return;
+    }
+    s_out_positions[base] = value.x;
+    s_out_positions[base + 1u] = value.y;
+    s_out_positions[base + 2u] = value.z;
+}
+
+fn write_color(out_index: u32, value: vec4<f32>, capacity: u32) {
```
skoudoro, line 84: duplicated with `write_vertex_color` in `streamtube_compute.wgsl`.  I am pretty sure other function are duplicated. this module is growing. 
is there a smart way to avoid those duplication. those functions could be imported from another file. 

Currently, this will not be sustainable and does not compound

fury/actor/curved.py
```diff
@@ -549,15 +836,37 @@ def streamlines(
...
+        provided, defaults to (0, 0, 0).
     enable_picking : bool, optional
         Whether the streamline should be pickable in a 3D scene.
 
     Returns
     -------
-    Streamline
-        The created streamline object.
+        Streamline
+            The created streamline object.
```
skoudoro: why this tab ?

skoudoro (commented):
Overall, looks good. 

I just feel that we need to improve shader management and composition.

Also, We need to find a way to test this wgsl file alone.

fury/wgsl/streamline_compute.wgsl
maharshi-gor: We have utils.wgsl let me find all these and put all of them in that and remove duplicates

fury/actor/curved.py
maharshi-gor: Typo, will fix it asap

fury/wgsl/streamline_compute.wgsl
maharshi-gor: Hello @skoudoro I researched and there is no way we can do it in the case of the writing to the buffer, which is the case in most of them.

The storage ptr cannot be passed to a function. And the binding name will differ from one to other and that will cause an issue if just keep the same name.

maharshi-gor: https://github.com/gpuweb/gpuweb/discussions/4309

skoudoro (approved):
Thanks @maharshi-gor,

merging.

Note: A follow up PR needs to be done to review shader architecture to avoid duplicated code and allow unittests

PR #1071, opened 2026-01-16
BF: peaks slicer not showing correct center

maharshi-gor:
**PR Context**

The peaks slicer was showing little off from the center when passed an affine.
We were overriding the get_bounding_box api which was internally used by pygfx to do so.

Which created issues in centering the element while rendering. This PR introduces a new variable to store the bounds and works as expected.

Added some test cases as well for the same.


skoudoro (approved):
all good, merging

PR #1074, opened 2026-01-22
RF: Introduced correct opacity attribute

maharshi-gor:
- Updated the `opacity` from a function to an attribute.
- Use the function from utils to make single source for updating the actor.
- Updated the `GfxGroup` in the `lib` to avoid circular dependencies.
- Visibility for a `Group` will also check the length of children in the group and visibility tuple provided.


skoudoro (approved):
Looks good to me, merging

Thanks @maharshi-gor

PR #1075, opened 2026-01-22
BF: Slicer flickering in the nearest interpolation.

maharshi-gor:
**Context**
The slicer used to `flicker` on the `middle-slice` in `nearest` interpolation, after applying the affine on some devices.

**Understanding the issue**
The middle value of the visualization falls on specific value which is exactly on a boundary of two voxels. Thus, it is required to choose from the nearing voxel in the `nearest` interpolation. Which creates conflicts on some devices and alternates between the neighbors based on the rounding of the value.

|--------| < middle value > |--------|

**Why only after applying the affine?**
Now the million dollar question is why not on the regular data and only after applying the affine?

The answer to that lies in how the pygfx calculates the bounding box and center. The bounding box ranges from origin - 0.5 to size - 0.5. Which is in regular cases lands in the decimal range which we never reach in the case of native coordinates.

**Solution**
The solution is pretty simple just push the center to decide on of the other by simply adding a small value to plane equation in slow slices function. This will push just a little so we do not end up in boundary and will not show incorrect data as we will never change the voxel. As boundary conditions are usually due to the index being exact integer.

**Note**
I am sorry for the whole file change as this was configured to be CRLF that I changed to LF. We soon need to implement this change in pre-commit.
mypy does it but just checks if it is valid does not enforce LF or CRLF.

### DIFF
--- a/fury/actor/tests/test_slicer.py
+++ b/fury/actor/tests/test_slicer.py
@@ -36,8 +36,10 @@ def test_valid_4d_data(channels):
     assert len(slicer_obj.children) == 3
     assert all(child.visible for child in slicer_obj.children)
 
-    expected_slices = np.array([shape[0] // 2, shape[1] // 2, shape[2] // 2])
-    np.testing.assert_array_equal(get_slices(slicer_obj), expected_slices)
+    expected_slices = np.array([shape[0] // 2, shape[1] // 2, shape[2] // 2]) + 1e-3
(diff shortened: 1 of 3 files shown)


skoudoro:
conflict to fix

skoudoro (approved):
All good, merging 

Thanks @maharshi-gor

PR #1077, opened 2026-01-26
BF: Background was not getting applied.

maharshi-gor:
- This was introduced when two scenes were added to Base Scene.
- Added test cases.
- Made the snapshot function use the `save_image` from `fury.io` to be consitent.


maharshi-gor:
Not required closing. The basic case work nicely there was some issue in checking

PR #1085, opened 2026-02-04
Fix progress bar to update in-place on Windows terminals

sebrbn:
### Description
On Windows (PowerShell / CMD), `fetch_viz_models()` prints the download progress on a new line for every update instead of updating a single line in-place.

This PR fixes the progress bar rendering by explicitly clearing the previous line before rewriting it, ensuring consistent behavior across terminals.

### Changes
- Clear previously printed progress line before updating
- Ensure newline is printed only once at completion
- Works on Windows, Linux, and macOS terminals

###Screenshot
image

### DIFF
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -54,12 +54,16 @@ class FetcherError(Exception):
     pass
 
 
+_last_line_length = 0
+
+
 def update_progressbar(progress, total_length):
     """Show progressbar.
 
     Takes a number between 0 and 1 to indicate progress from 0 to 100%.
     """
-    # Try to set the bar_length according to the console size
+    global _last_line_length
+
     try:
         if os.name == "nt":
             bar_length = 20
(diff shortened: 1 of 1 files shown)


maharshi-gor:
Hi @sebrbn Thanks for the work. We are moving to v2. Check the branch named v2 and update the PR from  a branch created from v2 to v2.

Again, if you read the GSoC wiki it mentions that we are moving to webgpu and v2 branch is being used currently.

sebrbn:
Hi @maharshi-gor  , thanks for the clarification.

I’ve addressed this issue by fixing the progress bar rendering so that it updates in-place on Windows instead of printing a new line for each update. The solution clears the previously printed line before rewriting and ensures the newline is printed only once at completion.

I’ll submit a new PR created from the v2 branch and target it back to v2, in line with the ongoing WebGPU migration.

Please let me know if any adjustments are needed.

Thanks!

PR #1086, opened 2026-02-04
Fix progress bar to update in-place on Windows terminals

sebrbn:
On Windows terminals , fetch_viz_models() prints the download progress on a new line for each update instead of updating the same line in-place. This results in cluttered and unreadable output during model downloads.

This PR fixes the progress bar behavior by explicitly clearing the previously printed line before writing the updated progress. A newline is printed only once after the download completes, ensuring consistent and clean progress rendering.

#Changes

Clear the previously printed progress line before updating it

Ensure the progress bar updates in-place instead of printing new lines

Print a newline only once after completion

Works consistently across Windows, Linux, and macOS terminals

#Notes

Fix is based on the v2 branch as per the current development direction

No changes to the download logic or public API

### DIFF
--- a/fury/data/fetcher.py
+++ b/fury/data/fetcher.py
@@ -54,12 +54,16 @@ class FetcherError(Exception):
     pass
 
 
+_last_line_length = 0
+
+
 def update_progressbar(progress, total_length):
     """Show progressbar.
 
     Takes a number between 0 and 1 to indicate progress from 0 to 100%.
     """
-    # Try to set the bar_length according to the console size
+    global _last_line_length
+
     try:
         if os.name == "nt":
             bar_length = 20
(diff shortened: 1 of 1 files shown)


fury/data/fetcher.py
```diff
@@ -54,12 +54,16 @@ class FetcherError(Exception):
...
+_last_line_length = 0
+
+
 def update_progressbar(progress, total_length):
     """Show progressbar.
 
     Takes a number between 0 and 1 to indicate progress from 0 to 100%.
     """
-    # Try to set the bar_length according to the console size
+    global _last_line_length
```
maharshi-gor, line 65: Let's remove a global variable.

Return the line_length on this method and pass it back to this method.

That parameter can be None, if None assume that it is getting updated for the first time and nothing is written in the console.

maharshi-gor (changes requested):
Hello @sebrbn Thanks for the work!

Also add some tests to verify this.

Please see the comments.

maharshi-gor:
Hello @sebrbn any updates on this?

PR #1093, opened 2026-02-12
Fix: Handle tuple colors in point() function (#1092)

### LINKED ISSUE #1092: Point actor fails if the colors are not provided.
## Description
From the API description Point actor can work only with centers but, if the colors are not provided it will fail as it assumes the actor to have colors in numpy array and the default is a tuple.

## Error Stack trace
```
  File "C:\Users\magor\devel\grg\tractome\tractome\compute.py", line 383, in filter_streamline_ids
    points = actor.point(np.asarray([min, max], dtype=np.float32))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\magor\AppData\Local\miniforge3\envs\tractome\Lib\site-packages\fury\actor\planar.py", line 389, in point
    colors=colors.astype("float32"),
           ^^^^^^^^^^^^^
AttributeError: 'tuple' object has no attribute 'astype'
```

AdityaGupta716:
Convert colors to numpy array before passing to sphere() function. This fixes the issue where point() fails when colors is passed as a tuple. The downstream code expects colors to have .astype() method which tuples don't have.

Closes #1092

### DIFF
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1825,8 +1825,6 @@ def point(points, colors=None, *, point_radius=0.1, phi=8, theta=8, opacity=1.0)
         opacity=opacity,
     )
 
-
-
 @warn_on_args_to_kwargs()
 def sphere(
     centers,
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1802,18 +1802,17 @@ def point(points, colors, *, point_radius=0.1, phi=8, theta=8, opacity=1.0):
 
     """
 if colors is not None:
-            colors = np.asarray(colors)    
-    
-        return sphere(
-        centers=points,
-        colors=colors,
-        radii=point_radius,
-        phi=phi,
-        theta=theta,
-        vertices=None,
-        faces=None,
-        opacity=opacity,
-    )
+    colors = np.asarray(colors)
+
+return sphere(
+    points,
+    colors=colors,
+    radii=point_radius,
+    phi=8,
+    theta=8,
+    opacity=opacity,
+)
+
 
 
 @warn_on_args_to_kwargs()
--- a/fury/actor.py
+++ b/fury/actor.py
@@ -1767,17 +1767,17 @@ def dot(points, *, colors=None, opacity=None, dot_size=5):
 
     return poly_actor
 
-
+  )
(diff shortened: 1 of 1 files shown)


fury/actor.py
```diff
@@ -1801,6 +1801,7 @@ def point(points, colors, *, point_radius=0.1, phi=8, theta=8, opacity=1.0):
     >>> # window.show(scene)
 
     """
+        colors = np.asarray(colors)
```
maharshi-gor, line 1802: What if someone passes it as None.

Also the formatting is not correct.

maharshi-gor (changes requested):
Please look at the comments provided.

fury/actor.py
AdityaGupta716: @maharshi-gor Done with the changes plz review it again

maharshi-gor: What will happen if it is None,

Please default it to red color.

maharshi-gor: Check if None
create a numpy array with red color.

Check if tuple
convert it to numpy array.

Check if wrong dimension provide error

if it is (3,) and (4,) change it to 2D (1,3), (1,4)
It can be of size (1,3) or (1,4)

It can also be of size(N,3) or (N,4) where n is the number of centers provided.

AdityaGupta716: @maharshi-gor i have redone it according to requirements if color is none it will default to red , tuple and none will convert to an numpy array, and 1d can promote to 2d and also works with n as a point

maharshi-gor:
Hi @AdityaGupta716 I think we made a mistake here. Thanks for the work but we have moved from the master to v2 branch. 

Can you create a PR from a branch created from v2 branch to v2 branch. Thanks for all the efforts. Sorry for not catching it earlier.

The API should look almost the same. But the file and module structure has changed.

AdityaGupta716:
I’ll close this PR and open a new one based on the v2 branch targeting v2, and will adapt the changes to the new file/module structure and it would be nice if u assign the issue to me

PR #1096, opened 2026-02-17
Fix: handle None and tuple colors in point actor (v2)(#1092)

AdityaGupta716:
## Summary

This PR fixes the `actor.point` API on the v2 branch to correctly handle
`colors` when it is `None` or passed as a tuple, and to validate the
expected shapes.

It resolves issue #1092: *"Point actor fails if the colors are not provided."*.

## Changes

- Default `colors` to red `[[1.0, 0.0, 0.0]]` when `colors is None`.
- Convert user-provided `colors` with `np.asarray(colors, dtype=float)` so
  tuples, lists, and arrays are all handled consistently.
- Support the following shapes for `colors`:
  - `(3,)`, `(4,)` (promoted to `(1, 3)` / `(1, 4)`)
  - `(1, 3)`, `(1, 4)` (single color for all points)
  - `(N, 3)`, `(N, 4)` (per-point colors)
- Add validation to ensure:
  - 1D `colors` has 3 (RGB) or 4 (RGBA) channels.
  - 2D `colors` has 3 or 4 channels per point.
  - Number of rows in `colors` is either `1` or matches the number of points.

## Behavior

- `actor.point(pts)` now works when `colors` is omitted and uses red as
  the default color.
- Passing tuples like `(1, 0, 0)` or `(1, 0, 0, 1)` now works without
  raising `AttributeError` from downstream code.
- Invalid shapes (e.g. `(2,)`, `(N, 2)`, `(N, 5)`, or mismatched number
  of rows) raise clear `ValueError` messages.


AdityaGupta716:
@maharshi-gor Done with the changes u asked me to plz review

PR #1101, opened 2026-02-21
 Allow create_text to accept multiple texts and materials fixes #1012

### LINKED ISSUE #1012: Provide Multiple text into one actor.
A text actor should accept multiple text sentences to place at different locations.

Currently, we need to create a new text actor for each separated text which will make things hard to work with. We should be able to provide all the text and all the configuration for the text at once

AdityaGupta716:
Fixes #1012

## Changes

- Updated `text()` in `fury/actor/planar.py` to accept a list of strings,
  returning a `Group` of Text actors
- `create_text()` in `fury/actor/core.py` is unchanged (single string only)
- Added tests in `fury/actor/tests/test_planar.py`

## Usage
```python
# Single text (unchanged behavior)
t = actor.text("Hello")

# Multiple texts, shared position and color
t = actor.text(["Hello", "World", "FURY"])

# Multiple texts with per-item positions
t = actor.text(
    ["Hello", "World", "FURY"],
    position=[(0, 0, 0), (1, 0, 0), (2, 0, 0)],
)

# Multiple texts with per-item colors
t = actor.text(
    ["Hello", "World"],
    colors=[(1, 0, 0), (0, 1, 0)],
)
```

### DIFF
--- a/fury/actor/core.py
+++ b/fury/actor/core.py
@@ -282,6 +282,11 @@ def create_text(text, material, **kwargs):
     ------
     TypeError
         If text is not a string or material is not an instance of TextMaterial.
+
+    Examples
+    --------
+    >>> mat = TextMaterial()
+    >>> t = create_text("Hello", mat)
     """
     if not isinstance(text, str):
         raise TypeError("text must be a string.")
@@ -289,8 +294,7 @@ def create_text(text, material, **kwargs):
     if not isinstance(material, TextMaterial):
         raise TypeError("material must be an instance of TextMaterial.")
 
-    text = Text(text=text, material=material, **kwargs)
-    return text
+    return Text(text=text, material=material, **kwargs)
 
 
 def create_image(image_input, material, **kwargs):
--- a/fury/actor/planar.py
+++ b/fury/actor/planar.py
@@ -7,6 +7,7 @@ import numpy as np
 from scipy.spatial.transform import Rotation as R
 
 from fury.actor import (
+    Group,
     Points,
     actor_from_primitive,
     create_image,
@@ -513,14 +514,17 @@ def text(
 
     Parameters
     ----------
-    text : str or list[str]
-        The plain text to render.
-        The text is split in one TextBlock per line,
-        unless a list is given, in which case each (str) item becomes a TextBlock.
-    colors : ndarray, shape (N, 3) or (N, 4) or tuple (3,) or tuple (4,), optional
-        RGB or RGBA values in the range [0, 1].
-    position : tuple, optional
-        The (x, y, z) coordinates to place the text in 3D space.
+    text : str or list of str
+        The plain text to render. When a list is given, each item becomes
+        a separate text actor and a Group is returned.
+    colors : tuple (3,) or tuple (4,) or list of tuple, optional
+        RGB or RGBA values in the range [0, 1]. When ``text`` is a list,
+        this can be a single color applied to all actors, or a list of
+        colors (one per text item).
+    position : tuple (3,) or list of tuple (3,), optional
+        The (x, y, z) coordinates to place the text in 3D space. When
+        ``text`` is a list, this can be a single position applied to all
+        actors, or a list of positions (one per text item).
     font_size : float, optional
         The size of the font, in object coordinates or pixel screen coordinates.
     family : str, optional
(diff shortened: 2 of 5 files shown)


AdityaGupta716:
@maharshi-gor  I've implemented the changes for this issue. Plz review

fury/actor/core.py
```diff
@@ -262,35 +262,56 @@ def create_point(geometry, material):
...
-    text : str
-        The text content.
-    material : TextMaterial
-        The material object.
+    text : str or list of str
+        The text content or list of text contents.
+    material : TextMaterial or list of TextMaterial
+        The material object or list of material objects.
     **kwargs : dict
         Additional properties like font_size, anchor, etc.
```
maharshi-gor: Please also think about how do you provide all the other keyword arguments in the case of multiple text?

Will the position be same for all of them?

maharshi-gor:
Hi @AdityaGupta716 thanks for the work. Please take a look at the repository structure before submitting a PR.
The text actor testcases are present in `actor\tests\test_core.py` where the new tests should also reside.

maharshi-gor (changes requested):

AdityaGupta716:
@maharshi-gor, the kwargs are fully flexible — pass a single value and it broadcasts to all text objects, or pass a list for per-text control. For example font_size=24 applies to all, while position=[(0,0,0), (1,0,0)] sets each independently. Length mismatches raise a ValueError immediately. Let me know if you need anything else.

fury/actor/core.py
```diff
@@ -262,35 +262,101 @@ def create_point(geometry, material):
...
+    Examples
+    --------
+    >>> mat = TextMaterial()
+    >>> # Single text (unchanged behavior)
+    >>> t = create_text("Hello", mat)
+
+    >>> # Multiple texts, shared material and kwargs
+    >>> t = create_text(["Hello", "World", "FURY"], mat)
+
+    >>> # Multiple texts with individual materials and per-text positions
```
maharshi-gor: Remove comments from examples.

fury/actor/core.py
```diff
@@ -262,35 +262,101 @@ def create_point(geometry, material):
...
+
     if not isinstance(text, str):
         raise TypeError("text must be a string.")
 
     if not isinstance(material, TextMaterial):
         raise TypeError("material must be an instance of TextMaterial.")
 
-    text = Text(text=text, material=material, **kwargs)
-    return text
+    return Text(text=text, material=material, **kwargs)
```
maharshi-gor, line 297: I feel this should not be done here. Please check the text method in the planar.py

https://github.com/fury-gl/fury/blob/8bb525c38ab8f5b37f4489080b3fa6673be88a11/fury/actor/planar.py#L497

This method should be updated to accept multiple text not this method. The text method is the one which people should use.

maharshi-gor (changes requested):
As I suspect the PR needs to update different file and add the test cases at different location according to that.

AdityaGupta716:
hi @maharshi-gor done with the changes ,check it out !

maharshi-gor (commented):
Hey @AdityaGupta716 Thank you for the changes.

Please check the file ending has changed that is forcing to show you entire file as a change instead of the specific changes.

I would also love if you can make a tutorial or update an existing one for example of the API use-case.

fury/actor/planar.py
```diff
@@ -561,6 +573,55 @@ def text(
     >>> show_manager = window.ShowManager(scene=scene, size=(600, 600))
     >>> show_manager.start()
     """
+    if isinstance(text, list):
```
maharshi-gor, line 576: Please also check tuple.

fury/actor/planar.py
```diff
@@ -561,6 +573,55 @@ def text(
...
+        if not all(isinstance(t, str) for t in text):
+            raise TypeError("All items in text list must be strings.")
+
+        n = len(text)
+
+        if isinstance(colors, list):
```
maharshi-gor, line 582: Here as well

fury/actor/planar.py
```diff
@@ -561,6 +573,55 @@ def text(
...
+            if len(colors) != n:
+                raise ValueError(
+                    "Length of colors list must match length of text list."
+                )
+            colors_list = colors
+        else:
+            colors_list = [colors] * n
+
+        if isinstance(position, list):
```
maharshi-gor, line 591: here

maharshi-gor (changes requested):
Thanks @AdityaGupta716 for working on this. It looks good the tutorial is running nicely.

I would recommend fixing tutorial to look good when you run it.

Also please check final round of comments that I have provided.

maharshi-gor (approved):
Thanks for the work @AdityaGupta716 LGTM

Issue #1091, opened 2026-02-10
Update actors to accept individual or ndarray for actor features


ganimtron-10:
**Is your feature request related to a problem? Please describe.**
We allow actors to instantiate multiple instances inside a single one, currently all of those instance share same actor features only differing on position, color, direction and radii not on actor specific features like `opacity`, `height in Cone`, `inner_radius in Ring2D`.

**Describe the solution you'd like**
Actors should accept individual or a ndarray which would specify these actor features different for different instance.

**Additional context**
Ring2D can be a good starting point, refer: https://github.com/fury-gl/fury/pull/1017#discussion_r2219284889

umeshgupta05:
Hi @maharshi-gor ,
I’m interested in working on this issue. Would it be okay to assign it to me so I can start working on a fix?
Thank you!

maharshi-gor:
Hello @umeshgupta05 you can try this!

MAKE SURE YOU USE v2 BRANCH.

So that it will not create any issue further down the line.

umeshgupta05:
Sure @maharshi-gor , Thank you!!

umeshgupta05:
Hi @maharshi-gor,

I have opened a PR for the first step of this issue:

- **PR:** #1102  

In this PR, I implemented per-instance actor-specific support for **Ring2D** (as suggested in the issue), including support for ndarray inputs for:
- `inner_radius`
- `outer_radius`
- `radial_segments`
- `circumferential_segments`

I also added/updated tests for this behavior.

This PR is intentionally scoped to Ring2D first.  
If this approach looks good, I can open a follow-up PR to extend the same pattern to other actors (e.g., cone `height`, etc.).

Thanks!

VanshAgarwal24036:
Hi @maharshi-gor 

I have noticed that previous PR(#1102) attempted to add per-instance geometry parameters for Ring2d actor but closed after review.

I would like to work on this issue. My plan is to:

- allow sclar or ndarrays inputs for parameters like `inner_radius`, `outer_radius`, `radial_segments`, `circumferential_segments`
- validate that the array size either 1 or equals to the number of centers
- keep the existing scalar behaviour unchanged for backward compatibility

I will follow the `v2` branch and try to keep the implementation aligned with the current actor architechture.

Please let me know if this approach looks correct or if you would suggest a different suggestion.

Thanks!

PR #1102, opened 2026-02-21
ENH: support per-instance Ring2D actor features

umeshgupta05:
## Summary
This PR addresses issue #1091 by updating Ring2D to support per-instance actor-specific parameters when creating multiple rings in one actor.

## What changed
- Added support for scalar or ndarray inputs for:
  - `inner_radius`
  - `outer_radius`
  - `radial_segments`
  - `circumferential_segments`
- Preserved existing behavior when scalar values are provided (single shared ring geometry path).
- Added validation for array-size mismatches (must be size 1 or match number of centers).

## Tests
- Added tests for per-instance Ring2D feature inputs.
- Added tests for invalid per-instance input sizes.
- Verified:
  - Ring-focused tests pass.
  - Planar actor test module passes.

## Motivation
Previously, multiple ring instances could differ by position/color/direction/scale, but ring-specific geometry parameters were shared across all instances. This PR enables per-instance geometry control for Ring2D, as discussed in the issue thread.

## Scope
- This PR intentionally focuses on Ring2D as a first step.
- It does not change other actors (e.g., cone height/other per-instance actor-specific properties).

## Branch/Target
- Source branch: `feature-ring2d-instance-features`
- Target branch: `v2`

## Checklist
- [x] Feature implemented
- [x] Tests added/updated
- [x] Existing related tests pass
- [x] Pre-commit checks addressed

### DIFF
--- a/fury/actor/planar.py
+++ b/fury/actor/planar.py
@@ -446,8 +446,7 @@ def marker(
         RGB or RGBA values in the range [0, 1].
     marker : str or MarkerShape, optional
         The shape of the marker.
-        Options are "●": "circle", "+": "plus", "x": "cross", "♥": "heart",
-        "✳": "asterix".
+        Options are "circle", "plus", "cross", "heart", "asterix".
     edge_color : str or tuple or Color, optional
         The color of line marking the edge of the markers.
     edge_width : float, optional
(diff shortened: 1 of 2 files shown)


fury/actor/planar.py
```diff
@@ -748,23 +752,242 @@ def ring(
...
+        outer_radius, value_name="outer_radius", cast=np.float64
+    )
+    radial_segments_arr = _as_per_center(
+        radial_segments, value_name="radial_segments", cast=np.int64
+    )
+    circumferential_segments_arr = _as_per_center(
+        circumferential_segments,
+        value_name="circumferential_segments",
+        cast=np.int64,
+    )
```
maharshi-gor: These should be casted to 32 bit. You might not need to do it as the actor from primitive will do it.

fury/actor/planar.py
```diff
@@ -748,23 +752,242 @@ def ring(
...
+            return arr.item()
+        if arr.ndim == 1 and arr.size == 1:
+            return arr.item()
+        if arr.ndim == 1 and arr.size == 3:
+            return arr
+        if arr.ndim == 1 and arr.size == n_centers:
+            return arr[idx]
+        if arr.ndim == 2 and arr.shape[0] in (1, n_centers) and arr.shape[1] == 3:
+            return arr[0] if arr.shape[0] == 1 else arr[idx]
+        raise ValueError("scales size should be 1 or equal to the numbers of centers")
```
maharshi-gor: I feel these functions are overkill. Let's check if ndarray and of the same size all of them and it should be good.

fury/actor/planar.py
```diff
@@ -748,23 +752,242 @@ def ring(
...
+        all_vertices.append(vertices)
+        all_faces.append(faces + face_offset)
+        color_i = _color_at(colors, idx)
+        all_colors.append(np.repeat(color_i[np.newaxis, :], len(vertices), axis=0))
+
+        face_offset += len(vertices)
+
+    big_vertices = np.concatenate(all_vertices, axis=0)
+    big_faces = np.concatenate(all_faces, axis=0)
+    big_colors = np.concatenate(all_colors, axis=0)
```
maharshi-gor: You are not required to do this as actor_from_primitive merges all the different ones. Scales them.

maharshi-gor (changes requested):
Thanks for the work @umeshgupta05 .

Please look at the comments provide. I feel let's just think to make the code legible for future maintenance.

umeshgupta05:
Hi @maharshi-gor, thanks for the review! I've addressed all your comments, removed the 64-bit casts, replaced the helper functions with a simple dict + size validation loop, and removed the manual scale/rotate/merge logic from the per-instance path. Please take a look!

maharshi-gor:
Thank you for the work.

Hey @umeshgupta05 Please check you have updated the wrong actor.
I strongly recommend self-reviewing the code!

PR #1103, opened 2026-02-21
fix: correct typo in boys2rgb - z4 should be z2*z2 not z*z2 #857

### LINKED ISSUE #857: Wrong boys2rgb colormap
## Description
[Please provide a general introduction to the issue/proposal.]
I think there is a typo in the function boys2rgb

https://github.com/fury-gl/fury/blob/master/fury/colormap.py#L118

it is actually should be:
z4 =  z2 * z2

[If reporting a bug, attach the entire traceback from Python and follow the way to reproduce below]
[If proposing an enhancement/new feature, provide links to related articles, reference examples, etc.]

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [ ] Code example: 
`import numpy as np
from fury import colormap

if __name__ == '__main__':
    v = np.array([[0.0183422, 0.00418003, 0.999823], [-0.0183422, -0.00418003, -0.999823]])
    print(colormap.boys2rgb(v))
`
- [ ] Relevant images (if any)
- [ ] Operating system and versions

AdityaGupta716:
Fixes #857 

The variable `z4` in `boys2rgb` was incorrectly computed as `z * z2` 
(which equals z³) instead of `z2 * z2` (which equals z⁴). 

This can be verified against the original MATLAB code:
https://cagataydemiralp.io/projects/boys/line2rgb.m

### DIFF
--- a/fury/colormap.py
+++ b/fury/colormap.py
@@ -144,7 +144,7 @@ def boys2rgb(v):
     y3 = y * y2
     z3 = z * z2
 
-    z4 = z * z2
+    z4 = z2 * z2
 
     xy = x * y
     xz = x * z


AdityaGupta716:
@maharshi-gor  tiny PR but someone had to do it, Not much to review but plz review.

maharshi-gor:
Thanks for the fix @AdityaGupta716

Issue #971, opened 2025-03-10
vtk rectalinear grid


elmbeech:
Hi All,

I would appreciate if fury-gl, beside the vtk poly data format (vtp), could read vtk rectilinear grid format (vtr).
Attached are two sample files, one from each type (had to add a txt extension to be able to upload).

+ https://vtk.org/doc/nightly/html/classvtkRectilinearGrid.html
+ https://vtk.org/doc/nightly/html/classvtkPolyData.html

This is how I load rectilinear grid and poly data files with the vtk python library: 

```python
import vtk

# load file rectalinear grid
vtr_reader = vtk.vtkXMLRectilinearGridReader()
vtr_reader.SetFileName('output00000023_conc.vtr')
vtr_reader.Update()

# load file poly data
vtp_reader = vtk.vtkXMLPolyDataReader()
vtp_reader.SetFileName('output/output00000023_cell.vtp')
vtp_reader.Update()
```

This is how I load poly data file in fury.

```python
import fury

# load file
v_cell = fury.io.load_polydata('output/output00000023_cell.vtp')

# actor
actor = fury.get_actor_from_polydata(v_cell)

# scene
scene = fury.window.Scene()
scene.add(actor)

# show
showm = fury.window.ShowManager(scene)
showm.initialize()
showm.start()
```

As said, unfortunately, there is no way to load rectilinear grid data in fury :' (.
And it would be nice, if this format would be supported.

Best, Elmar

[output00000023_cell.vtp.txt](https://github.com/user-attachments/files/19166539/output00000023_cell.vtp.txt)
[output00000023_conc.vtr.txt](https://github.com/user-attachments/files/19166538/output00000023_conc.vtr.txt)

skoudoro:
Thank you for your feedback @elmbeech.

We will add it for the next release

elmbeech:
Thank you, @skoudoro and fury developers! Happy to test the implementation as soon as it is released.

elmbeech:
it does not look to me like this is supported in release v0.12.0 .
i will give a try with pre-release v2.0.0.a1 ...

hmm ...
with pre-release v2.0.0.a1 i am not even able to load the polydata.

maharshi-gor:
Hey @elmbeech 

Thank you for keeping an eye on this issue.

v.0.12.0 was required to make the pre-release, so there was no new active development.

In the pre-release 2.0.0.a1 we are still working and we have not started supported any formats.

[...]

PR #1105, opened 2026-02-21
NF: Add load_rectilinear_grid support for .vtr files #971

AdityaGupta716:
Refs #971

## Changes
- Added `load_rectilinear_grid` function in `fury/io.py`
- Uses vtk to read `.vtr` files and converts data to numpy arrays
- Returns x, y, z coordinates and point/cell data as numpy arrays
- Added tests in `fury/tests/test_io.py`

## Notes
This PR adds the **loading** part of the feature. Visualization 
(actor creation from rectilinear grid) can be added as a follow-up 
once v2 actor support matures further.

## Usage
```python
from fury.io import load_rectilinear_grid

data = load_rectilinear_grid("output.vtr")
print(data["x_coords"])        # [-15.  15.  45. ...]
print(data["y_coords"])        # [-10.  10.  30. ...]
print(data["z_coords"])        # [-5.   5.  15. ...]
print(data["point_data"])      # {'oxygen': array(...), 'water': array(...)}
print(data["cell_data"])       # {}
```

Tested with the sample `.vtr` file attached to the issue.


AdityaGupta716:
@maharshi-gor plz review

m-agour:
Hi @AdityaGupta716, thanks for doing this!, we have been discussing this and whether we should use vtk or not for loading these files. Also the end goal would be to visualize it as Fury actor but before doing the visualization part, we will discuss this in the next Fury meeting and then will let you know how to proceed with this PR.

maharshi-gor:
This needs to be done in a separate sub repo for IO from other software.

PR #1106, opened 2026-02-22
Remove the `_frame_count == 2` workaround previously used to trigger

### LINKED ISSUE #1090: Introduce UI Layouting to refactor workarounds
To recalculate the layout on the 2nd frame, we use this `_frame_count`  as of now as a workaround.

_Originally posted by @ganimtron-10 in https://github.com/fury-gl/fury/pull/1052#discussion_r2723512037_

maliojiw:
## Summary

Removes the `_frame_count == 2` layout workaround and replaces it with a per-element `_layout_dirty` flag in `TextBlock2D`.

## Problem

The previous frame-based logic recalculated layout only during initial render and did not correctly handle runtime changes such as updates to `message`, `font_size`, or `justification`.

## Solution

Layout-affecting setters now mark the element as dirty.  
During rendering, `render_screens` recursively traverses the UI tree and calls `update_layout()` only for elements where `_layout_dirty` is `True`, once `_aabb` becomes valid.

## Tests

Added a regression test verifying that updating `TextBlock2D.message` at runtime triggers a layout recalculation.

Closes #1090

### DIFF
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -965,6 +965,7 @@ class TextBlock2D(UI):
         self._bg_size = size
 
         self._last_rendered_size = (0, 0)
+        self._layout_dirty = True
 
         if self._bg_size is None and not self.dynamic_bbox:
             raise ValueError("TextBlock size is required as it is not dynamic.")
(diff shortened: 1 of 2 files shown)


ganimtron-10:
Hi @maliojiw ,

Thanks for the PR. I’m going to close this one because the layout change is an architectural shift that needs to happen in the core UI. It involves a lot of complex changes and design decisions that affect everything downstream, so it’s better if the core devs handle it instead.

I really appreciate the work you put in! Please feel free to check out any other "good first issues" if you want to keep contributing. Thanks!

PR #1125, opened 2026-02-25
Fix: remove redundant `local.position` offset in `actor_from_primitive` (multiple centers)

### LINKED ISSUE #1124: `actor_from_primitive` applies double position offset when using multiple centers
With `actor_from_primitive(..., centers=[...])`, the geometry is already translated to each center by `repeat_primitive`.  
If the actor is also shifted by `centers[0]`, everything gets translated twice (double offset).  
As a result, the repeated actor does not align with the expected positions (markers / individual actors).

### Here a screenshot of the issue

Image

### Reproduction code for the issue
```python


import numpy as np

import fury.primitive as fp
from fury.actor.core import actor_from_primitive

centers = np.array([[3, 0, 0], [6, 0, 0], [9, 0, 0]], dtype=np.float32)

vertices, faces = fp.prim_sphere(phi=16, theta=16)


multi = actor_from_primitive(
    vertices, faces, centers, colors=colors, scales=1.0, smooth=True
)



singles = []
for i in range(3):
    s = actor_from_primitive(
        vertices, faces, centers[i : i + 1],
        colors=colors[i : i + 1], scales=1.0, smooth=True,
    )
    s.local.y -= 4
    singles.append(s)


marker_centers = np.vstack([centers, centers - [0, 4, 0]])
markers = actor.sphere(
    centers=marker_centers,
    colors=(1, 1, 1),
    radii=0.15,
)


label_top = actor.text(
    "repeat_primitive (1 actor)",
    position=[-2, 0, 0],
    colors=(1, 1, 1),
    anchor="middle-right",
)
label_bot = actor.text(
    "individual actors",
    position=[-2, -4, 0],
    colors=(1, 1, 1),
    anchor="middle-right",
)


scene = window.Scene()
scene.background = (0.1, 0.1, 0.15)

scene.add(multi)
for s in singles:
    scene.add(s)
scene.add(markers)
scene.add(label_top)
scene.add(label_bot)

show_m = window.ShowManager(
    scene=scene, size=(900, 500), title="repeat_primitive offset test"
)
show_m.start()
```

faris-agour:
Fixes #1124

`actor_from_primitive` set `obj.local.position = centers[0]` after `repeat_primitive` had already translated each primitive copy to its center in the vertex buffer, causing a double offset.

### Changes:
- Removes the redundant `obj.local.position = centers[0]` assignment
- Removes related assertions in test helpers that validated the incorrect behavior

### Results:
Spheres now align correctly with their expected center positions.

fix

### DIFF
--- a/fury/actor/core.py
+++ b/fury/actor/core.py
@@ -437,7 +437,7 @@ def actor_from_primitive(
         wireframe_thickness=wireframe_thickness,
     )
     obj = create_mesh(geometry=geo, material=mat)
-    obj.local.position = centers[0]
+    
     obj.prim_count = prim_count
     return obj
 
--- a/fury/actor/core.py
+++ b/fury/actor/core.py
@@ -437,7 +437,8 @@ def actor_from_primitive(
         wireframe_thickness=wireframe_thickness,
     )
     obj = create_mesh(geometry=geo, material=mat)
-    
+    if not repeat_primitive:
+        obj.local.position = centers[0]
     obj.prim_count = prim_count
     return obj
 
--- a/fury/actor/tests/_helpers.py
+++ b/fury/actor/tests/_helpers.py
@@ -45,6 +45,8 @@ def validate_actors(actor_type="actor_name", prim_count=1, **kwargs):
     colors = kwargs.get("colors", None)
 
     if centers is not None:
+        npt.assert_array_equal(get_actor.local.position, centers[0])
+
         mean_vertex = np.round(np.mean(get_actor.geometry.positions.view, axis=0))
         npt.assert_array_almost_equal(mean_vertex, centers[0])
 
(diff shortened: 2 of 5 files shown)


maharshi-gor:
Hello @faris-agour 

Thanks for the work. I feel the issue is more deep than that if you completely remove the setting of centers and repeat_premitive is false, They will never be propagated to the correct location.

we might just need to update center if the repeat_premitive is false.

rest looks good to me.

also update tests for the same in test_core.py

faris-agour:
Hi @maharshi-gor ,

I’ve updated the logic so that ```obj.local.position``` is only set when ```repeat_primitive=False```, ensuring centers are propagated correctly without causing a double offset.

I also updated the tests in ```test_core.py``` to validate both cases:
When ```repeat_primitive=True```, geometry is pre-offset and ```local.position``` remains at the origin.
When ```repeat_primitive=False```, ```local.position``` matches the provided center.

maharshi-gor:
Hey @faris-agour Thanks for the help.

There are some incorrect removal changes in the test files. I am removing them but rest looks good. 

Merging.

PR #1127, opened 2026-02-26
ci: Add shell scripts for test execution and project installation.

AmoghaKA:
I have investigated the codebase, specifically focusing around the ci folder where you were looking, and found two important bugs in the CI shell scripts. I've implemented fixes for both:

1. ci/run_tests.sh: Because set -ex was active, whenever # pytest -svv $file failed, the script would immediately abort. This completely broke the error handling, which was designed to run through all tests and aggregate failures at the end . I updated the script to safely catch the error without aborting (pytest -svv $file || retVal=$?) and added a final exit code check to ensure the script fails the CI step accurately if any test fails.

2. ci/install.sh: The script was trying to call legacy scripts that don't exist in the project anymore (setup.py and setup_egg.py), which would cause failures for certain INSTALL_TYPE values. Since the project uses modern packaging with pyproject.toml (and Hatchling), I updated these sections to use standard pip install . and python -m build --sdist/--wheel commands instead.


maharshi-gor:
Hi @AmoghaKA thanks for looking into it.

We have currently moved to v2 branch. Which is going to become our main branch soon. That where the newer development is happening. Would you mind doing the proposed changes there.

Make sure to create a branch from v2 branch and create a pull request to v2.

PR #1132, opened 2026-02-27
V2 updates

AmoghaKA:
I have investigated the codebase, specifically focusing around the ci folder where you were looking, and found two important bugs in the CI shell scripts. I've implemented fixes for both:

1. ci/run_tests.sh: Because set -ex was active, whenever # pytest -svv $file failed, the script would immediately abort. This completely broke the error handling, which was designed to run through all tests and aggregate failures at the end. I updated the script to safely catch the error without aborting (`pytest -svv $file || retVal=$?`) and added a final exit code check to ensure the script fails the CI step accurately if any test fails.

2. ci/install.sh: The script was trying to call legacy scripts that don't exist in the project anymore (`setup.py` and `setup_egg.py`), which would cause failures for certain `INSTALL_TYPE` values. Since the project uses modern packaging with `pyproject.toml` (and Hatchling), I updated these sections to use standard `pip install .` and `python -m build --sdist/--wheel` commands instead.


ganimtron-10:
Hi @AmoghaKA,

Thanks for the PR. Could you please update the title and add a clear description? "V2 updates" is too vague, we need to understand the gist of the changes at a glance.

Also, the base branch was incorrect, so I’ve updated it for you. Please keep an eye on that for future PRs.

There are currently merge conflicts that need to be fixed and regarding the changes I’m still not clear on what this exactly do. Could you clarify that in the description in simple way with examples if possible?

maharshi-gor:
Hello @AmoghaKA thank for the help. I see lot of untracked txt files being added with this PR.

ganimtron-10:
Hi @AmoghaKA ,

Just a heads-up that this PR has been inactive for over a week now. If there are no updates or progress by next week, I’ll go ahead and close it to keep the PR queue clean.

Let me know if you’re still working on this or if you need any help!

Issue #1109, opened 2026-02-22
Port `TextBox2D` Component to v2


ganimtron-10:
**Is your feature request related to a problem?**
We are porting existing UI components to the v2 branch to remains consistent and functional as we transition to the new architecture.

**Describe the solution you'd like**
Port `TextBox2D` to the v2 branch. The implementation should follow the new architecture present in v2 branch while almost preserving the original API and user experience.

Key Requirements:
- Implementation: Port the core logic from the legacy branch into the v2.
- Dependencies: Identify and port any required sub-components or internal helper classes necessary for this UI to function correctly.
- Testing: Port and adapt all existing and create new tests associated with this component to verify features.

**Additional context**
- Base Branch: v2
- Reference the existing v2 UI components for the standards and architecture.

medha-14:
@ganimtron-10 👋

I’d be happy to take this up as well. Since this is currently blocking a few related issues, I’ll prioritize working on it and plan to submit a PR soon.

Please let me know if there’s anything specific you’d like me to keep in mind.

Thanks!

Issue #1128, opened 2026-02-26
Update `handle_events` to become more robust


ganimtron-10:
The `handle_events` should be refactored/rearchitected to enhance DX, maybe by allowing UI or actor or actors so that we dont need to manually specify all actors that needs to pickup specific events.

_Originally posted by @ganimtron-10 in https://github.com/fury-gl/fury/pull/1056#discussion_r2860801113_

vatsal-kakadiya:
Hi @ganimtron-10 👋

I’m Vatsal, a student developer interested in contributing to Fury, and I’ve recently started exploring the codebase. I’d like to work on this issue.

My idea is to refactor handle_events so it doesn’t require manually passing each actor. Instead of only accepting a single actor, it could take an optional actors argument and, if none is provided, automatically register callbacks for all actors returned by self.actors. This would improve DX, reduce repetitive calls, and keep backward compatibility.

If this approach looks good, I’d be happy to work on it. Please let me know if I can take this issue or if you’d suggest any changes before I start.

ganimtron-10:
Hi @vatsal-kakadiya ,
It would be great if you choose any other [good-first-issue](https://github.com/fury-gl/fury/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22) as this needs more discussion and ideation to be done before working on it.
Thanks!

JigyasuRajput:
Hey @ganimtron-10, here's an approach i came up with, lmk if that suits well:
                                                                                                       
  I'll refactor `handle_events` so it doesn't require manually passing each actor anymore. Right now
  every subclass has to call `self.handle_events(actor)` for each actor individually ...TextButton2D even
  calls it twice, once for the text and once for the background.

  My plan is to make the actor parameter optional. When no argument is passed, it'll automatically
  register event handlers on all actors returned by `self.actors`. I'll also add support for passing a UI
   element directly (so it picks up all its actors), or a list of actors if someone needs that.

  The existing calls that pass a single actor will still work as-is, so nothing breaks. Once this is
  in, we can simplify the subclass `_setup` methods to just call `self.handle_events()` with no arguments
  instead of repeating it for every actor.

PR #1134, opened 2026-02-28
Ports `Textbox2d` to v2 architecture

medha-14:
Issue #1109 

image

### DIFF
--- a/docs/examples/viz_textbox.py
+++ b/docs/examples/viz_textbox.py
@@ -13,7 +13,7 @@ textbox_single = TextBox2D(
     width=20,
     height=1,
     text="Single line",
-    position=(50, 400),
+    position=(50, 450),
     color=(1, 0, 0),
     font_size=20,
 )
@@ -27,11 +27,20 @@ textbox_multi = TextBox2D(
     font_size=25,
 )
 
+textbox_static = TextBox2D(
+    width=20,
+    height=2,
+    text="Static Text box",
+    position=(50, 200),
+    color=(0, 0, 0),
+    font_size=20,
+)
+
 textbox_styled = TextBox2D(
     width=20,
     height=2,
     text="Bold & Italic\nCenter justification",
-    position=(50, 150),
+    position=(50, 100),
     color=(0, 0.5, 0),
     font_size=18,
     justification="center",
@@ -41,6 +50,7 @@ textbox_styled = TextBox2D(
 
 scene.add(textbox_single)
 scene.add(textbox_multi)
+scene.add(textbox_static)
 scene.add(textbox_styled)
 
 if __name__ == "__main__":
--- a/docs/examples/viz_textbox.py
+++ b/docs/examples/viz_textbox.py
@@ -2,13 +2,22 @@
 ============
 UI Text Box
 ============
+
+This example demonstrates how to create and customize 2D Text Boxes
+using the FURY UI elements. We'll explore different configurations
+such as single-line, multi-line, and styled text.
 """
 
 from fury.ui import TextBox2D
 from fury.window import Scene, ShowManager
 
+##############################################################################
+# A Scene acts as a container for all our UI elements and 3D objects.
 scene = Scene()
 
+##############################################################################
+# 1. Simple Single-Line Text Box
+# We define a basic text box with a specified width, height, and color (red).
 textbox_single = TextBox2D(
     width=20,
     height=1,
@@ -18,6 +27,10 @@ textbox_single = TextBox2D(
     font_size=20,
 )
 
+##############################################################################
+# 2. Multi-Line Text Box
+# The text can span across multiple lines by using the newline character '\n'.
+# We also increase the height to accommodate the extra lines.
 textbox_multi = TextBox2D(
     width=20,
     height=3,
@@ -27,38 +40,51 @@ textbox_multi = TextBox2D(
     font_size=25,
 )
 
+##############################################################################
+# 3. Static Text Box
+# A standard text box that displays static text in black.
 textbox_static = TextBox2D(
     width=20,
     height=2,
     text="Static Text box",
     position=(50, 200),
-    color=(0, 0, 0),
+    color=(0, 0, 0),  # Black color
     font_size=20,
 )
 
+##############################################################################
+# 4. Styled Text Box
+# We can customize the appearance by enabling bold, italic properties
+# and setting the text justification to "center".
 textbox_styled = TextBox2D(
     width=20,
     height=2,
     text="Bold & Italic\nCenter justification",
     position=(50, 100),
-    color=(0, 0.5, 0),
+    color=(0, 0.5, 0),  # Dark green color
     font_size=18,
     justification="center",
     bold=True,
     italic=True,
 )
 
+##############################################################################
+# After configuring our UI elements, we must add them to the scene
+# so they will be rendered on the screen.
 scene.add(textbox_single)
 scene.add(textbox_multi)
 scene.add(textbox_static)
 scene.add(textbox_styled)
 
-if __name__ == "__main__":
-    current_size = (800, 700)
-    show_manager = ShowManager(
-        scene=scene,
-        size=current_size,
-        title="FURY 2.0: Text Box Example",
-    )
+##############################################################################
+# ShowManager handles the window creation, interaction, and rendering loop.
+current_size = (800, 700)
+show_manager = ShowManager(
+    scene=scene,
+    size=current_size,
+    title="FURY: Text Box Example",
+)
 
-    show_manager.start()
+##############################################################################
+# Finally, we start the interactive rendering loop to display our UI.
+show_manager.start()
--- a/docs/examples/viz_textbox.py
+++ b/docs/examples/viz_textbox.py
@@ -1,0 +1,54 @@
+"""
+============
+UI Text Box
+============
+"""
+
+from fury.ui import TextBox2D
+from fury.window import Scene, ShowManager
+
+scene = Scene()
+
+textbox_single = TextBox2D(
+    width=20,
+    height=1,
+    text="Single line",
+    position=(50, 400),
+    color=(1, 0, 0),
+    font_size=20,
+)
+
+textbox_multi = TextBox2D(
+    width=20,
+    height=3,
+    text="Multi-line\nText box",
+    position=(50, 300),
+    color=(0, 0, 1),
+    font_size=25,
+)
+
+textbox_styled = TextBox2D(
+    width=20,
+    height=2,
+    text="Bold & Italic\nCenter justification",
+    position=(50, 150),
+    color=(0, 0.5, 0),
+    font_size=18,
+    justification="center",
+    bold=True,
+    italic=True,
+)
+
+scene.add(textbox_single)
+scene.add(textbox_multi)
+scene.add(textbox_styled)
+
+if __name__ == "__main__":
+    current_size = (800, 700)
+    show_manager = ShowManager(
+        scene=scene,
+        size=current_size,
+        title="FURY 2.0: Text Box Example",
+    )
+
+    show_manager.start()
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -495,12 +495,19 @@ class UI(object, metaclass=abc.ABCMeta):
     def key_press_callback(self, event):
         """Handle key press event.
 
+        If there is an active UI element (e.g. a focused text box), the
+        event is forwarded to it. Otherwise, the event is handled by this
+        component's own ``on_key_press`` callback.
+
         Parameters
         ----------
         event : KeyboardEvent
             The PyGfx keyboard event object.
         """
-        self.on_key_press(event)
+        if UIContext.active_ui is not None:
+            UIContext.active_ui.on_key_press(event)
+        else:
+            self.on_key_press(event)
 
     def pointer_enter_callback(self, event):
         """Handle pointer enter event.
@@ -1007,9 +1014,17 @@ class TextBlock2D(UI):
         )
         self.background = Rectangle2D()
         self.handle_events(self.actor)
-        self.background.on_left_mouse_button_pressed = (
-            lambda event: self.on_left_mouse_button_pressed(event)
-        )
+        self.background.on_left_mouse_button_pressed = self._forward_background_press
+
+    def _forward_background_press(self, event):
+        """Forward background click to the text block's own press handler.
+
+        Parameters
+        ----------
+        event : PointerEvent
+            The PyGfx pointer event object.
+        """
+        self.on_left_mouse_button_pressed(event)
 
     def resize(self, size):
         """Resize the TextBlock2D bounding box.
@@ -1019,6 +1034,7 @@ class TextBlock2D(UI):
         size : (int, int)
             The new (width, height) in pixels.
         """
+        # Constrain text wrapping to the new width
         self.actor.max_width = size[0]
         self.update_bounding_box(size=size)
 
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -392,8 +392,9 @@ class UI(object, metaclass=abc.ABCMeta):
             self.middle_button_release_callback(event)
 
     def left_button_click_callback(self, event):
-        """
-        Handle left mouse button press event.
+        """Handle left mouse button press event.
+
+        Promotes the currently **Hot** UI element to **Active** state.
 
         Parameters
         ----------
@@ -403,13 +404,13 @@ class UI(object, metaclass=abc.ABCMeta):
         self.left_button_state = "pressing"
 
         if UIContext.hot_ui is not None:
+            # Deactivate a different previously-active element.
             if (
                 UIContext.active_ui is not None
                 and UIContext.active_ui is not UIContext.hot_ui
             ):
-                UIContext.active_ui.on_blur(event)
+                UIContext.active_ui = None
             UIContext.active_ui = UIContext.hot_ui
-            UIContext.active_ui.on_focus(event)
 
         self.on_left_mouse_button_pressed(event)
 
@@ -531,8 +532,7 @@ class UI(object, metaclass=abc.ABCMeta):
             self.on_key_press(event)
 
     def pointer_enter_callback(self, event):
-        """
-        Handle pointer enter event.
+        """Handle pointer enter event.
 
         Parameters
         ----------
@@ -540,11 +540,10 @@ class UI(object, metaclass=abc.ABCMeta):
             The PyGfx pointer event object.
         """
         UIContext.hot_ui = self
-        self.on_hover(event)
+        self.on_pointer_enter(event)
 
     def pointer_leave_callback(self, event):
-        """
-        Handle pointer leave event.
+        """Handle pointer leave event.
 
         Parameters
         ----------
@@ -553,7 +552,7 @@ class UI(object, metaclass=abc.ABCMeta):
         """
         if UIContext.hot_ui is self:
             UIContext.hot_ui = None
-        self.on_dishover(event)
+        self.on_pointer_leave(event)
 
 
 class Rectangle2D(UI):
@@ -677,10 +676,8 @@ class Rectangle2D(UI):
         size : (float, float)
             Rectangle size (width, height) in pixels.
         """
-        self._size = list(size)
-        w, h = size
-        w = w if w != 0 else 1
-        h = h if h != 0 else 1
+        w = max(size[0], 1)
+        h = max(size[1], 1)
         self.actor.geometry = plane_geometry(width=w, height=h)
         self._update_actors_position()
 
@@ -1016,6 +1013,8 @@ class TextBlock2D(UI):
         self._children.append(self.background)
         self.handle_events(self.actor)
         self.background.on_left_mouse_button_pressed = self._forward_background_press
+        self.background.on_pointer_enter = self._forward_background_pointer_enter
+        self.background.on_pointer_leave = self._forward_background_pointer_leave
 
     def _forward_background_press(self, event):
         """Forward background click to the text block's own press handler.
@@ -1026,6 +1025,29 @@ class TextBlock2D(UI):
             The PyGfx pointer event object.
         """
         self.on_left_mouse_button_pressed(event)
+
+    def _forward_background_pointer_enter(self, event):
+        """Forward background pointer-enter to set hot_ui to this TextBlock2D.
+
+        Parameters
+        ----------
+        event : PointerEvent
+            The PyGfx pointer event object.
+        """
+        UIContext.hot_ui = self
+        self.on_pointer_enter(event)
+
+    def _forward_background_pointer_leave(self, event):
+        """Forward background pointer-leave to clear hot_ui if still this element.
+
+        Parameters
+        ----------
+        event : PointerEvent
+            The PyGfx pointer event object.
+        """
+        if UIContext.hot_ui is self:
+            UIContext.hot_ui = None
+        self.on_pointer_leave(event)
 
     def resize(self, size):
         """
@@ -1082,6 +1104,10 @@ class TextBlock2D(UI):
         """
         Format the given text with markdown syntax for bold/italic styles.
 
+        pygfx markdown only supports ``**`` (bold) and ``*`` (italic) as
+        separate two-char / one-char star tokens.  It does **not** recognise
+        ``***`` for bold-italic, so we must wrap each word individually.
+
         Parameters
         ----------
         text : str
@@ -1092,14 +1118,40 @@ class TextBlock2D(UI):
         str
             The formatted markdown string.
         """
-        affix_char = ""
-        if self.bold:
-            affix_char = "**"
-        elif self.italic:
-            affix_char = "*"
+        if not self.bold and not self.italic:
+            return text
+
+        def _wrap_word(word):
+            """Wrap a single word with the appropriate markdown markers.
+
+            Parameters
+            ----------
+            word : str
+                The word to wrap with markdown markers.
+
+            Returns
+            -------
+            str
+                The word wrapped with bold/italic markdown markers.
+            """
+            if not word.strip():
+                return word
+            if self.bold and self.italic:
+                zws = "\u200b"
+                return f"*{zws}**{word}**{zws}*"
+            if self.bold:
+                return f"**{word}**"
+            return f"*{word}*"
 
         lines = text.replace("\r\n", "\n").replace("\r", "\n").split("\n")
-        formatted_lines = [f"{affix_char}{line}{affix_char}" for line in lines]
+        formatted_lines = []
+        for line in lines:
+            if not line.strip():
+                formatted_lines.append(line)
+                continue
+            words = line.split(" ")
+            formatted_words = [_wrap_word(w) if w else w for w in words]
+            formatted_lines.append(" ".join(formatted_words))
 
         return "\n".join(formatted_lines)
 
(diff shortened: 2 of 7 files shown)


medha-14:
I’ve made the required changes to the `Textbox2D` element and added more comprehensive tests, following the approach used for `LineSlider2D`. Could you please review the changes and let me know if anything else is needed?

medha-14:
@ganimtron-10 @maharshi-gor Whenever you get the chance, could you please take a look? I’d really appreciate your feedback so I can make the necessary improvements.

ganimtron-10:
Hi @medha-14 ,
Thanks for the PR!
Will review this PR sometime later today or tomorrow. Meanwhile can you add a demo/example so its quicker to test and verify.

medha-14:
@ganimtron-10 Thank you for the reminder. I have added the example now. Whenever you have the time, I would really appreciate it if you could take a look.

fury/ui/tests/test_elements.py
```diff
@@ -1,5 +1,7 @@
 """Test for components module."""
 
+from types import SimpleNamespace
```
ganimtron-10: Not required, we should ideally test logic through programmatic code and interactions through EventRecorder (which is still not in the v2, so can be skipped or handled other way around like it was done in RingSlider ig)

fury/ui/core.py
```diff
@@ -1007,6 +1007,9 @@ def _setup(self):
         )
         self.background = Rectangle2D()
         self.handle_events(self.actor)
+        self.background.on_left_mouse_button_pressed = (
+            lambda event: self.on_left_mouse_button_pressed(event)
+        )
```
ganimtron-10: Not a recommended way to do this, instead I would suggest creating a custom function and then assign that instead of lambda. Ideally it should be handled once this issue (https://github.com/fury-gl/fury/issues/1128) is resolved.

fury/ui/core.py
```diff
@@ -1016,7 +1019,7 @@ def resize(self, size):
         size : (int, int)
             The new (width, height) in pixels.
         """
-        self.actor.max_width = size[1]
+        self.actor.max_width = size[0]
```
ganimtron-10, line 1040: why assign height to max_width?

fury/ui/elements.py
```diff
@@ -24,14 +24,18 @@
 ]
 
 
+from string import printable
+
 import numpy as np
 
 from fury.actor import create_mesh
+from fury.decorators import warn_on_args_to_kwargs
```
ganimtron-10: not used anymore in v2

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
-
-#         self.width = width
-#         self.height = height
-#         self.window_left = 0
-#         self.window_right = 0
-#         self.caret_pos = 0
-#         self.init = True
-
-#         self.off_focus = lambda ui: None
+    @warn_on_args_to_kwargs()
```
ganimtron-10: remove

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
+        self._shadow = shadow
+        self._z_order = z_order
+
+        self.window_left = 0
+        self.window_right = 0
+        self.caret_pos = 0
+        self.init = True
+        self._has_focus = False
+
+        # user hooks
```
ganimtron-10, line 634: remove

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
-#         self.text.on_left_mouse_button_pressed = self.left_button_press
-#         self.text.on_key_press = self.key_press
+        Create the TextBlock2D component used for the textbox.
+        """
+        self.text = TextBlock2D(
+            text=self._message,
+            font_size=self._font_size,
+            font_family=self._font_family,
+            justification=self._justification,
+            dynamic_bbox=True,
```
ganimtron-10, line 655: Ideally this shouldnt be default behaviour

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
+        """Get the actors composing this UI component.
 
-#         """
-#         self.text.add_to_scene(scene)
+        Returns
+        -------
+        tuple
+            Width and height of the text bounding box.
+        """
+        return self.text.actors
```
ganimtron-10: Add self.text to self._children and let UI handle adding it

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
 
-#     def _set_position(self, coords):
-#         """Set the lower-left corner position of this UI component.
+    def _add_to_scene(self, scene):
+        """Add all subcomponents or VTK props that compose this UI component.
```
ganimtron-10: not required any more in v2

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
-#         self.window_left = 0
-#         self.caret_pos = self.window_right
+    def _update_actors_position(self):
+        """Update the position of the text actor (anchor-aware).
 
-#     def width_set_text(self, text):
-#         """Add newlines to text where necessary.
+        This implements the abstract layout hook required by UI.
+        """
+        if not hasattr(self, "text") or self.text is None:
```
ganimtron-10: why do we need this, should this be always present

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
+        ret_text = ret_text[self.window_left : self.window_right + 1]
+        return ret_text
 
-#         Parameters
-#         ----------
-#         i_ren: :class:`CustomInteractorStyle`
-#         obj: :class:`vtkActor`
-#             The picked actor
-#         _textbox_object: :class:`TextBox2D`
+    @warn_on_args_to_kwargs()
```
ganimtron-10: remove

fury/ui/elements.py
```diff
@@ -509,360 +513,442 @@ def handle_release_callback(self, event):
...
-#         if is_done:
-#             i_ren.remove_active_prop(self.text.actor)
+        Parameters
+        ----------
+        event : PointerEvent
+            The pointer event.
+        """
+        key = getattr(event, "key", None)
+        key_char = getattr(event, "key_char", "")
+        if not key_char and key and len(key) == 1:
```
ganimtron-10: do we still need to do this, even if we know its an KeyEvent?

fury/window.py
```diff
@@ -650,6 +650,9 @@ def __init__(
             EventType.POINTER_UP,
             EventType.POINTER_MOVE,
         )
+        self.renderer.add_event_handler(
+            self._global_key_handler, EventType.KEY_DOWN, EventType.KEY_UP
+        )
```
ganimtron-10, line 650: Why do we need this can't this be handled with the UI core on_key_press method?

fury/window.py
```diff
@@ -703,6 +706,17 @@ def _register_drag(self, event):
...
+        """Handle global key events by forwarding them to the active UI element.
+
+        Parameters
+        ----------
+        event : KeyboardEvent
+            The PyGfx keyboard event object.
+        """
+        if UIContext.active_ui:
+            UIContext.active_ui.on_key_press(event)
+
```
ganimtron-10, line 716: Can't we do this in UI Core instead of doing it here?
Also handle the hot and active state?

fury/window.py
```diff
@@ -850,9 +864,15 @@ def _set_key_long_press_event(self, event):
...
-            )
+            if self._key_long_press is not None:
+                return
+            try:
+                loop = asyncio.get_running_loop()
+                self._key_long_press = loop.create_task(
+                    self._handle_key_long_press(event)
+                )
+            except RuntimeError:
+                pass
```
ganimtron-10: I dont specifically thing this is a good idea.

ganimtron-10 (changes requested):
Hi @medha-14 ,

You are on right tracks but found few initial hiccups while playing with the demo.

1. Bold and Italic example is not exactly working as expected (might need to check TextBlock2D)

Image

2. When I try typing into the TextBox same key is registered multiple times

Image

I have also added few comments below, PTAL.

fury/ui/core.py
medha-14: Thanks for catching this I have removed the redundant lambda here now

maharshi-gor:
Hey @medha-14 are you still planning to work on this?

medha-14:
> Hey @medha-14 are you still planning to work on this?

Yes, I am just in the process of addressing the reviews , sorry for the delay.

fury/window.py
medha-14: I didn't realise initially that this could cause duplicate events, I have removed the entire `_set_key_long_press_event` / `_handle_key_long_press` async mechanism now.

fury/ui/elements.py
medha-14: Thanks for pointing this out! since we know it's a pygfx event, event.key is always available directly. I have removed it now.

fury/ui/core.py
medha-14: It’s actually using the width, not the height here. The size tuple is in the form (width, height), so size[0] represents the width.
max_width controls how the text wraps horizontally, so it should be set based on the width of the container.

medha-14:
Hi @ganimtron-10 and @maharshi-gor, thanks a lot for the detailed l review. I’ve addressed all the comments and also fixed the issues related to multiple key presses and italics.

Could you please take another look and let me know if anything else needs to be improved?

maharshi-gor:
Hey @medha-14 the CIs are failing please check those.

medha-14:
> Hey @medha-14 the CIs are failing please check those.

@maharshi-gor could you please run the checks again, I have fixed the issue now. Thanks :)

ganimtron-10:
Hi @medha-14, while skimming through the code, I didnt feel the implementation to be as expected.

Main thing that didnt felt right was the `active_ui` setup, currently when the left mouse is pressed we directly set the textbox to be `active_ui` which should not be the case because moving forward we might have a hierarchy of things and setting `active_ui` might be complex.

Ideally how it should work is UI flows between mainly three states (for understanding simplicity)
a. Normal b. Hot c. Active

Normal is the default state of the UI.
Hot is when the user might be trying to active the UI ie. when he hover over the UI area or entering into it.
Active is when the UI was hot and then the user started interacting with it ie. he hovered onto the UI area and then clicked it.
(Great Video for more context: https://youtu.be/Z1qyvQsjK5Y?si=kg9pYfwSdNQtQNO8)


Currently we are directly setting it into `active_ui` which works now but cause us problem down the line.
If you could fix this flow and add the `hot_ui` logic on entry and disable it on exit and then if the UI is interacted then setting it to active needs to be done. Also all this logic will not be surely not be into the individual textbox, it will be in a combination of core UI, windows.py and ui/context.py when and where required

medha-14:
@ganimtron-10 Thanks for sharing the resource , this is really helpful. I understand your point about the UI state flow and how directly setting `active_ui` on mouse press could cause issues as the system scales.

I’ll go through the implementation again, to add  the proper `hot_ui` handling on hover or entry and exit, and update the flow so that `active_ui` is only set after the correct transition. 

I’ll push a fix soon, thanks again for the detailed feedback!

maharshi-gor:
Hey @medha-14 any new updates?

medha-14:
@ganimtron-10 @maharshi-gor I have added the suggested  changes, please have a look and let me know if anything else is needed :)

fury/ui/core.py
```diff
@@ -391,12 +391,24 @@ def mouse_button_up_callback(self, event):
     def left_button_click_callback(self, event):
         """Handle left mouse button press event.
 
+        Promotes the currently **Hot** UI element to **Active** state.
```
ganimtron-10: Remove comment not required.

fury/ui/core.py
```diff
@@ -391,12 +391,24 @@ def mouse_button_up_callback(self, event):
...
+
         Parameters
         ----------
         event : PointerEvent
             The PyGfx pointer event object.
         """
         self.left_button_state = "pressing"
+
+        if UIContext.hot_ui is not None:
+            # Deactivate a different previously-active element.
```
ganimtron-10: same here

fury/ui/core.py
```diff
@@ -495,20 +507,29 @@ def mouse_move_callback(self, event):
     def key_press_callback(self, event):
         """Handle key press event.
 
+        If there is an active UI element (e.g. a focused text box), the
+        event is forwarded to it. Otherwise, the event is handled by this
+        component's own ``on_key_press`` callback.
+
```
ganimtron-10: same here

fury/ui/core.py
```diff
@@ -1016,7 +1078,8 @@ def resize(self, size):
         size : (int, int)
             The new (width, height) in pixels.
         """
-        self.actor.max_width = size[1]
+        # Constrain text wrapping to the new width
```
ganimtron-10: same here

fury/ui/core.py
```diff
@@ -1045,6 +1108,10 @@ def _get_actors(self):
     def get_formatted_text(self, text):
         """Format the given text with markdown syntax for bold/italic styles.
 
+        pygfx markdown only supports ``**`` (bold) and ``*`` (italic) as
+        separate two-char / one-char star tokens.  It does **not** recognise
+        ``***`` for bold-italic, so we must wrap each word individually.
+
```
ganimtron-10: same here

fury/ui/elements.py
```diff
@@ -509,360 +512,421 @@ def handle_release_callback(self, event):
...
-#         self.off_focus = lambda ui: None
+    Currently supports:
+    - Basic text editing.
+    - Cursor movements.
+    - Single and multi-line text boxes.
+    - Pre text formatting (text needs to be formatted beforehand).
 
-#     def _setup(self):
-#         """Setup this UI component.
+    Init this UI element.
```
ganimtron-10: remove this line

fury/ui/elements.py
```diff
@@ -509,360 +512,421 @@ def handle_release_callback(self, event):
...
+        self._shadow = shadow
+        self._z_order = z_order
+
+        self.window_left = 0
+        self.window_right = 0
+        self.caret_pos = 0
+        self.init = True
+        self._has_focus = False
+
+        # user hooks
```
ganimtron-10: remove

fury/ui/elements.py
```diff
@@ -509,360 +512,421 @@ def handle_release_callback(self, event):
...
-#         Parameters
-#         ----------
-#         scene : scene
+    def _setup(self):
+        """Setup this UI component.
 
-#         """
-#         self.text.add_to_scene(scene)
+        Create the TextBlock2D component used for the textbox.
+        Uses dynamic_bbox so the bounding box adapts as the user types.
```
ganimtron-10, line 944: remove

fury/ui/elements.py
```diff
@@ -509,360 +512,421 @@ def handle_release_callback(self, event):
...
+        """
+        return self.text.size
 
-#     def width_set_text(self, text):
-#         """Add newlines to text where necessary.
+    def _update_actors_position(self):
+        """Update the position of the text actor (anchor-aware).
 
-#         This is needed for multi-line text boxes.
+        This implements the abstract layout hook required by UI.
```
ganimtron-10: remove

fury/ui/core.py
```diff
@@ -391,12 +391,24 @@ def mouse_button_up_callback(self, event):
...
         """
         self.left_button_state = "pressing"
+
+        if UIContext.hot_ui is not None:
+            # Deactivate a different previously-active element.
+            if (
+                UIContext.active_ui is not None
+                and UIContext.active_ui is not UIContext.hot_ui
+            ):
+                UIContext.active_ui = None
```
ganimtron-10, line 415: This is not specifically required as you are overriding it anyway in the next line.

But it can be a good start to handle the `on_focus`, `on_blur` and `on_hover` methods. Lets not add too many things for now in this PR and keep this in mind to create these methods later on once this is completed.

`on_focus` triggers when the ui becomes active
`on_blur` triggers when the ui gets removed from the active state 
`on_hover` is when the ui gets hot and stops when the ui become active (we need to also think about dehover (dont have the exact action name for now))

fury/ui/core.py
```diff
@@ -1007,6 +1033,42 @@ def _setup(self):
...
+        """Forward background pointer-leave to clear hot_ui if still this element.
+
+        Parameters
+        ----------
+        event : PointerEvent
+            The PyGfx pointer event object.
+        """
+        if UIContext.hot_ui is self:
+            UIContext.hot_ui = None
+        self.on_pointer_leave(event)
```
ganimtron-10: I dont think we need these extra method, instead we can directly assign the `Textbox2D`'s   pointer callbacks to the background rect's callback.

ganimtron-10 (changes requested):
Hi @medha-14 ,
PTAL at these below comments.
Also afaik we had a static background textbox too which would have a constant sized background and the text that overflowed was clipped and it would be a necessary feature for most of our users in this v2 textbox too. Please try and test the v1 textbox to understand more and let us know if you have any questions.
Thanks!

maharshi-gor:
Hey @medha-14 any updates on this?

medha-14:
@ganimtron-10 @maharshi-gor Thanks for the review. Apologies for the delay, this took a bit longer than expected to work through. I’ve added the static text box now. Could you please take a look and let me know if it works as expected.

skoudoro:
@ganimtron-10 and @maharshi-gor , Status of the review ? Can we get an update ?

Let me know if I should look at this.

medha-14:
The CI is failing on two tests that are unrelated to my changes. Both fail with a FetcherError as the SHA256 checksum of `icomoon.tar.gz` downloaded from the URL doesn't match the hardcoded hash in `fetcher.py`. Should I fix this or is it part of some other PR?

docs/examples/viz_textbox.py
```diff
@@ -0,0 +1,64 @@
...
+from fury.ui import TextBox2D
+from fury.window import Scene, ShowManager
+
+scene = Scene()
+
+textbox_single = TextBox2D(
+    width=20,
+    height=1,
+    text="Single line",
+    position=(50, 400),
```
ganimtron-10: update y position to 450 to stay consistent with others

Image

ganimtron-10 (changes requested):
Hi @medha-14,

I noticed you haven't addressed my previous two review comments yet, please take a look at those.

Also, the following features/fixes are still missing from the implementation:

1. **Cursor movement:** The cursor is always stuck at the end of the text. We can't move it between characters to edit, delete, or insert text.
2. **Text clipping issue:** In single-line mode, once the text overflows and clips, it disappears and never becomes visible again. In multi-line mode the text appears only after all the visible text is removed where as it should be a scrolling effect. 
3. **Unnecessary padding:** Your last changes added some extra padding tot he background. We don't need this until we actually support borders, decorations, and other styling features.
4. **Long key press:** Holding down a key doesn't repeat the action. For example, if I want to clear multiple characters, I have to click the backspace key multiple times instead of just holding it down.
5. **Data loss on click:** Whenever I first click on the textbox, all existing data is immediately wiped out, leaving a blank UI element. This shouldn't happen.

Attaching video below for your reference

https://github.com/user-attachments/assets/92fd5ebf-101b-4a40-896b-604dc41242db

Please fix these issues and update the PR. Thanks!

maharshi-gor:
Hello @medha-14 This PR needs attention as well.

medha-14:
@ganimtron-10 @maharshi-gor Thank you for the detailed review. I have made the requested changes please have a look :)

I also added `an on_blur` and updated the click logic to fix the caret persistence issue. Although you mentioned not adding focus/blur logic in this PR, I went ahead and included it because having the caret visible in multiple text boxes at once made it very confusing to identify which one was currently active. I am happy to remove it if you feel it is out of scope for this PR.

ganimtron-10:
Hi @medha-14 , Thanks for this! 
I tried the demo and it looks great with almost all the features required, I went through the code a bit, but wanna spend some more time and understand and test few nitty gritties. Will add my comment if I have any.

ganimtron-10:
Hi @medha-14 ,

This is very close to getting merged, just had below few things and then after a round of testing this would be good to go.

- The textbox currently doesnt support Capital Letters (with CapsLock or Shift), any special character and "Shift + Enter" for a new line, can you add these too into it?
- There is some kind of inconsistency with the number of lines the textbox shows, if you initially see the first textbox has 3 lines but when you type into it, it reduces to 2 and stays to two; similarly with third textbox initially its 2 and max it goes to 4 when its completely filled and after that it remains at 3. Can you try to figure out the reason and fix it?
- I noticed you removed the long_press_key_event from renderer and move it to Textbox, its great to have that at renderer level so that it is globally applicable and not specifically in Textbox only, as other future UIs might need that feature too.

Also while you do these updates please make sure to also add relevant testcases for all these cases so that those are always tested.

maharshi-gor:
Hey @medha-14 Please rebase this PR

medha-14:
I have made the suggested changes, PTAL . Thanks!

fury/ui/core.py
```diff
@@ -117,10 +117,14 @@ def __init__(
         self.on_middle_mouse_double_clicked = lambda event: None
         self.on_middle_mouse_button_dragged = lambda event: None
         self.on_key_press = lambda event: None
+        self.on_key_release = lambda event: None
         self.on_hover = lambda event: None
         self.on_dishover = lambda event: None
+        self.on_pointer_enter = lambda event: None
+        self.on_pointer_leave = lambda event: None
```
ganimtron-10: This is redundant as the `on_hover` and `on_dishover` are emitting the same events can you use those instead?

fury/ui/core.py
```diff
@@ -998,13 +1037,50 @@ def __init__(
 
     def _setup(self):
         """Set up this UI component."""
-        self.actor = Text(
-            markdown=self._message, screen_space=True, anchor="middle-center"
-        )
+        self.actor = Text(markdown=self._message, screen_space=True, anchor="top-left")
```
ganimtron-10: why use a `top-left` default anchor?

fury/window.py
```diff
@@ -709,15 +708,15 @@ def __init__(
...
-        )
         self.renderer.add_event_handler(
             self._register_drag,
             EventType.POINTER_DOWN,
             EventType.POINTER_UP,
             EventType.POINTER_MOVE,
         )
+        self.renderer.add_event_handler(
+            self._global_key_handler, EventType.KEY_DOWN, EventType.KEY_UP
+        )
```
ganimtron-10, line 721: why create a new function when we already have one? why not reuse it?

fury/ui/elements.py
```diff
@@ -798,360 +801,618 @@ def left_button_dragged(self, event):
...
+            self.move_left()
+        elif k in ("arrowright", "right"):
+            self.move_right()
+        elif k in ("arrowup", "up"):
+            self.move_up()
+        elif k in ("arrowdown", "down"):
+            self.move_down()
+
+        self.render_text()
+        return False
```
ganimtron-10, line 1132: This whole part can be further optimized

For eg
```python
# Define these once at the top of your script to avoid recreation of dictionary everytime
LOWERS = r"`1234567890-=[]\;',./"
UPPERS = r'~!@#$%^&*()_+{}|:"<>?'
SHIFT_TRANS = str.maketrans(LOWERS, UPPERS)

# Inside your function:
if is_shift:
    key_char = key_char.translate(SHIFT_TRANS)
```

ganimtron-10 (changes requested):
Hi @medha-14 ,

PTAL at these below comments.

Thanks!

fury/ui/core.py
medha-14: Thanks for pointing it out, I was trying something out to see how I could calculate the box size and left that in by mistake.

docs/examples/viz_textbox.py
```diff
@@ -0,0 +1,64 @@
...
+    bold=True,
+    italic=True,
+)
+
+scene.add(textbox_single)
+scene.add(textbox_multi)
+scene.add(textbox_static)
+scene.add(textbox_styled)
+
+if __name__ == "__main__":
```
maharshi-gor: can remove this if __name__ == "__main__" check?

docs/examples/viz_textbox.py
```diff
@@ -0,0 +1,64 @@
+"""
```
maharshi-gor, line 1: Can you make the tutorial little bit verbose so that user will be able to understand?

maharshi-gor (commented):
Hey @medha-14 Thank you for the work.

Just some minor comments, and then it should be GTG.

fury/window.py
```diff
@@ -709,13 +709,12 @@ def __init__(
             EventType.RESIZE,
         )
         self.renderer.add_event_handler(
-            self._register_drag,
+            self._global_event_handler,
```
ganimtron-10: I dont think mixing drag and key events is a good idea, please keep it seperate.

fury/ui/tests/test_elements.py
```diff
@@ -1,5 +1,7 @@
 """Test for components module."""
 
+from unittest.mock import Mock
```
ganimtron-10: as mentioned earlier we avoid using Mock for tests, please remove Mock

fury/ui/core.py
```diff
@@ -1037,7 +1035,9 @@ def __init__(
 
     def _setup(self):
         """Set up this UI component."""
-        self.actor = Text(markdown=self._message, screen_space=True, anchor="top-left")
+        self.actor = Text(
+            markdown=self._message, screen_space=True, anchor="middle-center"
+        )
```
ganimtron-10, line 1042: you reverted this back but the background calculation are not revert back, can you PTAL as I can see excess padding around text as shown below

Image

ganimtron-10 (changes requested):
Hi @medha-14 , PTAL at these comments too. 

Also from last few changes I am noticing that the changes are messing up the existing working functionality, it might be your AI but you need to check verify these changes and be mindful while pushing. Multiple recent to and fro reviews could have been saved if you would have tested and viewed your changes according to the review comments.

cc: @skoudoro @maharshi-gor

maharshi-gor:
As mentioned @medha-14 Please reduce it to legible 5-6 commits.

medha-14:
> As mentioned @medha-14 Please reduce it to legible 5-6 commits.

Thanks for the reviews I have reduced the commits to 6 legible ones. PTAL

maharshi-gor:
Hello @medha-14 I think there is some small conflict after the merging of other PR can you rebase?

medha-14:
> Hello @medha-14 I think there is some small conflict after the merging of other PR can you rebase?

@maharshi-gor I have rebased this, PTAL.

maharshi-gor:
Hey @medha-14 we again got some conflict as I merged ListBox2D.

maharshi-gor:
Hey @medha-14 I think there is some issue in rebase one of the test case is failing.

medha-14:
> Hey @medha-14 I think there is some issue in rebase one of the test case is failing.

fixed now :)

maharshi-gor:
Hey @medha-14 Thank you for the work. I am just curious there is a lot of space in the right and no space on left. Is it on purpose?

Otherwise everything was working as expected.

medha-14:
> Hey @medha-14 Thank you for the work. I am just curious there is a lot of space in the right and no space on left. Is it on purpose?
> 
> Otherwise everything was working as expected.

Yes, this is on purpose because here we define a fixed width (in terms of the number of characters). This creates a static background box that is wide enough to fit that many characters. Because the text justification defaults to "left", the text aligns to the left edge of this box. The remaining empty space on the right is reserved for when you type more characters. If it didn't have this space, the text box would jitter and constantly resize its background

PR #1142, opened 2026-03-04
UI: Adding PlaybackPanel UI

ganimtron-10:
Adding PlaybackPanel UI also called as Timeline UI which helps in controlling animation

image

### DIFF
--- a/fury/ui/elements.py
+++ b/fury/ui/elements.py
@@ -18,7 +18,7 @@ __all__ = [
     #     "FileMenu2D",
     #     "DrawShape",
     #     "DrawPanel",
-    #     "PlaybackPanel",
+    "PlaybackPanel",
     #     "Card2D",
     #     "SpinBox",
 ]
@@ -27,11 +27,13 @@ __all__ = [
 import numpy as np
 
 from fury.actor import create_mesh
+from fury.data import read_viz_icons
 from fury.io import load_image_texture
 from fury.lib import (
     plane_geometry,
 )
 from fury.material import _create_mesh_material
+from fury.ui.containers import Panel2D
 from fury.ui.core import UI, Anchor, Button2D, Disk2D, Rectangle2D, TextBlock2D
 
 
@@ -46,9 +48,11 @@ class TexturedButton2D(Button2D):
         Absolute coordinates (x, y) for placement.
     size : (int, int)
         Width and height in pixels.
+    is_toggle : bool, optional
+        If True, the button behaves as a toggle switch.
     """
 
-    def __init__(self, states, position=(0, 0), size=(30, 30)):
+    def __init__(self, states, position=(0, 0), size=(30, 30), is_toggle=False):
         """Initialize the textured button instance.
 
         Parameters
@@ -59,9 +63,11 @@ class TexturedButton2D(Button2D):
             Absolute coordinates (x, y) for placement.
         size : (int, int)
             Width and height in pixels.
+        is_toggle : bool, optional
+                If True, the button behaves as a toggle switch.
         """
         self.texture_map = self._load_textures(states)
-        super().__init__(position=position, size=size)
+        super().__init__(position=position, size=size, is_toggle=is_toggle)
 
     def _load_textures(self, states):
         """Load image files into PyGfx textures.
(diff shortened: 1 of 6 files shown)


maharshi-gor:
Hi @ganimtron-10 you can rebase this PR for fixing CIs

maharshi-gor:
Hey @ganimtron-10 I checked out to the playback panel.
Thank you for the PR.
The thing I do not like is that cube behind should not move while the interaction with the panel.

let's make the final UI eat the event in such a way that it can go to renderer for drag check but not to other elements.

I know it is a challenging task in itself, but let's try that.

I will provide indepth review on this once we can overcome. Please let me know if you need me for any discussion of any sort.

fury/ui/elements.py
```diff
@@ -352,7 +370,7 @@ def _get_actors(self):
         list
             List of actors from the track, handle, and text elements.
         """
-        return self.track.actors + self.handle.actors + self.text.actors
+        return []
```
maharshi-gor, line 373: what is the point of having this if we are sending empty list. But if we must have to implement this please change the docstring.

fury/ui/elements.py
```diff
@@ -509,6 +527,359 @@ def handle_release_callback(self, event):
...
+    Parameters
+    ----------
+    loop : bool, optional
+        If True, the playback starts in looping mode.
+    position : (float, float), optional
+        Absolute coordinates (x, y) for placement.
+    width : int, optional
+        The total width of the playback panel in pixels.
+    z_order : int, optional
+        The stacking priority of the panel.
```
maharshi-gor, line 542: Please mention the range allowed for `z_order`

fury/ui/elements.py
```diff
@@ -509,6 +527,359 @@ def handle_release_callback(self, event):
...
+        Parameters
+        ----------
+        loop : bool, optional
+            If True, the playback starts in looping mode.
+        position : (float, float), optional
+            Absolute coordinates (x, y) for placement.
+        width : int, optional
+            The total width of the playback panel in pixels.
+        z_order : int, optional
+            The stacking priority of the panel.
```
maharshi-gor, line 557: same here.

fury/ui/elements.py
```diff
@@ -509,6 +527,359 @@ def handle_release_callback(self, event):
...
+
+        self.time_text.set_position(
+            (pos[0] + 250 + pbar_length, pos[1] + 27),
+            x_anchor=Anchor.LEFT,
+            y_anchor=Anchor.CENTER,
+        )
+
+        self._children.extend([self.panel, self._progress_bar, self.time_text])
+
+    def _get_actors(self):
```
maharshi-gor, line 675: Again please update the docstring if it is just a placeholder.

maharshi-gor (changes requested):

maharshi-gor (changes requested):
Hello @ganimtron-10 Thanks for the work.

Please find my review comments.

maharshi-gor:
Also in the screenshot below the distance of text from + and - button doesn't look equal can you check that?

Screenshot 2026-04-17 121308

fury/ui/elements.py
ganimtron-10: This is an abstract method which is mandatory to be implemented in the child. Currently as this UI uses other UI elements its better to add those as childrens, instead of manually exposing their actors here as it helps segregate the management of actors to that UI specifically also it helps in the layouting that we will be implementing later down the line.

fury/ui/elements.py
ganimtron-10: The range currently supported is -inf to inf which is then internally normalized, so I dont think we should mention it

ganimtron-10:
> The thing I do not like is that cube behind should not move while the interaction with the panel.
>
> let's make the final UI eat the event in such a way that it can go to renderer for drag check but not to other elements.
>
> I know it is a challenging task in itself, but let's try that.

I tried figuring this out but this issue is trickier than it seems. I initially thought of blocking these events at UI scene level but as the dragging events are configured at the renderer level, blocking them at the UI scene level doesn't stop them from propagating. Also the controllers are hooked directly to the renderer, so they catch the event regardless. I’m looking for a workaround to intercept this before it hits the controllers, just wanted to highlight if you have some perspective or solution to it.

maharshi-gor:
How about naming them differently something that only FURY understands and then passing it to renderer? @ganimtron-10 

I have some idea let me know if you have time to discuss.

maharshi-gor:
Hey @ganimtron-10 I am merging this PR as it looks good. Let's tackle the dragging issue in separate as it would require changes elsewhere.

PR #1144, opened 2026-03-05
Fix 16-bit image loading returning flat 1D array

### LINKED ISSUE #1140: [BUG] load_image() returns wrong shape for 16-bit grayscale PNG
## Description

Loading 16-bit grayscale PNGs with `load_image()` returns an array with a completely wrong shape and instead of a 2D array matching the image dimensions, it comes back as a flat 1D array with no error or warning. not sure if this is intentional

Passing `as_vtktype=True` crashes outright with:
```
IOError: only 2D (L, RGB, RGBA) or 3D image available
```

Traced it back to `io.py` ~line 122 where the result of `reshape().astype()` isn't being assigned back, so the reshaped array gets thrown away.

## Way to reproduce

```python
import numpy as np
from PIL import Image
import tempfile, os
from fury.io import load_image

arr = np.random.randint(0, 65535, size=(256, 512), dtype=np.uint16)
img = Image.fromarray(arr, mode="I;16")
tmp = os.path.join(tempfile.gettempdir(), "test_16bit.png")
img.save(tmp)

result = load_image(tmp, as_vtktype=False)
print(result.shape)  # expected (256, 512), got (131072,)
```

A PR with a fix can be up shortly maybe after someone can review this issue.

- **OS:** Windows 11
- **System info:** `[paste get_info() output here]`

muzman123:
## Summary
- Fixes #1140
- `image.reshape().astype()` on line 122 of `fury/io.py` returns a new array, but the result was never assigned back to `image`, so the reshaped array was silently discarded
- Added `image = ` so the properly shaped array is kept

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -109,7 +109,7 @@ def load_image(filename):
             raw = pil_image.tobytes("raw", pil_image.mode)
             dtype = ">u2" if pil_image.mode.endswith("B") else "<u2"
             image = np.frombuffer(raw, dtype=dtype)
-            image.reshape(pil_image.size[::-1]).astype("=u2")
+            image = image.reshape(pil_image.size[::-1]).astype("=u2")
         else:
             try:
                 image = pil_image.convert("RGBA")
--- a/fury/tests/test_io.py
+++ b/fury/tests/test_io.py
@@ -2,7 +2,7 @@ import os
 from os.path import join as pjoin
 from tempfile import TemporaryDirectory as InTemporaryDirectory
 
-# from PIL import Image
+from PIL import Image
 import numpy as np
 import numpy.testing as npt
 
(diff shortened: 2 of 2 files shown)


ganimtron-10:
Hi @muzman123 ,
Thanks for the PR!
Please add relevant tests to make sure changes are tested and validated.

muzman123:
Hey, just added a test for 16 bit image loading, added it in test_io.py as that is where all the other image loading tests exist

maharshi-gor:
Hey @muzman123 Thanks for the change.

I think the real question is underlying shader supports those loaded images. I would like to see a tutorial loading 16bit image. using image actor.

also for the failing CIs

I would recommend install
```
pip install -r requirements\style.txt
```
Then
```
pre-commit install
```
This will make sure you have the correct code-format on your changes before you push.

ganimtron-10:
Hi @muzman123  ,

It's been a while since the last update on this PR. Could you let us know if you're still working on this? If we don't see any progress by next week, I’ll go ahead and close it to keep the queue manageable.

If you need any help or clarification to move forward, just let us know. Thanks!

maharshi-gor:
There are no updates on this. Closing.

PR #1147, opened 2026-03-06
BF/MTN: Bounding box for gpu streamtube actor

maharshi-gor:
The positions were calculated in gpu thus making compute incorrect bounding box.
Showing incorrect position of the camera. You can test `viz_streamtube.py` in example with and without changes from this PR.
The hidden failure was from the new pygfx version. updated the codebase and test cases for the same.


skoudoro (approved):
All good, thanks

PR #1153, opened 2026-03-07
Fix/windows apigen utf8 encoding

### LINKED ISSUE #1139: BLD: Sphinx doc build crashes on Windows due to missing encoding in apigen.py
The FURY v2 documentation build crashes completely on Windows before generating a single HTML page. This is a hard crash, not a warning — the entire Sphinx process aborts during the `builder-inited` phase with a `UnicodeDecodeError`. Any contributor on Windows who tries to build the docs locally hits this wall immediately and gets zero output. The root cause is four `open()` calls in `docs/source/ext/apigen.py` that do not specify a file encoding, combined with the fact that at least two FURY source files contain non-ASCII UTF-8 bytes that Windows cannot interpret under its default encoding.

## Environment Where This Reproduces

- OS: Windows 11 (10.0.26200)
- Python: 3.14.2 (CPython)
- Sphinx: 8.2.3
- System locale encoding: cp1252 (Windows Western European default)
- Branch: v2

This does not reproduce on Linux or macOS because those systems default to UTF-8 as their locale encoding. It is a Windows-only crash.

## Full Traceback

The Sphinx error log written to `%TEMP%\sphinx-err-*.log` contains the following:

```
File "...\fury\docs\source\ext\apigen.py", line 234, in _parse_module_with_import
    with open(mod.__file__) as fi:
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 14569: character maps to <undefined>
```

Sphinx wraps this as:

```
sphinx.errors.ExtensionError: 'charmap' codec can't decode byte 0x8f in position 14569: character maps to <undefined>
```

The build aborts at this point. No RST files are generated, no HTML is produced, and the build log contains zero warnings because the process never reaches the warning-emission stage.

## Root Cause — Detailed Explanation

Python 3's `open()` function, when called without an `encoding=` argument, uses `locale.getpreferredencoding(False)` to determine the encoding. On Windows with a standard installation this returns `cp1252` (Windows-1252, Western European). This encoding only covers 256 code points and cannot represent many UTF-8 byte sequences.

`apigen.py` is a Sphinx extension that crawls the FURY package, imports each module, opens the corresponding `.py` source file to parse its docstrings, and then writes the generated API reference RST files. There are four `open()` calls in this file, none of which specify `encoding=`:

**Line 207** — reads a Python source file by filesystem path:

    f = open(filename)

**Line 233** — reads a Python source file via the module's `__file__` attribute:

    with open(mod.__file__) as fi:

**Line 479** — writes a generated RST output file:

    fileobj = open(outfile, "w")

**Line 539** — writes another generated RST index file:

    idx = open(path, "w")

When `apigen.py` reaches `fury/actor/planar.py` during its module crawl, it opens the file using `cp1252`. The file contains the byte `0x8F` at position 14569. In `cp1252`, byte `0x8F` is undefined — it is not mapped to any character. Python raises `UnicodeDecodeError` and the entire Sphinx build crashes.

`fury/actor/odf_slicer.py` is a second confirmed triggering file. There may be others.

## Why This Byte Exists

UTF-8 uses multi-byte sequences for non-ASCII characters. The byte `0x8F` is a valid UTF-8 continuation byte and appears as part of a multi-byte UTF-8 sequence encoding a non-ASCII character — likely a special symbol, a non-breaking space, or a character in a contributor name inside a docstring or comment. It is perfectly valid UTF-8 and perfectly normal in Python source files, which are required by PEP 3120 to be UTF-8.

## Proof That the Fix Works

Running the same file-open logic with `encoding='utf-8'` across the entire FURY package produces zero UnicodeDecodeError instances. Every file in `fury/` opens and reads cleanly when UTF-8 is explicitly specified.

## The Fix

Add `encoding='utf-8'` to all four `open()` calls in `apigen.py`:

- Line 207: `f = open(filename)` becomes `f = open(filename, encoding='utf-8')`
- Line 233: `with open(mod.__file__) as fi:` becomes `with open(mod.__file__, encoding='utf-8') as fi:`
- Line 479: `fileobj = open(outfile, "w")` becomes `fileobj = open(outfile, "w", encoding='utf-8')`
- Line 539: `idx = open(path, "w")` becomes `idx = open(path, "w", encoding='utf-8')`

This is the correct and complete fix because PEP 3120 mandates that Python source files are UTF-8, and Sphinx itself expects UTF-8 for all RST output it processes. There is no situation where a different encoding would be more correct here.

I will submit a PR targeting the v2 branch with this fix.

Kaushalt2004:
On Windows, the docs build can fail early with a UnicodeDecodeError because apigen.py opens Python source files using the system default encoding (often cp1252). Some FURY modules contain valid UTF‑8 non‑ASCII characters, so the build aborts before any HTML is generated.
This PR makes apigen explicitly use encoding=utf-8for:
- reading module .py files
- writing the generated .rst files
Also adds a small regression test to ensure text-mode open() calls in apigen always pass an explicit encoding.
Fixes #1139.

### DIFF
--- a/docs/source/ext/apigen.py
+++ b/docs/source/ext/apigen.py
@@ -204,7 +204,7 @@ class ApiDocWriter:
             # nothing that we could handle here.
             return ([], [])
 
-        f = open(filename)
+        f = open(filename, encoding="utf-8")
         functions, classes = self._parse_lines(f)
         f.close()
         return functions, classes
@@ -230,7 +230,7 @@ class ApiDocWriter:
         patterns = "(?:{0})".format("|".join(self.object_skip_patterns))
         pat = re.compile(patterns)
 
-        with open(mod.__file__) as fi:
+        with open(mod.__file__, encoding="utf-8") as fi:
             node = ast.parse(fi.read())
 
         functions = []
@@ -476,7 +476,7 @@ class ApiDocWriter:
 
             out_module = ulm + self.rst_extension
             outfile = os.path.join(outdir, out_module)
-            fileobj = open(outfile, "w")
+            fileobj = open(outfile, "w", encoding="utf-8")
 
             fileobj.writelines(document_head + document_body)
             fileobj.close()
@@ -536,7 +536,7 @@ class ApiDocWriter:
             relpath = (outdir + os.path.sep).replace(relative_to + os.path.sep, "")
         else:
             relpath = outdir
-        idx = open(path, "w")
+        idx = open(path, "w", encoding="utf-8")
         w = idx.write
         w(".. AUTO-GENERATED FILE -- DO NOT EDIT!\n\n")
 
--- a/fury/tests/test_apigen_encoding.py
+++ b/fury/tests/test_apigen_encoding.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import builtins
+import importlib.util
+import sys
+from pathlib import Path
+
+
+def _load_apigen_module():
+    repo_root = Path(__file__).resolve().parents[2]
+    apigen_path = repo_root / "docs" / "source" / "ext" / "apigen.py"
+    spec = importlib.util.spec_from_file_location("fury_docs_apigen", apigen_path)
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_apigen_passes_utf8_encoding_for_text_open(monkeypatch, tmp_path):
+    apigen = _load_apigen_module()
+
+    # Create a module containing a UTF-8 sequence that includes byte 0x8F.
+    # U+03CF encodes to UTF-8 as: 0xCF 0x8F.
+    module_name = "_fury_apigen_encoding_smoke"
+    module_file = tmp_path / f"{module_name}.py"
+    module_file.write_text(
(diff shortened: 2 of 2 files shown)


maharshi-gor:
Hey @Kaushalt2004 can you explain how this helps on what we are doing.

The document generation happens on linux so that this might be an overkill.

Please expand if I am not getting it correctly.

Kaushalt2004:
Thanks for taking a look!

You're right that documentation generation in CI typically runs on Linux, where UTF-8 is the default, so this issue may not appear there.

However, this change is mainly aimed at improving reliability for local development, especially on Windows. On Windows systems, the default file encoding is often cp1252 instead of UTF-8, which can lead to UnicodeDecodeError when reading source files that contain non-ASCII characters.

Since apigen reads Python source files and writes generated .rst files, explicitly specifying encoding="utf-8" ensures consistent behavior across platforms and avoids environment-dependent failures.

This makes the documentation generation more robust for contributors working on Windows, even if CI itself doesn’t expose the issue.

That said, if cross-platform support for apigen is not a priority or if Windows-based doc generation is out of scope, I’m happy to adjust or drop the change.

maharshi-gor:
Hey @Kaushalt2004 I really do not think this is required. 

Rather, I would recommend implementing something for changing the file to follow only LF line ending instead of allowing both CRLF and LF.

PR #1159, opened 2026-03-08
Fix: TextBlock2D text positioning due to stale pygfx bounding box

### LINKED ISSUE #838: Text positioning issue in `TextBlock2D`
## Description
The TextBlock2D has an issue with the text positioning. If we scale a TextBlock2D to be completely visible we would be able to clearly notice that the text touches the top of the background boundary. To fix this we tried adjusting the text actors position at #837 but that wasn't a correct way as we only need that to be done when `auto_font_scale` is set to True. We also tried reducing the Bounding box size of the textactor to 95 percent but still the issue persists that we need to offset the text a bit to make the text placed at center. This issue needs a bit of thinking and discussion as the TextBlock2D being a core part needs to be changed effectively so that it doesn't harm any other related component. So, for now we create this issue as a remainder to what all things we have discovered and where we stopped so we could continue from here on!

## Way to reproduce

- Simplest reference program
```python
from fury import ui, window

tb = ui.TextBlock2D(size=(300, 150), bg_color=(0.5, 0.5, 0.5), auto_font_scale=True, text="HI")
sm = window.ShowManager()
sm.scene.add(tb)
sm.start()
```

- Issue image
image

- System Information
```
{'fury_version': '0+unknown', 'pkg_path': 'D:\\FuryWorkspace\\fury\\fury', 'commit_hash': 'b29a518a', 'sys_version': '3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]', 'sys_executable': 'D:\\FuryWorkspace\\fury-env\\Scripts\\python.exe', 'sys_platform': 'win32', 'numpy_version': '1.22.4', 'scipy_version': '1.8.1', 'vtk_version': '9.1.0', 'matplotlib_version': '3.5.2', 'dipy_version': '1.5.0'}
```

JigyasuRajput:
pygfx Text._aabb is not recalculated when font_size or font_family is changed after construction. Since TextBlock2D creates the text actor with defaults and sets properties afterward, get_text_actor_size() returned stale values, causing text to be mispositioned for all non-center justifications.

Fix by re-setting the markdown text after font_size/font_family changes to force pygfx to recompute its layout metrics, and reorder __init__ so message (which calls set_markdown) runs last with final property values.

Closes #838
This pull request addresses a regression in the `TextBlock2D` UI component where changes to `font_size` or `font_family` after construction did not update the text layout and bounding box, leading to incorrect sizing and positioning. The fix ensures that the text layout is refreshed whenever these properties are changed, and adds a regression test to verify correct behavior.

**Bug Fixes and Improvements to Text Layout:**

* Added a private method `_refresh_text_layout` to `TextBlock2D` that forces recalculation of the text bounding box by resetting the markdown text. This method is now called whenever `font_size` or `font_family` is updated, ensuring the rendered text size and positioning are always accurate. [[1]](diffhunk://#diff-b2cc6095d447746c62a6bba581af84a9727e75acdaca37bfd31202a5a42d9b57R1377-R1384) [[2]](diffhunk://#diff-b2cc6095d447746c62a6bba581af84a9727e75acdaca37bfd31202a5a42d9b57R1115) [[3]](diffhunk://#diff-b2cc6095d447746c62a6bba581af84a9727e75acdaca37bfd31202a5a42d9b57R1140)

**Testing and Regression Coverage:**

* Added a new test `test_textblock2d_text_positioning` to verify that the text actor size correctly reflects changes in `font_size` and that text is properly positioned within its bounding box. This test guards against regressions related to text sizing and layout.

**Code Organization:**

* Minor reordering in the `__init__` method of `TextBlock2D` to maintain consistent initialization order for attributes.

### DIFF
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -992,11 +992,11 @@ class TextBlock2D(UI):
         self.have_bg = bool(bg_color)
         self.color = color
         self.background_color = bg_color
-        self.font_family = font_family
         self.bold = bold
         self.italic = italic
-        self.message = text
+        self.font_family = font_family
         self.font_size = font_size
+        self.message = text
 
         self.update_bounding_box()
 
@@ -1112,6 +1112,7 @@ class TextBlock2D(UI):
             Text font size.
         """
         self.actor.font_size = size
+        self._refresh_text_layout()
         if self.dynamic_bbox:
             self.update_bounding_box()
 
@@ -1136,6 +1137,7 @@ class TextBlock2D(UI):
             The font family.
         """
         self.actor.family = family
+        self._refresh_text_layout()
         if self.dynamic_bbox:
             self.update_bounding_box()
 
@@ -1371,6 +1373,14 @@ class TextBlock2D(UI):
     def _update_actors_position(self):
         """Update the position of the internal actors."""
         self.update_bounding_box()
+
+    def _refresh_text_layout(self):
+        """Force recalculation of the text bounding box.
+
+        Re-sets the markdown text so that pygfx recomputes its internal
+        layout metrics after property changes like font_size or font_family.
+        """
+        self.actor.set_markdown(self.get_formatted_text(self._message))
 
     def get_text_actor_size(self):
         """Get the rendered size of the text actor.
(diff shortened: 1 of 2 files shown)


ganimtron-10:
Hi @JigyasuRajput ,
The issue you are trying to fix is present on the v1 version, and as you might have seen we are moving to v2.
Also the PR explains something else and changes are completely different.
Closing this for now, Please create a new one if you get proper clarity on what specifically you are trying to fix.

PR #1164, opened 2026-03-08
Fix documentation build warnings

medha-14:
While I was trying to build docs locally via `make -C . clean && make -C . html-no-examples` in docs. I got dozens of warnings mostly in `.rst` files and few related to attributes not being private when they are supposed to. I have fixed them - if someone please run CI on this PR then it would be clearer for me to know whether the changes I made are correct. Note for this PR I did take help of llm as there was redundant work, but I have verified all the changes. Thanks!


docs/source/ext/apigen.py
```diff
@@ -254,9 +254,15 @@ def _parse_module_with_import(self, uri):
...
+                        target_id = n.targets[0].id
+                        if target_id.startswith("_") or pat.search(target_id):
+                            continue
+                        functions.append(target_id)
                     elif hasattr(n.value, "attr") and n.value.attr.startswith("vtk"):
-                        classes.append(n.targets[0].id)
+                        target_id = n.targets[0].id
+                        if target_id.startswith("_") or pat.search(target_id):
+                            continue
+                        classes.append(target_id)
```
maharshi-gor: We should not have any vtk related things. If present we need to remove it.

docs/source/ext/build_modref_templates.py
```diff
@@ -43,6 +43,8 @@ def generate_api_reference_rst(
         r".*FetcherError.*$",
         r".*urlopen.*",
         r".*add_callback.*",
+        r"^[A-Z][A-Z0-9_]+$",  # Skip ALL_CAPS module-level constants
+        r"^UIContext$",  # Skip re-exported class (documented in its own module)
```
maharshi-gor: I do not believe wildcarding all the constants will help as it will allow other things to be cone under the hood.

fury/colormap.py
```diff
@@ -987,7 +987,7 @@ def rgb2xyz(rgb):
     mask = rgb > 0.04045
     rgb[mask] = np.power((rgb[mask] + 0.055) / 1.055, 2.4)
     rgb[~mask] /= 12.92
-    return rgb @ xyz_from_rgb.T.astype(rgb.dtype)
+    return rgb @ _xyz_from_rgb.T.astype(rgb.dtype)
```
maharshi-gor, line 991: Why are we adapting to _ infront of the variable names?

maharshi-gor (changes requested):
Hello @medha-14 
Thanks for the efforts.

This needs more thought. Please check my comments.

docs/source/ext/build_modref_templates.py
medha-14: Ah right since Sphinx's `apigen.py` already imports the module we now just check `callable(getattr(mod, target_id))` before deciding it's a function.

fury/colormap.py
medha-14: I was making it private to ignores from autodoc, but I realized this is also fixed by skipping non-callable assignments (solution for above comment).

medha-14:
Can you please run CI on this? I want to see the built doc output and doc build errors and warnings thanks! 
Converting to draft as I am still working on this.

docs/source/ext/apigen.py
medha-14: Ah alright, removed it now. Thank you!

maharshi-gor:
Hi @medha-14 Thank you for working on this.

Can you make a log of what kind of warnings it is removing?

medha-14:
Hi checkout this file, it contains all the warnings when you run `make -C . clean && make -C . html-no-examples` [warninglogs.txt](https://github.com/user-attachments/files/26930411/warninglogs.txt) Thanks!

docs/source/conf.py
```diff
@@ -336,8 +335,7 @@
     "reference_url": {
         "fury": None,
     },
-    "filename_pattern": re.escape(os.sep),
```
skoudoro, line 339: Why are we removing this. Can you explain ?

fury/animation/animation.py
```diff
@@ -41,25 +41,6 @@ class Animation:
...
-    current_timestamp : float
-        The current time of the animation in seconds.
-    loop : bool
-        Whether the animation is in loop mode (True) or play once mode (False).
-    parent_animation : Animation
-        The hierarchical parent Animation for the current Animation.
-    static_actors : list
-        List of static actors that are not animated but added to the scene.
-    timeline : Timeline
-        The Timeline instance handling the current animation.
```
skoudoro, line 62: It does not make sense to remove this. It needs to be handle differently. 

See https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring

fury/animation/animation.py
```diff
@@ -267,9 +248,10 @@ def set_keyframe(
             Interpolator will be reinitialized if True.
         **kwargs : dict, optional
             Additional parameters for the keyframe. This can include:
-            - `in_cp` and `out_cp` for cubic Bézier interpolation (float).
-            - `in_tangent` and `out_tangent` for cubic spline interpolation,
-                ndarray, shape (1, M).
+
+            * `in_cp` and `out_cp` for cubic Bézier interpolation (float).
+            * `in_tangent` and `out_tangent` for cubic spline interpolation,
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/animation.py
```diff
@@ -413,13 +395,15 @@ def set_interpolator(self, attrib, interpolator, *, is_evaluator=False, **kwargs
...
+                def get_position(t):
+                    return np.array([np.sin(t), np.cos(t) * 5, 5])
 
-            def get_position(t):
-                return np.array([np.sin(t), np.cos(t) * 5, 5])
         **kwargs : dict, optional
             Additional parameters for the interpolator. This can include:
-            - spline_degree : int - The degree of the spline in case of setting a
+
+            * spline_degree : int - The degree of the spline in case of setting a
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/animation.py
```diff
@@ -489,7 +473,8 @@ def set_position_interpolator(self, interpolator, *, is_evaluator=False, **kwarg
             function that does not depend on keyframes.
         **kwargs : dict, optional
             Additional parameters for the interpolator. This can include:
-            - degree : int - The degree of the spline interpolation in case of setting
+
+            * degree : int - The degree of the spline interpolation in case of setting
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/animation.py
```diff
@@ -618,8 +603,9 @@ def set_position(self, timestamp, position, **kwargs):
             Position value.
         **kwargs : dict, optional
             Additional parameters for the keyframe. This can include:
-            - `in_cp` and `out_cp` for cubic Bézier interpolation (float).
-            - `in_tangent` and `out_tangent` for cubic spline interpolation,
+
+            * `in_cp` and `out_cp` for cubic Bézier interpolation (float).
+            * `in_tangent` and `out_tangent` for cubic spline interpolation,
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/animation.py
```diff
@@ -1238,15 +1224,6 @@ class CameraAnimation(Animation):
...
         path (position visualization). Higher values create smoother paths.
-
-    Attributes
-    ----------
-    camera : Camera
-        Camera being animated by the CameraAnimation.
-    duration : float
-        The duration of the animation in seconds.
-    loop : bool
-        Whether the animation is in loop mode (True) or play one mode (False).
```
skoudoro, line 1249: same as above, more thinking is needed

fury/animation/animation.py
```diff
@@ -1282,11 +1259,12 @@ def set_focal(self, timestamp, position, **kwargs):
...
-            - in_cp: ndarray - The in control point for cubic Bézier interpolation.
-            - out_cp: ndarray - The out control point for cubic Bézier interpolation.
-            - in_tangent: ndarray - The in tangent at that position for cubic spline
+
+            * in_cp: ndarray - The in control point for cubic Bézier interpolation.
+            * out_cp: ndarray - The out control point for cubic Bézier interpolation.
+            * in_tangent: ndarray - The in tangent at that position for cubic spline
               curve.
-            - out_tangent: ndarray - The out tangent at that position for cubic spline
+            * out_tangent: ndarray - The out tangent at that position for cubic spline
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/animation.py
```diff
@@ -1303,11 +1281,12 @@ def set_view_up(self, timestamp, direction, **kwargs):
...
```
skoudoro: Can you put back `-`. it is less confusing

fury/animation/timeline.py
```diff
@@ -38,29 +38,6 @@ class Timeline:
...
-    stopped : bool
-        Whether the timeline is currently stopped.
-    paused : bool
-        Whether the timeline is currently paused.
-    loop : bool
-        Whether the timeline is set to loop or play once.
-    animations : list
-        List of animations controlled by the timeline.
-    has_playback_panel : bool
-        Whether the timeline has a playback panel.
```
skoudoro, line 63: same as above

fury/ui/helpers.py
```diff
@@ -157,11 +157,12 @@ def cal_bounding_box_2d(vertices):
...
-        - bounding_box_min : ndarray
+
+        * bounding_box_min : ndarray
             Minimum coordinates of the bounding box [min_x, min_y].
-        - bounding_box_max : ndarray
+        * bounding_box_max : ndarray
             Maximum coordinates of the bounding box [max_x, max_y].
-        - bounding_box_size : ndarray
+        * bounding_box_size : ndarray
             Size of the bounding box [width, height].
```
skoudoro, line 166: Can you put back `-`. it is less confusing

docs/source/ext/apigen.py
```diff
@@ -248,19 +248,6 @@ def _parse_module_with_import(self, uri):
...
-                    if isinstance(n.value, ast.Call):
-                        if isinstance(n.targets[0], ast.Tuple):
-                            continue
-                        functions.append(n.targets[0].id)
-                    elif hasattr(n.value, "attr") and n.value.attr.startswith("vtk"):
-                        classes.append(n.targets[0].id)
-                except Exception:
-                    print(mod.__file__)
-                    print(n.lineno)
-                    print(n.targets[0])
```
skoudoro, line 263: we should keep it. We might need vtk to read some files

skoudoro (changes requested):
Hi @medha-14,

Thank you for this work.

Please, You have to put more thinking in your change and understand better some of the errors. 

Fixing the error by removing is not always a good solution. The goal is to improve and not just delete. For sure deleting will fix the problem.

docs/source/conf.py
medha-14: Oops I forgot to revert when I was experimenting sorry.

medha-14:
@skoudoro I did more research to fix `WARNING: duplicate object description`, and I think removing redundant doc strings attributes approach is the better than using :noindex: workarounds

This is what I tough and my intuition: the warning is coming because NumPy is not able to resolve the ambiguity between class attribute docstrings and the individual property docstrings for these attributes. That is why I removed them on purpose, as I felt they were redundant. Also in the built docs, duplicate attribute definitions can be seen.

I am aware that attributes and property methods are not programmatically the same, but I also feel that in this case they have the same meaning.

Could you please specify how you want me to tackle this?


Edit [ref](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring): 
image
 Attributes - `may be used to describe non-method attributes of the class` -- but we already have method attributes (with doc-string) for that class.

fury/animation/animation.py
medha-14: Please see - https://github.com/fury-gl/fury/pull/1164#issuecomment-4470312078 :)

skoudoro:
Hi @medha-14,

Your answer is very generic. Please, give me an example. give me a lot of details. 

In the `animation/timeline.py` files, where is the duplicate of `playback_panel` Attributes in the docstring  ? this is not a function. Please, explain. 

Are they all redundants? if not which one is redundant? 

I need more details @medha-14 ....

skoudoro:
> In a class, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters. 

This means, we do not need `Parameters` sections in the class since we document the `__init__` => to remove

> An Attributes section may be used to describe **non-method attributes of the class**:

**non-method attributes**  is important here. if you see `@property` for an attribute, it means it is a method (already documented) so it can not be documented  again at the class level.  So you need to check each parameter, it is case by case. 

I hope it helps @medha-14

medha-14:
Hi @skoudoro,

> In the animation/timeline.py files, where is the duplicate of playback_panel Attributes in the docstring ? this is not a function. Please, explain.

Yes totally agree that is why I had to bring it back in https://github.com/fury-gl/fury/pull/1164/commits/0ead011b8908a1b5d7a89fd7743d67de500ba520 along with attribute strings of `CameraAnimation` Class (which I removed because it was inheriting Animation class -- I realized this was wrong and I need to keep them).

For other Attributes strings (Class Animation). Few examples:

```
actors : list
        List of actors controlled by the Animation.
child_animations : list
    List of child Animations of this Animation.
duration : float
    The duration of the animation in seconds.
```

image
image
image

medha-14:
Thanks for the review! 


> > In a class, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters.
> 
> This means, we do not need `Parameters` sections in the class since we document the `__init__` => to remove
> 

Yes agree, I also feel they are duplicate and needs to be removed but this won't fix the duplicate object error as the error is due to duplicate attributes doc strings (not parameters). I will remove parameters docs  from class. 
> > An Attributes section may be used to describe **non-method attributes of the class**:
> 
> **non-method attributes** is important here. if you see `@property` for an attribute, it means it is a method (already documented) so it can not be documented again at the class level. So you need to check each parameter, it is case by case.
> 
Yes agree.

medha-14:
> > > In a class, the [Parameters](https://numpydoc.readthedocs.io/en/latest/format.html#params) section of the docstring details the constructor’s parameters.
> > 
> > 
> > This means, we do not need `Parameters` sections in the class since we document the `__init__` => to remove

@skoudoro Actually no the Parameter section from class is not optional, but the parameters of `__init__` are optional. https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring  
image

This can also been seen when I run pre-commit after removing it from class -> 
```
numpydoc-validation......................................................Failed
- hook id: numpydoc-validation
- exit code: 1

fury/animation/timeline.py:20: PR01 Parameters {'loop', 'length', 'animations', 'playback_panel'} not documented
```
So therefore I need to remove parameters doc strings from `__init__`. 


Edit: When I remove it from `__init__` I am getting same error so to avoid duplicates I exclude initialization methods from numpy validation see https://github.com/fury-gl/fury/pull/1164/commits/b70eaf44af826be4319c57cd64b067a9a5123e20. Another way would be to include not PR01 in `[tool.numpydoc_validation]` in toml.

skoudoro:
@maharshi-gor or @medha-14, can you fix the conflict here ?

Otherwise it is ready to be merged 👍 .

Thanks for this work @medha-14

medha-14:
Hi, I’ve squashed the commits and rewritten the commit history. Please let me know if you’d like me to change anything else. Thanks :)

skoudoro (approved):
Thank you @medha-14 , good work

merging

PR #1167, opened 2026-03-09
Relax VTK constraint for Python 3.14 compatibility

tejeshvenkat:
## Summary
Enables FURY to install on Python 3.14 (and newer) by relaxing the VTK version constraint.

## Problem
- `pyproject.toml` currently requires `vtk>=9.1.0, <9.4.0`
- PyPI only offers VTK 9.6.0 for Python 3.14
- Installation fails with: `No matching distribution found for vtk<9.4.0,>=9.1.0`

## Solution
- Remove upper bound: `vtk>=9.1.0`
- Add Python 3.12, 3.13, 3.14 to classifiers

## Testing
- `pip install -e .` succeeds on Python 3.14 with VTK 9.6.0
- `import fury` works
- No runtime issues observed in basic usage

Closes: N/A (improvement, not tied to an issue)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,13 +38,16 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
     "Topic :: Scientific/Engineering",
 ]
 
 dependencies = [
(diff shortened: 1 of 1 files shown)


maharshi-gor:
Hello @tejeshvenkat Thanks for the work. But, we have moved away from vtk. Please check v2 branch

tejeshvenkat:
Thanks for the feedback, Maharshi. I understand FURY has moved to v2 with
pygfx. I'll switch to the v2 branch and look for good-first-issues there.
Any specific issues you'd recommend for a GSoC applicant?

PR #1156, opened 2026-03-08
Fix: Panel2D crashes when used without borders (#1155)

JigyasuRajput:
Fixes #1155 

This pull request improves the robustness of the `Panel2D` UI container when borders are disabled (`has_border=False`). Several methods now safely handle the absence of borders, preventing potential errors and ensuring consistent behavior. Additionally, a new test verifies that `Panel2D` functions correctly without borders.

**Robustness improvements for borderless `Panel2D`:**

* Guard clauses were added to methods such as `_get_actors`, `_update_actors_position`, `update_border_coords`, `border_color`, and `border_width` to ensure they handle the case where `has_border` is `False`, returning empty lists or exiting early as appropriate. [[1]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR137) [[2]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdL188-R189) [[3]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR399-R400) [[4]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR416-R417) [[5]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR434-R436) [[6]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR450-R452) [[7]](diffhunk://#diff-53918b2a3199d9c46ad955f4e1c36bd0c776fd5228674415f4be2306080f23bdR475-R477)

**Testing improvements:**

* Added `test_panel2d_no_border_actors_and_scene` to verify that `Panel2D` can be used without borders, including adding actors, updating positions, adding to a scene, and accessing border properties without errors.Guard all self.borders references with self.has_border checks in _get_actors, _update_actors_position, update_border_coords, border_color, and border_width properties.

PR #1168, opened 2026-03-09
fix:Panel2D works with only size provided, no borders

### LINKED ISSUE #1155: Panel2D Does not work without borders
## Description
Panel2D does not work just the size is provided

suchiii29:
Hey! I was going through the issues and found that Panel2D crashes 
when you create it with just the size and no border arguments.

The problem was that self.borders was never initialized when 
has_border=False, so _get_actors() and _update_actors_position() 
were failing trying to access it.

I fixed it by:
- setting self.borders = {} when has_border is False
- adding has_border checks before accessing borders in the two methods

Tested it locally and it works fine now.

Fixes #1155


VanshAgarwal24036 (commented):
Nice catch on the issue and fix.

**One suggestion:** it might be helpful to add a small regression test to verify that `Panel2D` works properly when `has_border=False` (e.g., adding actors, updating positions, and adding the panel to a scene). 

Also, I noticed the PR removes the commented `TabPanel2D` class. Since this change is unrelated to the issue being fixed here, it might be better to restore that commented section in this PR.

suchiii29:
Thanks for the review! I've restored the commented TabPanel2D section. I'll add a regression test for the fix too, working on it now.

ganimtron-10:
Hi @suchiii29 ,
Thanks for the PR!
Closing it as its a duplicate of https://github.com/fury-gl/fury/pull/1156.
Please feel free to checkout good first issues section for things that you can pick up.

PR #1177, opened 2026-03-13
Remove _frame_count layout workaround

### LINKED ISSUE #1090: Introduce UI Layouting to refactor workarounds

VanshAgarwal24036:
### Summary

This PR addresses issue #1090 by removing the `_frame_count == 2` workaround that was previously used to trigger layout recalculation during the initial rendering frames.

Instead of relying on a hardcoded frame count check, the layout update logic now preserves the required initialization frames while allowing layout recalculation to be controlled through a dirty-flag style mechanism. This makes the behavior clearer and more maintainable while maintaining compatibility with the current rendering pipeline.

Additionally, the `_StreamtubeBakingShader` API was updated to align with the current shader interface by allowing an optional `_scene` argument and providing a compatible `get_bindings_info` override.

### Changes

* Removed the `_frame_count == 2` layout workaround in `window.py`
* Preserved required initialization frames to ensure correct renderer setup
* Updated `_StreamtubeBakingShader` method signatures for compatibility with the shader API
* All tests pass locally (`pytest`)

### Issue

Fixes #1090

### DIFF
--- a/fury/shader.py
+++ b/fury/shader.py
@@ -791,7 +791,7 @@ class _StreamtubeBakingShader(BaseShader):
 
         return {}
 
-    def get_bindings(self, wobject, _shared, _scene):
+    def get_bindings(self, wobject, _shared, _scene=None):
         """Describe storage buffers used by the compute shader.
 
         Parameters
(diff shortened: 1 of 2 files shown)


VanshAgarwal24036:
@ganimtron-10 @maharshi-gor Please review it.

ganimtron-10:
PTAL https://github.com/fury-gl/fury/pull/1106#issuecomment-3946566295

PR #1181, opened 2026-03-14
fix(docs): Implement exponential backoff for GitHub API to resolve HTTP 429/403 crashes

### LINKED ISSUE #924: Doc Generation Error - HTTP Error 429

williamcancodee:
## Description
The PR attempts to resolve the documentation generation crash caused by GitHub API rate limiting (HTTP Error 429/403), as indicated by issue #924. 

In the previous implementation, when an HTTP error occurred, `fetch_url()` would return an empty dictionary. Consequently, when `get_paged_request()` tried to access `f.status`, an `AttributeError` would occur. However, even if this did not occur, the 5-attempt retry loop would terminate instantaneously without a delay.

**Changes Made:**
* **Fixed Error Handling:** Modified `fetch_url()` to return the real `HTTPError` object (which naturally has a `.code/.status` attribute in recent versions of Python) or a dummy object for other types of exceptions to avoid `AttributeError` crashes.
* **Exponential Backoff:** Introduced a `time.sleep()` for `429` (Rate Limit) and `403` (Forbidden/Abuse) status codes.
* The script will now pause for progressively longer durations (1s, 2s, 4s, 8s, 16s) between attempts, giving the API quota window a chance to reset.
* Added `print` statements to inform the user that the script is deliberately pausing rather than freezing.

Fixes #924 

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code


docs/source/ext/github_tools.py
```diff
@@ -62,10 +63,15 @@ def fetch_url(url):
         f = urlopen(req)
     except Exception as e:
         print(e)
+        # In modern Python, HTTPError already has a built-in read-only .status property
+        if hasattr(e, 'code'):
+            return e
```
maharshi-gor: Please make sure you validate things. I think the HTTPError does not have .status property which will again go into same issue.

maharshi-gor (changes requested):
Thank you for working on this @williamcancodee 

It is a needed fix but we need to make sure that we do it correctly.

williamcancodee:
Hi. @maharshi-gor 

Thank you for looking into this and calling out those changes, I really appreciate your time and effort

You are right that urllib.error.HTTPError does not necessarily have a .status attribute, although it does have a .code attribute for the status code. To fix this, I made the following changes:

Added a function `_get_http_status()` with docstring to check for .status, then .code for HTTPError instances, and finally .getcode() for response instances.
Modified `get_paged_request()` to use this function’s result for backoff and logging (no longer direct access to f.status).
Made sure that the non-HTTP case still defines headers = {}, and made parsing links more lenient with `getattr(f, "headers", {})`.

Please let me know if this is in line with what you requested or if I should tweak anything.
I look forward to hearing from you

THANK YOU once again

docs/source/ext/github_tools.py
```diff
@@ -60,12 +75,19 @@ def fetch_url(url):
...
+            
+        # If it's a network drop or non-HTTP error, return a dummy object to prevent the dictionary crash
+        class MockError:
+            status = 500
+            headers = {}
```
maharshi-gor: Please remove the comments, these are not correct as well.

docs/source/ext/github_tools.py
```diff
@@ -37,7 +38,21 @@
...
 # Functions
 # ----------------------------------------------------------------------------
+def _get_http_status(response):
+    """Safely extract an HTTP status code from a urlopen()-like response.
 
+    Notes
+    -----
+    `urllib.error.HTTPError` exposes the HTTP status as `.code`,
+    while some response objects expose `.status`. Do not assume `.status` exists.
+    """
```
maharshi-gor, line 60: I would like better explanation of the response.

maharshi-gor: Also try to match the overall fury docstrings.

maharshi-gor (commented):
Thanks for the work @williamcancodee 

Please check my few comments on this. We can merge after those changes.

maharshi-gor (approved):
Thank you for the changes @williamcancodee 

LGTM

williamcancodee:
> Thank you for the changes @williamcancodee 
> 
> LGTM



Hi @maharshi-gor 

Thank you for taking the time to review and provide feedback and clarification.
I sincerely appreciate it 

It was a pleasure contributing and I look forward to working with you


Till next time,
William

PR #1193, opened 2026-03-26
BF: FURY import should not cause exception if no screen.

maharshi-gor:
- Updated the lines in the lib to not call display specific methods.
- Removed monkey patch to stabilize the test cases.


skoudoro (approved):
Much better, Thank you for cleaning up !

merging

PR #1195, opened 2026-03-27
Fix: update TextBlock2D background when font size changes (#1194)

### LINKED ISSUE #1194: Text background does not align as expected.
The text with background does not update on changing the size of the text.

oscarjfernandezim-design:
## Description
This PR fixes an issue where the background of a TextBlock2D element does not resize when font_size is changed while dynamic_bbox=True.
## Root Cause
The font_size setter updated the actor's font size but _aabb (pygfx's internal bounding box) is only recalculated after a render cycle. This meant that get_text_actor_size() always returned the old size, so the background never updated.
## Solution
Modified the @font_size.setter in fury/ui/core.py. Since _aabb cannot be forced to update without a render, the new background size is calculated proportionally using the ratio between the old and new font_size, applied to the current _aabb dimensions.
## Testing
Verified that background.size changes correctly after setting font_size when dynamic_bbox=True.
All 30 tests in fury/ui/tests/test_core.py pass.

Fixes #1194.

### DIFF
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -1362,13 +1362,11 @@ class TextBlock2D(UI):
             pos[0] + size[0],
             pos[1] + size[1],
         ]
-
         self.background.resize(size)
         self._bg_size = size
         self.background.set_position(pos)
 
         self.update_alignment()
-
 
     def _update_actors_position(self):
         """Update the position of the internal actors."""
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -1111,20 +1111,9 @@ class TextBlock2D(UI):
         size : int
             Text font size.
         """
-        old_font_size = self.actor._font_size
         self.actor.font_size = size
-
-        if self.dynamic_bbox and old_font_size > 0:
-            ratio = size / old_font_size
-            aabb = self.actor._aabb
-            current_w = aabb[1][0] - aabb[0][0]
-            current_h = aabb[1][1] - aabb[0][1]
-            new_w = max(1, int(current_w * ratio))
-            new_h = max(1, int(current_h * ratio))
-            self.background.resize((new_w, new_h))
-            self._bg_size = (new_w, new_h)
-        else:
-            self.update_layout()
+        if self.dynamic_bbox:
+            self.update_bounding_box()
 
     @property
     def font_family(self):
@@ -1373,11 +1362,13 @@ class TextBlock2D(UI):
             pos[0] + size[0],
             pos[1] + size[1],
         ]
+
         self.background.resize(size)
         self._bg_size = size
         self.background.set_position(pos)
 
         self.update_alignment()
+
 
     def _update_actors_position(self):
         """Update the position of the internal actors."""
@@ -1391,6 +1382,7 @@ class TextBlock2D(UI):
         (float, float)
             The (width, height) of the rendered text.
         """
+        self.actor._update_object()
         return (
             self.actor._aabb[1][0] - self.actor._aabb[0][0],
             self.actor._aabb[1][1] - self.actor._aabb[0][1],
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -1111,8 +1111,20 @@ class TextBlock2D(UI):
         size : int
             Text font size.
         """
+        old_font_size = self.actor._font_size
         self.actor.font_size = size
-        self.update_layout()
+
+        if self.dynamic_bbox and old_font_size > 0:
+            ratio = size / old_font_size
+            aabb = self.actor._aabb
+            current_w = aabb[1][0] - aabb[0][0]
+            current_h = aabb[1][1] - aabb[0][1]
+            new_w = max(1, int(current_w * ratio))
+            new_h = max(1, int(current_h * ratio))
+            self.background.resize((new_w, new_h))
+            self._bg_size = (new_w, new_h)
+        else:
+            self.update_layout()
 
     @property
     def font_family(self):
--- a/fury/ui/core.py
+++ b/fury/ui/core.py
@@ -1112,8 +1112,7 @@ class TextBlock2D(UI):
             Text font size.
         """
         self.actor.font_size = size
-        if self.dynamic_bbox:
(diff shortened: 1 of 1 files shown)


dnyan-codes:
#my first open source contribution

ganimtron-10:
Hi,

Thanks for the PR! I’m not sure about the current approach and think there might be a better way to handle this. I'll take a deeper look later this week to see what makes the most sense.

oscarjfernandezim-design:
Hi! Thanks for the feedback.

​That makes sense. My approach was focused on avoiding the stale _aabb issue immediately after font_size updates, so I used proportional scaling to keep the background consistent as a quick fix.
​Since this is my first PR to the project, I’m very eager to learn the 'FURY way' to handle this. Do you think it would be better to centralize this logic within the layout or the render pipeline instead?

​I’m happy to refactor the implementation based on your suggestions later this week. Thanks for taking the time to look into it!

maharshi-gor:
Hi @oscarjfernandezim-design I think I might have added more information about the issue.

The current issue is that it does not fit the complete text even without changing the font-size. It is not an issue of stale update on background size but rather. It is due to calculation of the font width and height on different machines.

Thank you @ganimtron-10 for initial review and let me know if my assessment is correct?

oscarjfernandezim-design:
@maharshi-gor Thank you for the feedback. I want to understand the root issue better before refactoring. 

Could you clarify:
1. Is the problem that `_aabb` calculations differ across machines due to font rendering differences?
2. Does this affect both the initial sizing AND font_size changes?
3. Should I focus on fixing `get_text_actor_size()` instead of the `font_size.setter`?

I appreciate your patience and guidance!

maharshi-gor:
Hey @oscarjfernandezim-design I will wait for @ganimtron-10 he is the author of this code and will have a better understanding before proceeding here.

ganimtron-10:
> Hi @oscarjfernandezim-design I think I might have added more information about the issue.
> 
> The current issue is that it does not fit the complete text even without changing the font-size. It is not an issue of stale update on background size but rather. It is due to calculation of the font width and height on different machines.
> 
> Thank you @ganimtron-10 for initial review and let me know if my assessment is correct?

Hi @maharshi-gor , do you have any example, screenshot or a video showing what issue you actually faced?
I assume you created the TextBlock with the default `dynamic_bbox` to `False` and you were expecting the background to cover the text (which will not happen and its expected, maybe to avoid confusion we can set the default `dynamic_bbox` to `True`). 
If its the above mentioned case then no specific change is required as of now.

maharshi-gor:
I set the dynamic_bbox to explicitly true, I might need to create an example as it was part of DIPY skyline so I do not have an example at hand.

Let me find something that can showcase what I am saying

oscarjfernandezim-design:
Hi @maharshi-gor, @ganimtron-10,

I’ve been looking into the failing test "test_textblock2d_font_size_updates_background_when_dynamic_bbox" and wanted to share a few observations while I continue investigating.

- When updating "font_size" (e.g., from 18 to 50), the "background.size" remains unchanged in my environment (stays around [8, 13]).
- I traced the flow through the "font_size" setter and "update_bounding_box", and added some debug prints inside "get_text_actor_size".
- It seems that even after updating the font size, "self.actor.compute_aabb()" returns very similar values (only a minimal change in width), which might indicate that the bounding box is not being fully recomputed after the text property update.

From this, I’m wondering if there might be a synchronization issue between the text actor properties and the bounding box calculation, or if an explicit update/refresh step is missing before querying the size.

I might be missing something here since I’m still getting familiar with this part of the codebase, so I’d really appreciate any guidance or pointers on whether this line of reasoning makes sense.

Happy to dig deeper or help build a minimal reproducible example if needed.

maharshi-gor:
I think you are on right direction @oscarjfernandezim-design But can you post some screenshots of the trails you did?

For reference for @ganimtron-10

oscarjfernandezim-design:
Thanks! Here are the screenshots and logs from my debugging.

As shown, even after updating the font_size to 50, the computed AABB dimensions remain unchanged, which keeps the background size constant.

This seems to confirm that the actor’s bounding box is not being recomputed after the font property update.

I’m currently investigating whether an explicit update/refresh step is needed before calling compute_aabb(), or if something is missing in the font_size setter.

Please let me know if you’d like me to explore a specific direction or test a potential fix.
test_2
test_1

oscarjfernandezim-design:
### Root cause & fix for TextBlock2D background not resizing on font_size change
After investigation, the root cause was confirmed: pygfx only updates the text actor's _aabb (axis-aligned bounding box) during an actual render cycle. This means that after setting self.actor.font_size = size, the _aabb still held the old dimensions, so get_text_actor_size() always returned the same values regardless of the new font size.
The fix was minimal and non-invasive: calling self.actor._update_object() inside get_text_actor_size() before reading _aabb. This forces pygfx to recompute the bounding box immediately, without waiting for a render.
pythondef get_text_actor_size(self):
image

This was confirmed by testing _update_object(), _layout_blocks(), and get_bounding_box() directly on a gfx.Text instance — only _update_object() correctly updated _aabb after a font_size change. All 30 tests now pass.

maharshi-gor:
Hey @oscarjfernandezim-design I will take deep look into the PR. 

Thanks for the investigation. 


A general tip:
Always make sure that your changes to fix this issue should not affect the existing structure.

maharshi-gor:
Also @ganimtron-10 I feel this is a probable cause.

oscarjfernandezim-design:
Hi @maharshi-gor, thanks for the feedback! I've ensured the fix specifically targets the _aabb update logic within TextBlock2D when the font size changes. All 30 existing tests passed, confirming that the core structure remains intact. Looking forward to your review!

ganimtron-10:
Hi @oscarjfernandezim-design ,

I am still trying to understand what exactly is happening under the hood, and what is causing it and when, so to pinpoint the exact solution. I understand you have a solution but I still think there is more to it than just surfacial change. I will take some time out and look into it deeply and get back.

maharshi-gor:
I am closing there is no new development. Please feel free to reopen.

PR #1215, opened 2026-04-17
NF: Vector Field Actor chunked based on device limits.

maharshi-gor:
## Description

This PR introduces the changes for chunking the vector field actor beyond the allowed size of workgroups by webgpu.

## Motivation

The current implementation of the vector field fails when you have more than 65535 * 64 i.e. ~4M voxels the `vector_field_slicer` and in turn `peaks_slicer` fails with a webgpu error.

## How to Test

You are free to use any data beyond 4M voxels or you can use the tutorial `viz_vector_field.py`. Change the grid size from 20, 20, 20 to 200, 200, 200 and it will still work as expected.

## Changes

Brings no changes to the user. Just that the user should you use `get_slice` and `show_slice` methods rather than directly changing the `cross_section` for better user experience.

### DIFF
--- a/docs/examples/viz_vector_field.py
+++ b/docs/examples/viz_vector_field.py
@@ -106,13 +106,13 @@ show_manager = window.ShowManager(
 
 def handle_key_event(event):
     print("Key pressed:", event.key)
-    position = vector_field_slicer_actor.cross_section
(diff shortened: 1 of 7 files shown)


skoudoro (approved):
All good, Thank you @maharshi-gor,

merging

PR #1216, opened 2026-04-19
Docs resolve duplicate object warnings in apigen and fix blog post paths

Sidd231:
This update tackles part two of my summer coding plan - clearing out build noise. A persistent flaw in apigen.py kept triggering repeated index alerts, something now resolved. Alongside that, outdated links scattered through older articles have been corrected. The clutter is gone, connections restored.

Changes Made

Now includes :no-index: within the class and function auto-generation process through apigen.py. The update quietly slips into place without drawing attention, adjusting how entries are handled behind the scenes. Code behavior shifts slightly - indexing avoided where needed. Each generated item follows this tweak without extra steps. Hidden but effective changes roll out with the next build.

When Sphinx processes the files, it skips reprocessing items already picked up by automodule near the beginning. That avoids counting the same elements twice. Near the file start, those entries get pulled in automatically. So later sections leave them out. Duplicate indexing gets blocked this way. The system stays clean because of how early collection works.
Thirty three build warnings vanished after updates to multiple parts of the system, such as fury.animation and fury.colormap. Though small, these changes cleaned up what had been lingering. Each fix tackled quietly without fanfare. Now the output runs clearer. Modules behave better during compilation. A few lines here, a tweak there - done.

On line three of the file, a broken link reference got corrected inside posts/2023/2023-07-17-week-7-tvcastillod.rst. Syntax errors like that one now match proper structure again. The update appears small yet changes how links display across the blog section. Behind the scenes, formatting rules apply more consistently after this tweak. Mistakes in markup tend to ripple unless caught early. This time around, clarity wins without extra noise.

Now links properly to the 2020 article by adjusting the file path. Direction fixed so it finds the right place.
Impact Total Warnings Reduced: 47 → 14.

Start strong with better health in place. The API reference now works more smoothly since every class and attribute has its own distinct index. This change avoids overlap, keeps things clear, a step forward in consistency.

Testing
After removing the build/ and source/reference/ folders, I ran .\make.bat html.
Build passed successfully, just 14 leftover warnings - none tied to these updates.

It checks out - the API items still show up when hunting through the HTML. Search keeps working just fine on those bits tucked inside the page code.

### DIFF
--- a/docs/source/ext/apigen.py
+++ b/docs/source/ext/apigen.py
@@ -352,7 +352,7 @@ class ApiDocWriter:
             body += f + "\n"
             body += self.rst_section_levels[3] * len(f) + "\n"
             # Smart check: If it's one of our constants, use autodata
-            if f.startswith('_FACE_') or f.isupper():
+            if f.startswith("_FACE_") or f.isupper():
                 body += "\n.. autodata:: " + f + "\n\n"
             else:
                 body += "\n.. autofunction:: " + f + "\n\n"
--- a/docs/source/ext/apigen.py
+++ b/docs/source/ext/apigen.py
@@ -351,7 +351,11 @@ class ApiDocWriter:
             # must NOT exclude from index to keep cross-refs working
             body += f + "\n"
             body += self.rst_section_levels[3] * len(f) + "\n"
-            body += "\n.. autofunction:: " + f + "\n\n"
+            # Smart check: If it's one of our constants, use autodata
+            if f.startswith('_FACE_') or f.isupper():
+                body += "\n.. autodata:: " + f + "\n\n"
+            else:
+                body += "\n.. autofunction:: " + f + "\n\n"
 
         return head, body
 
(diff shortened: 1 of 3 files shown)


Sidd231:
Hi @maharshi-gor, @ganimtron-10!

I’ve just pushed a significant update to this branch. I identified a systemic issue where apigen.py was causing duplicate object descriptions by double-indexing members already covered by the automodule directive.

By implementing a surgical fix in the generator logic, I've managed to reduce the build warnings from 47 down to 14.

Could you please review these changes when you have a moment? I’d love to get your feedback on this approach to cleaning up the documentation build.

Sidd231:
Hi @maharshi-gor @ganimtron-10! 

I just pushed out a large update to this branch. I found out that apigen.py was putting out duplicate object descriptions which it did by double indexing members that the automodule directive already covered. I got in and did a precise fix to the generator code which in turn reduced build warnings from 47 to 14. Once you have a chance, I’d love for you to review these changes. 

Also I’m interested in what you think of this approach to improve docs build.

maharshi-gor:
Thank you for the work but I do not believe that it is going to be helpful to put them in exclude list.

PR #1218, opened 2026-04-22
Animation fix docs v2

Sidd231:
Fix 7 documentation errors in fury/animation/animation.py that was caused due to wrong indentation


maharshi-gor:
Thanks for contributing but, I do not believe this will be helpful as the animation.py will be re written.

PR #1223, opened 2026-05-15
Feat: Adding Network Vizualization

ganimtron-10:
Adding Network Viz using Force Directed Layouting and currently supports .xnet, .gml and .gexf files

https://github.com/user-attachments/assets/2135f6e4-2c66-4e33-81f7-6a70b41dfa78

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -13,6 +13,7 @@ import numpy as np
 from fury.lib import Texture, wgpu
 
 # from fury.utils import set_input
+from fury.network.parser import parse_network, stringify_network
 
 
 def load_cube_map_texture(fnames, *, size=None, generate_mipmaps=True):
@@ -283,6 +284,61 @@ def get_extension(file_path):
     if ext.startswith("."):
         return ext[1:]
     return ext
+
+
+def load_network(file_path, format=None):
+    """
+    Load a network from a file.
+
+    Parameters
+    ----------
+    file_path : str
+        The path to the network file.
+    format : str, optional
+        The specific file format of the network file (e.g., 'gexf', 'gml', or 'xnet').
+
+    Returns
+    -------
+    tuple
+        A tuple containing:
+        - nodes_xyz (np.ndarray): Shape (N, 3) float32 array of node positions.
+        - edges_indices (np.ndarray): Shape (E, 2) int32 array of edge connections.
+        - colors (np.ndarray): Shape (N, 4) float32 array of node colors (RGBA).
+    """
+    if format is None:
+        _, ext = os.path.splitext(file_path)
+        format = ext.lstrip(".").lower()
+
+    with open(file_path, "r", encoding="utf-8") as f:
+        data = f.read()
+
+    return parse_network(data, format)
+
+
+def save_network(network_data, file_path, format=None):
+    """
+    Save a network to a file.
+
+    Parameters
+    ----------
+    network_data : tuple
+        A tuple containing:
+        - nodes_xyz (np.ndarray): Shape (N, 3) float32 array of node positions.
+        - edges_indices (np.ndarray): Shape (E, 2) int32 array of edge connections.
+        - colors (np.ndarray): Shape (N, 4) float32 array of node colors (RGBA).
+    file_path : str
+        The destination path where the network file will be written.
+    format : str, optional
+        The specific export format of the network file.
+    """
+    if format is None:
+        _, ext = os.path.splitext(file_path)
+        format = ext.lstrip(".").lower()
+
+    data = stringify_network(network_data, format)
+
+    with open(file_path, "w", encoding="utf-8") as f:
+        f.write(data)
 
 
 # def load_polydata(file_name):
--- a/fury/network/core.py
+++ b/fury/network/core.py
@@ -1,0 +1,179 @@
+"""Network actors."""
+
+import numpy as np
+
+from fury.actor import Actor, Line
+from fury.geometry import Geometry, buffer_to_geometry
+from fury.lib import (
+    Buffer,
+    PointsShader,
+    WorldObject,
+    register_wgpu_render_function,
+)
+from fury.material import LineMaterial, NetworkMaterial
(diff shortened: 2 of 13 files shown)


fury/io.py
```diff
@@ -259,6 +260,53 @@ def save_image(
...
 
+def load_network(file_path, format=None):
+    """Load a network from a file.
+
+    Parameters
+    ----------
+    file_path : str
+        The path to the network file.
+    format : str, optional
+        The specific file format of the network file (e.g., 'json', 'gml').
```
maharshi-gor: Can you provide the list of supported formats? As I see many more parser than json and gml

fury/network/core.py
```diff
@@ -0,0 +1,500 @@
...
+        edge_opacity : float, optional
+            Opacity of the edges (0.0 to 1.0).
+        """
+        super().__init__()
+
+        if not isinstance(nodes, np.ndarray) or nodes.ndim != 2 or nodes.shape[1] != 3:
+            raise ValueError("nodes must be a (N, 3) numpy array")
+
+        if not isinstance(edges, np.ndarray) or edges.ndim != 2 or edges.shape[1] != 2:
+            raise ValueError("edges must be a (M, 2) numpy array")
```
maharshi-gor, line 85: Please also check if edges should not be negative

fury/network/parser.py
```diff
@@ -0,0 +1,634 @@
...
+def parse_network(data, format):
+    """Parse string data into a Network object.
+
+    Parameters
+    ----------
+    data : str
+        The raw data containing network details.
+    format : str
+        The file layout format ('gexf', 'gml', or 'xnet').
+
```
maharshi-gor, line 609: I do not see JSON but you mention json in the io.

maharshi-gor (changes requested):
Hey @ganimtron-10 Thank you for the work.

I see some required changes in the io and parser mostly.

I do believe you can make the visualization cool with addition of some colors in the tutorial.

ganimtron-10:
Hi @maharshi-gor, I have updated the demo to use inferno colormap depending on node degree instead of random colors, and also fixed all of your comments. PTAL.

maharshi-gor:
Hey @ganimtron-10 some code format changes required can you check those.

ganimtron-10:
@maharshi-gor  @medha-14 PTAL

fury/io.py
```diff
@@ -263,6 +264,55 @@ def save_image(
...
+        _, ext = os.path.splitext(file_path)
+        format = ext.lstrip(".").lower()
+
+    with open(file_path, "r", encoding="utf-8") as f:
+        data = f.read()
+
+    return parse_network(data, format)
+
+
+def save_network(network, file_path, format=None):
```
medha-14: `stringify_network` passes the Network object directly to the parser's `stringify()` method but looking at `stringify()` in all 3 parsrs, it expects a tuple and raises a ValueError otherwise. Wouldn't this mean `save_network` would always crash when called with a Network object? Should there be a conversion step here ?

fury/network/core.py
```diff
@@ -0,0 +1,525 @@
...
+
+        if np.any(edges < 0):
```
medha-14, line 87: Should we also add an upper-bounds check here to make sure edges don't reference a node index that's out of range?  like edges = [[0, 999]] with only 3 nodes would slip through and crash later when creating a adjacency list.

fury/network/core.py
```diff
@@ -0,0 +1,525 @@
...
+        -------
+        dict
+            A dictionary containing workgroup dimension sizes.
+        """
+        n = int(ceil(wobject.n_nodes / 64))
+        return {
+            "indices": (n, 1, 1),
+        }
+
+    def get_bindings(self, wobject, _shared, _scene):
```
medha-14: we should   ` def get_bindings(self, wobject, _shared, _scene=None):` as it is giving a type error at runtime.

medha-14 (changes requested):

fury/network/core.py
medha-14: I was running this on a old version of pygfx , it seems to be working now, so please ignore this.

maharshi-gor:
I agree with @medha-14 something got overwritten in last change, as I remeber the stringify_network was before passing it as tuple but now it just passes the entire network.

Please look into this @ganimtron-10 so we can move ahead quickly.

ganimtron-10:
Thanks @medha-14 for catching it.
@maharshi-gor PTAL.

Thanks!

fury/network/core.py
```diff
@@ -0,0 +1,532 @@
...
+    BaseShader,
+    Binding,
+    Buffer,
+    PointsShader,
+    register_wgpu_render_function,
+)
+from fury.material import LineMaterial, PointsMaterial
+
+
+class Network(Mesh):
```
maharshi-gor: I did not catch this before? Why Mesh?

fury/network/core.py
```diff
@@ -0,0 +1,532 @@
...
+            if (current_speed > max_speed) {
+                vel = normalize(vel) * max_speed;
+            }
+
+            let new_pos = pos_i + vel * u_material.speed;
+
+            set_pos(i, new_pos);
+            s_velocities[i] = vec4<f32>(vel, 0.0);
+        }
+        """
```
maharshi-gor: This needs to in shader.py

And somethings can be reused. Also pass them in wgsl files the shader.

maharshi-gor (changes requested):
Thanks for the work @ganimtron-10 

Please check the shader and choosing of correct parent class for the Network object.

maharshi-gor:
Hey @ganimtron-10 please rebase the branch

fury/network/core.py
ganimtron-10: I copied SphGlyph missed removing it, update it to `WorldObject` and `Actor`

ganimtron-10:
Hi @maharshi-gor , Rebased and Updated PTAL

PR #1227, opened 2026-05-18
Feat: Adding Solar System Animation Example

ganimtron-10:
https://github.com/user-attachments/assets/1ab31040-5d64-4003-8a35-736f406c1d21

https://github.com/user-attachments/assets/35099863-5964-416f-9c1b-16d53bc5a42d

### DIFF
--- a/docs/examples/viz_helical_solar_system.py
+++ b/docs/examples/viz_helical_solar_system.py
@@ -31,72 +31,72 @@ scene = window.Scene(background=(0.01, 0.01, 0.02))
 planets_data = [
     {
         "filename": "8k_mercury.jpg",
-        "position": 7,
+        "position": 8,
         "earth_days": 58.0,
-        "scale": (0.4, 0.4, 0.4),
+        "scale": (0.3, 0.3, 0.3),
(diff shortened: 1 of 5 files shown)


maharshi-gor:
Hey @ganimtron-10 I tried doing this. But it has the seem on the sphere texture. We need to understand how can we solve that part with the Pygfx APIs.

ganimtron-10:
Hi @maharshi-gor , Can you PTAL I updated the `prim_sphere` to duplicate initial points(u=0 and u=1) at the last so the UV coords can seamlessly wrap, the scipy's ConvexHull was strict and dedupliated these points which is why we were getting that inconsistency. Please verify and let me know.

maharshi-gor (changes requested):
Hey @ganimtron-10 amazing work!

There is a factual problem in the setup.

As Earth is somehow smaller than every other planet. Also the position in can you make it little bit relative.

Otherwise it is good to merge!

ganimtron-10:
Hi @maharshi-gor, I have updated planets and their orbits to be factually correct, PTAL. Thanks!

maharshi-gor:
The failures are unrelated.

PR #1233, opened 2026-05-24
Integrate Version Switcher, Add Local Docs Server and Fix Version Redirection

medha-14:
image

image

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -33,6 +33,11 @@ html-no-examples:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+serve:
+	@echo "Serving docs at http://localhost:8000 ..."
+	@python -m webbrowser http://localhost:8000 &
+	@python -m http.server 8000 --directory "$(BUILDDIR)/html"
+
 upload:
 	python upload_to_gh-pages.py
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -45,3 +45,13 @@ To generate the documentation without running the examples:
 $ make clean
 $ make html-no-examples
 ```
+
+### Serve the Documentation Locally
+
+To view the generated documentation and test features like the version switcher (which require a local server to bypass CORS policies), run:
+
+```bash
+$ make serve
+```
+
+This will start an HTTP server at `http://localhost:8000` and open it in your default web browser. Press `Ctrl+C` to stop the server.
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -52,6 +52,14 @@ if "%1" == "html-no-examples" (
 	exit /B
 )
 
+if "%1" == "serve" (
+	:serve
+	echo Serving docs at http://localhost:8000 ...
+	python -m webbrowser http://localhost:8000
+	python -m http.server 8000 --directory "%BUILDDIR%\html"
+	exit /B
+)
+
 
 %SPHINXBUILD% -M %1 %SOURCEDIR% %BUILDDIR% %SPHINXOPTS%
 goto end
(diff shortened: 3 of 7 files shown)


medha-14:
@maharshi-gor @skoudoro When I am on a page other than `index.html` in the newly built local docs and select a particular version it redirects to that version's `index.html` instead of the page the user was previously on. Is this behavior intended or do you want me to change it? If that page or docs are not available in the previous version it will redirect to the index page.

maharshi-gor:
Please rebase this PR with master. As we do not see new releases beyond v0.10.x. 

Please verify after that rebase.

medha-14:
Done :)

ganimtron-10:
Hi @medha-14 ,
I tried building and serving docs and I can see these errors/warnings in the logs, can you PTAL if those are related to this PR else handle it in different PR?

image
image

ganimtron-10:
> @maharshi-gor @skoudoro When I am on a page other than `index.html` in the newly built local docs and select a particular version it redirects to that version's `index.html` instead of the page the user was previously on. Is this behavior intended or do you want me to change it? If that page or docs are not available in the previous version it will redirect to the index page.

Ideally if user is on a page and he selects different version then he should be redirected to the same page in that version and if its not present in that specific version then we can redirect him to index.html

ganimtron-10:
@maharshi-gor did you also wanted alpha releases to be there in the version selector? I don't recall what you specifically mentioned regarding that.

I can see these versions as of now
image

medha-14:
@ganimtron-10 thanks for the reviews!

> Hi @medha-14 ,
> I tried building and serving docs and I can see these errors/warnings in the logs, can you PTAL if those are related to this PR else handle it in different PR?

Hey I think these are introduced [in this commit](https://github.com/medha-14/fury/commit/8f823af20592f40e283f2835fec17318b37751c6) -- therefore I will address them in a separate PR.  

>  Ideally if user is on a page and he selects different version then he should be redirected to the same page in that version and if its not present in that specific version then we can redirect him to index.html


I fixed the version switcher bug just by removing `index.html` from the URLs in `versions_switcher.json`.

The issue was that `pydata-sphinx-theme` automatically adds the current page's name to the end of those base URLs. Because we had `index.html` in there, it was creating broken links like `https://fury.gl/v0.12.x/index.htmlinstallation.html`. Now that the URLs just end with a slash, the theme builds the link correctly. It even checks if the page actually exists in the older version if it does, it takes you right there if it doesn't, it safely falls back to the main index page. 

Note: if you test this locally, it might still look like it's just sending you to the index page every time. That's just because testing from `localhost` causes a CORS error when trying to ping the live `fury.gl` site. Once this gets merged and is running on the actual website, that CORS error goes away.

ganimtron-10:
Hi @medha-14 ,

I verified the version switching with redirection to the same page on which user currently is and it works fine.

Thanks, LGTM!

maharshi-gor (changes requested):
Hey @medha-14 Thanks for the work.

Please add these lines in make.bat for windows to have the serve functionality.

```
if "%1" == "serve" (
	:serve
	echo Serving docs at http://localhost:8000 ...
	python -m webbrowser http://localhost:8000
	python -m http.server 8000 --directory "%BUILDDIR%\html"
	exit /B
)
```

Please change the color of the selected version to white in the version switcher for light theme if possible as it does not show nicely.

medha-14:
@maharshi-gor Done!

> Please change the color of the selected version to white in the version switcher for the light theme, if possible, as it doesn't display well currently.

Done. I had left this out earlier because I planned to address it during the UI revamp.

Updated: 
image

PR #1236, opened 2026-05-25
Tab UI Introduced

maharshi-gor:
This PR aims to bring tab UI.

It also brings modes for tab ui like top, bottom and accordion.

image

### DIFF
--- a/docs/examples/viz_tab_ui.py
+++ b/docs/examples/viz_tab_ui.py
@@ -1,0 +1,281 @@
+"""
+=====
+TabUI
+=====
+
+This example shows how to use :class:`fury.ui.TabUI` to organize multiple
+2D UI panels inside the same window.
+
+Left click on a tab header to show its content. Left click on the active tab
+again to hide that tab's content. Right click on a tab header to collapse the
+tab UI. The whole widget can be dragged from the tab header or panel
+background.
+"""
+
+##############################################################################
+# First, a bunch of imports.
+
+from fury.ui import Disk2D, LineSlider2D, Rectangle2D, TabUI, TextBlock2D
+from fury.window import Scene, ShowManager
+
+
+##############################################################################
+# Creating a Scene.
+
+scene = Scene()
+
+
+##############################################################################
+# Create a TabUI with three tabs. The first tab is visible on startup.
+
+tab_ui = TabUI(
+    position=(80, 80),
+    size=(420, 320),
+    nb_tabs=3,
+    startup_tab_id=0,
+    draggable=True,
+    active_color=(0.95, 0.95, 1.0),
+    inactive_color=(0.45, 0.48, 0.56),
+)
+
+tab_ui.tabs[0].title = "Shapes"
+tab_ui.tabs[1].title = "Slider"
+tab_ui.tabs[2].title = "Notes"
+
+for tab in tab_ui.tabs:
+    tab.title_font_size = 16
+
+
+##############################################################################
+# Add content to the first tab.
+
+rect = Rectangle2D(size=(90, 70), color=(0.9, 0.25, 0.35))
+disk = Disk2D(outer_radius=45, color=(0.2, 0.75, 0.95))
+caption = TextBlock2D(
+    text="The first tab contains simple 2D shapes.",
+    size=(340, 40),
+    color=(1, 1, 1),
+    font_size=16,
+)
+
+tab_ui.add_element(0, rect, (45, 75))
+tab_ui.add_element(0, disk, (285, 110), anchor="center")
+tab_ui.add_element(0, caption, (40, 205))
+
+
+##############################################################################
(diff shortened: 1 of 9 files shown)


ganimtron-10:
Hi @maharshi-gor,

This looks great. Can we also have vertical tabs something similar used in mordern browsers?

Thanks!

docs/examples/viz_tab_ui.py
```diff
@@ -0,0 +1,346 @@
...
+    nb_tabs=3,
+    startup_tab_id=0,
+    draggable=True,
+    active_color=(0.95, 0.95, 1.0),
+    inactive_color=(0.45, 0.48, 0.56),
+)
+
+tab_ui.tabs[0].title = "Shapes"
+tab_ui.tabs[1].title = "Slider"
+tab_ui.tabs[2].title = "Notes"
```
ganimtron-10: I didnt specifically like this way of index based title setting, shouldnt we have this as a list(title names) or dictionary(title : all elements in that tab with metadata pos, anchor)?

Also this is currenly two step process like set nb_tabs while creating and then set the title names, shouldnt we merge these in initialization itself?

docs/examples/viz_tab_ui.py
```diff
@@ -0,0 +1,346 @@
...
+
+for tab in tab_ui.tabs:
+    tab.title_font_size = 16
```
ganimtron-10: should be a initialization param

fury/ui/containers.py
```diff
@@ -489,6 +489,768 @@ def border_width(self, side_width):
...
+    def _setup(self):
+        """Set up the parent panel and tab panels."""
+        self.parent_panel = Panel2D(size=self.parent_size, opacity=0.0)
+        self._children.append(self.parent_panel)
+        self.on_change = lambda ui: None
+        self.on_collapse = lambda ui: None
+
+        self._update_sizes()
+        for _ in range(self.nb_tabs):
+            content_panel = Panel2D(size=self.content_size)
```
ganimtron-10: why create a content_panel here and pass to the TabPanel2D when it can create its own?

I see content_panel visibility is being set here, but can't it be done inside TabPanel with init parameter specifying it?

fury/ui/containers.py
```diff
@@ -489,6 +489,768 @@ def border_width(self, side_width):
...
+        Returns
+        -------
+        list
+            Actors from the transparent parent panel and all tab panels.
+        """
+        actors = self.parent_panel.actors
+        for tab_panel in self.tabs:
+            actors += tab_panel.actors
+        return actors
+
```
ganimtron-10: as we are adding the parent_panel as children we dont need to get those actors again as chlidrens are added to the scene by Parent

fury/ui/containers.py
```diff
@@ -489,6 +489,768 @@ def border_width(self, side_width):
...
+                    self.left_button_pressed(event, tab_idx)
+                )
+                element.on_left_mouse_button_dragged = self.left_button_dragged
+                element.on_left_mouse_button_released = lambda event: None
+
+            for element in [tab_panel.content_panel.background]:
+                element.on_left_mouse_button_pressed = self.left_button_pressed
+                element.on_left_mouse_button_dragged = self.left_button_dragged
+                element.on_left_mouse_button_released = lambda event: None
+        else:
```
ganimtron-10, line 1088: If you click on any text either on the title or in the content_panel and start dragging it doesn't works, we might need to fix it in the TextBlock itself.

ganimtron-10 (changes requested):
Hi @maharshi-gor ,

PTAL at my below comments, rest LGTM.

Thanks!

docs/examples/viz_tab_ui.py
maharshi-gor: The only problem. I do not want the dict (title: all elements) that is too much and title uniqueness comes into question. I will keep list (title names)

fury/ui/containers.py
maharshi-gor: Technically it should propagate the click to panel as well.

fury/ui/containers.py
maharshi-gor: I understand what you are saying but then _get_actors does not make any sense. If not we can remove the actors property to rely on children attribute rather than get_actors. May be in next PR.

maharshi-gor:
Hey @ganimtron-10 Thank you for the review, I have some concerns about the attributes. Let's take it up in the next FURY meeting.

But this PR should be good to go.

skoudoro (approved):
Hi @maharshi-gor , 

Overall, looks good.

It seems there are multiple tiny bugs. But this PR is big and the overall functionallity is working. We will have to fix them via small PRs.

Can you just fix the conflict.

skoudoro:
Thanks @maharshi-gor, merging

PR #1238, opened 2026-05-29
Add subsections to sphinx gallery

medha-14:


ganimtron-10:
Hi @medha-14, Did you got a chance to look into the subsection and specific file ordering part? If there isnt something specific, lets continue with the toml file as discussed in our meet and close this PR.
Thanks!

medha-14:
> Hi @medha-14, Did you got a chance to look into the subsection and specific file ordering part? If there isnt something specific, lets continue with the toml file as discussed in our meet and close this PR. Thanks!

I checked and the current TOML does not arrange examples in a specific order within a subsection, they are  still being displayed alphabetically rather than in the order in which they are written in the TOML. There is also still the issue of an example belonging to multiple subsections, which cannot be solved by this approach, as we would have to place the example in multiple subfolders, such as `ui` and `introductory`.
So we can close this PR.

PR #1228, opened 2026-05-19
Feat: Adding Demos and Examples

ganimtron-10:
https://github.com/user-attachments/assets/9ecb42ec-4651-458a-a291-371d1c47cedc

feat: adding brownian motion
feat: adding animated surface
feat: adding emwaves
feat: adding fractals
feat: adding helical motion
feat: adding tesseract demo
fix: remove comments, fix position and docs

PR #1243, opened 2026-06-03
Add automated GIF recording support for animation examples

medha-14:
This PR is based upon #1228.

[fury_gif_examples.webm](https://github.com/user-attachments/assets/53887b2c-fa9b-4f22-a053-89478b1ef202)

[gif_example_2.webm](https://github.com/user-attachments/assets/c15080b5-9019-479f-810e-cd3227cdf176)

### DIFF
--- a/docs/source/ext/scrap.py
+++ b/docs/source/ext/scrap.py
@@ -19,7 +19,7 @@ class ImageFileScraper:
         # Iterate through files, copy them to the SG output directory
         image_names = []
         image_path_iterator = block_vars["image_path_iterator"]
-        for path_orig in image_files:
+        for path_orig in sorted(image_files):
             # If we already know about this image and it hasn't been modified
             # since starting, then skip it
             mod_time = os.stat(path_orig).st_mtime
@@ -32,6 +32,10 @@ class ImageFileScraper:
                 continue
             # Else, we assume the image has just been modified and is displayed
             path_new = next(image_path_iterator)
+            orig_ext = os.path.splitext(path_orig)[1].lower()
+            new_ext = os.path.splitext(path_new)[1].lower()
+            if orig_ext != new_ext:
+                path_new = os.path.splitext(path_new)[0] + orig_ext
             self.embedded_images[path_orig] = mod_time
             image_names.append(path_new)
             shutil.copyfile(path_orig, path_new)
--- a/fury/window.py
+++ b/fury/window.py
@@ -1460,8 +1460,38 @@ class ShowManager:
             "true",
             "1",
         ]:
-            self._draw_function()
-            self.snapshot(f"{self._title}.png")
+            frames = []
+            if self._callbacks:
+                max_frames = 300
+                env_max_frames = os.environ.get("FURY_OFFSCREEN_MAX_FRAMES")
+                if env_max_frames and env_max_frames.isdigit():
+                    max_frames = int(env_max_frames)
+
+                for _ in range(max_frames):
+                    if not self._callbacks:
+                        break
+
+                    for _name, cb_info in list(self._callbacks.items()):
+                        func, cb_time, repeat, args = cb_info
+                        func(*args)
+
+                    self._draw_function()
+                    arr = np.asarray(self.renderer.snapshot())
+                    frames.append(image_from_array(arr))
+
+            if frames:
+                gif_name = f"{self._title.replace(' ', '_')}.gif"
+                frames[0].save(
+                    gif_name,
+                    append_images=frames[1:],
+                    save_all=True,
+                    duration=30,
+                    loop=0,
+                )
+            else:
+                self._draw_function()
+                self.snapshot(f"{self._title}.png")
+
             self.window.close()
             return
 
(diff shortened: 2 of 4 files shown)


docs/source/ext/scrap.py
```diff
@@ -32,6 +32,10 @@ def __call__(self, block, block_vars, gallery_conf):
                 continue
             # Else, we assume the image has just been modified and is displayed
             path_new = next(image_path_iterator)
+            orig_ext = os.path.splitext(path_orig)[1].lower()
+            new_ext = os.path.splitext(path_new)[1].lower()
+            if orig_ext != new_ext:
+                path_new = os.path.splitext(path_new)[0] + orig_ext
```
ganimtron-10, line 38: Why do we need this? This would be constant as gif right?

fury/window.py
```diff
@@ -1460,8 +1460,38 @@ def start(self):
...
+                    gif_name,
+                    append_images=frames[1:],
+                    save_all=True,
+                    duration=30,
+                    loop=0,
+                )
+            else:
+                self._draw_function()
+                self.snapshot(f"{self._title}.png")
+
```
ganimtron-10, line 1498: Wouldnt this run even if we are just testing offscreen locally? Isnt there any way to only trigger it if we need it for docs ?

ganimtron-10 (commented):
Hi @medha-14 ,

Had few quick comments, can you take a look and revert?

Thanks!

docs/source/ext/scrap.py
medha-14: The `image_path_iterator` is provided by Sphinx-Gallery and generates output paths with a fixed default extension (typically `.png`). However, `_find_images()` discovers files with multiple extensions, specifically `.png`, `.jpg`, `.jpeg`, and `.gif`.

Without this change, if an example generated an animated `.gif`, the iterator would produce a path such as `fury_brownian_motion.png`, and `shutil.copyfile` would copy GIF binary content into a `.png` file.

I also looked into whether this could be handled through the Sphinx-Gallery configuration instead, but it uses a simple converter that defaults to a single extension and cannot adapt on a per-file basis. Even if we configured it to use `.gif`, examples that should generate `.png` files would break.

fury/window.py
medha-14: I can add a `FURY_BUILD_DOCS` environment variable check and set it in `docs/source/conf.py`. This way, Sphinx-Gallery will generate `GIFs` when building the documentation, while local offsceen testing with `FURY_OFFSCREEN=1` will continue to generate a single static `.png`, like it was doing before.

ganimtron-10: @maharshi-gor your thoughts?

docs/source/ext/scrap.py
ganimtron-10: Thanks for the explanation.

fury/window.py
maharshi-gor: I do believe there should be a variable, but I am thinking the name of the variable can be more descriptive than `FURY_BUILD_DOCS` reason being we might want to store gif apart from generating the docs.

So, `FURY_OFFSCREEN_ANIMATION`? may be

ganimtron-10: @skoudoro suggest a good env name for this please :)

fury/shader.py
```diff
@@ -73,7 +73,7 @@ def get_pipeline_info(self, _wobject, _shared):
         """
         return {}
 
-    def get_bindings(self, wobject, _shared, _scene):
+    def get_bindings(self, wobject, _shared, _scene=None):
```
skoudoro: `def get_bindings(self, wobject, _shared, *,_scene=None):`

to check

medha-14:
I have named the env variable `FURY_RECORD_ANIMATION` for now, though its open to changing it if it's not the best fit. I've also added tests for this. PTAL and let me know if there's anything else that needs to be done here.

maharshi-gor:
Hey @medha-14 can you investigate on fury/shader.py if you can put * before the keyword argument _scene to make keyword only?

maharshi-gor:
Also, let's wait for the base PR from @ganimtron-10 to be clear and merged than we can merge it.

medha-14:
> `def get_bindings(self, wobject, _shared, *,_scene=None):`
> 
> to check

All 5 `shader` classes inherit from `pygfx's BaseShader`, which defines get_bindings with only 2 parameters ie no _scene, and the calls to `get_bindings through get_bindings_info()` in `base.py`. So _scene is never passed and it always defaults to None. I found no direct calls to .get_bindings(), nobody passes it positionally.

The test `test__StreamtubeBakingShader_setup_and_bindings` is currently  failing. It calls shader.get_bindings_info(wobject, {}, None) with 3 args, when it only accepts 2, we can remove the None and  fix this.

maharshi-gor:
Hey @medha-14 please rebase this PR with @ganimtron-10. Or you can just cherry-pick your commit and put in a different branch and we can merge that.

medha-14:
> Hey @medha-14 please rebase this PR with @ganimtron-10. Or you can just cherry-pick your commit and put in a different branch and we can merge that.

Done, PTAL also I have removed the _scene=None changes it seems to be working fine with the newer versions of pygfx .

maharshi-gor:
Thank you for the work @medha-14 .

Works as expected!

PR #1244, opened 2026-06-04
Enables strict documentation builds with warnings treated as errors

medha-14:


skoudoro (approved):
All good, merging, thanks @medha-14.

Let fix now the warnings in a new PR

PR #1246, opened 2026-06-05
RF: Initial Animation Port from vtk to pygfx.

maharshi-gor:
This PR aims to port Animation from vtk to pygfx.

How should you review this PR?

- Animation creation.
- Key frames, Interpolator based actor state updates.
- Transform correctness.

NOTE: This is 1/3 PR of Animation PRs that I have planned.

### DIFF
--- a/fury/animation/animation.py
+++ b/fury/animation/animation.py
@@ -7,17 +7,12 @@ from warnings import warn
 import numpy as np
 from scipy.spatial import transform
 
-from fury import utils
-
-# from fury.actor import line
-from fury.animation.interpolator import (  # noqa F401
+from fury.animation.helpers import compose_transform_matrix
+from fury.animation.interpolator import (
     linear_interpolator,
     slerp,
-    spline_interpolator,
     step_interpolator,
 )
-
-# from fury.lib import Actor, Camera, Transform
 
 
 class Animation:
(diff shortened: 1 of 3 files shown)


skoudoro (approved):
Hi @maharshi-gor,

All good, thank you for that. 

in a new PR, I propose to rename `from animation.animation import Animation` to `from motion.core import Animation`

Also, AI stuff to remember to check if it is True:

test_animation_update_applies_pygfx_actor_state: 🔴 bug: np.diag(matrix)[:3] is not scale when rotation present. Diagonal of TRS matrix gives scale only with identity rotation

  ---
fury/animation/helpers.py
  
  compose_transform_matrix: TRS composition order wrong. Current produces T @ parent @ S @ R. Correct is parent @ T @ R @ S.
  Fix: build local matrix T @ R @ S first, then parent @ local:
  local = np.identity(4, dtype=np.float32)
  if scale_factors is not None:
      local = local @ scale_mat
  if rotation_quat is not None:
      local = local @ rot_mat
  if position is not None:
      local = trans_mat @ local   # T @ R @ S
  return parent_matrix @ local if parent_matrix is not None else local
  This is only invisible in tests because the test uses pure translations (T commutes, so T_child @ T_parent == T_parent @ 
  T_child for the translation column). Add a test with parent rotation to expose it.

  ---
  fury/animation/animation.py
  
  L1160-ish (update_animation): 🔴 bug: parent_matrix reads only first actor (_actors[0]). If parent has multiple actors at
  different positions, child inherits wrong transform. Use the animation's computed transform directly rather than sniffing
  an actor's matrix. Consider storing _current_transform on Animation and reading that.

  ---
  L916 (add method): 🟡 risk: hasattr(item, "local") too broad. Any object with .local passes — accidental accepts non-actor
  objects silently. Use isinstance(item, pygfx.WorldObject) or the fury actor base class instead.

  L1148-ish (opacity update):
  opacity_val = float(np.asarray(opacity).flatten()[0])
  🟡 risk: silently drops multi-value opacity. If interpolator returns shape (N,) > 1, only first value used with no warning.
  Assert scalar or document this is intended.

PR #1254, opened 2026-06-14
Implement custom homepage layout with hero section and navigation overrides (light mode)

medha-14:
image
image

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -144,8 +144,8 @@ html_theme_options = {
     "navigation_depth": 1,
     "navigation_with_keys": True,
     "navbar_start": ["navbar-logo"],
-    "navbar_center": ["navbar-nav"],
-    "navbar_end": ["theme-switcher", "version-switcher", "navbar-icon-links"],
+    "navbar_center": ["custom-navbar-nav.html"],
+    "navbar_end": ["navbar-icon-links", "theme-switcher"],
     "footer_start": ["copyright"],
     "footer_center": "",
     "footer_end": "",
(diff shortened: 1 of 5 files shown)


medha-14:
[Screencast from 2026-06-14 23-00-31.webm](https://github.com/user-attachments/assets/e6f63e3e-d6bf-4baa-a401-91d991bd44b5)

Can we add this opening animation? It might not be the most suitable, so please let me know which one you'd like me to add to the home page.

I also kept the search bar since most other scientific Python packages have it on their home page, such as PyBaMM and scikit-learn. Should I keep it or remove it?

docs/source/conf.py
```diff
@@ -144,23 +144,32 @@
     "navigation_depth": 1,
     "navigation_with_keys": True,
     "navbar_start": ["navbar-logo"],
-    "navbar_center": ["navbar-nav"],
-    "navbar_end": ["theme-switcher", "version-switcher", "navbar-icon-links"],
+    "navbar_center": ["custom-navbar-nav.html"],
+    "navbar_end": ["navbar-icon-links", "theme-switcher"],
```
maharshi-gor, line 148: Just curious why remove version switcher?

You can disable this on the home page not on the entire website.

maharshi-gor (commented):
Hey @medha-14 Thanks for the work.

Please find my comment below. Rest looks good to merge.

docs/source/conf.py
medha-14: >  Just curious why remove version switcher?
> You can disable this on the home page not on the entire website.

Yes I am planning to bring it back (and hide it for only for homepage) in separate PR.

maharshi-gor: Thanks for the confirmation.

PR #1259, opened 2026-06-15
Feat: Adding UI Event Recorder

### LINKED ISSUE #1122: Re-implement UI Event Recorder and Counter
We are currently missing a way to record and replay UI interactions for testing. In FURY v1, we had an **Event Recorder** and **Event Counter** that allowed us to capture user actions and replay them to make sure the UI behaves correctly. Without this, testing complex interactions is a manual headache.

**Describe the solution you'd like**
We need to bring back the record/play functionality into v2. Which should be able to record, store and replay the event from the stored data. Refer v1 implementation.

ganimtron-10:
Adding UI Event Recorder to record, replay and simulate UI interaction tests 

Fixes #1122


ganimtron-10:
This works fine in Show Simulation Mode with `glfw` window but it throws the error as in the CI for `offscreen` and it behaves weirdly. I initially thought its due to Panel2D but ListBox2D works and only these four cause the issue. 

image


I am still figuring out why specifically, @maharshi-gor do you have any idea what might cause this?
Also if you want to make release with this change, let me know will remove these specific testcase from this PR and create a new PR with these testcase and we can keep figuring that part out.

maharshi-gor:
I have no idea. But let me try locally and I will get back to you @ganimtron-10

maharshi-gor:
Hello @ganimtron-10 

I have an idea why it is causing and I fixed it on my local I can push the commit if you want.

There was a small bug that we were dispatching the event before the step_count getting updated in testing.py.

Now, we had everything correct just few frames back. So simulation works nicely but the validation where exact pixel matters became problematic.

Let me know if that make sense.

ganimtron-10:
Sure, Please push the change.

It didn't completely make sense the last part but might make after seeing your change.

maharshi-gor:
Hey @ganimtron-10 I have pushed the PR please check

maharshi-gor:
Hey @ganimtron-10 these are taking too long and operations are cancelled can we do something about them

ganimtron-10:
Hi @maharshi-gor, that what I am looking into, Ideally it shouldn't take this long.

PR #1266, opened 2026-06-18
Implement and integrate custom documentation footer

medha-14:
To be merged after: #1265 

image

image

PR #1267, opened 2026-06-18
refactor: enhance homepage responsiveness and refine mobile UI components 

medha-14:
To be merged after: #1266 

- fixed right top toggle 
image


medha-14:
This PR might need rebasing after merging https://github.com/fury-gl/fury/pull/1266 due to messy rebase history, please let me know if that is the case. 

Otherwise this PR can also be squash merged to keep things simple since there are only two meaningful commits.

PR #1272, opened 2026-06-18
NF: Record functionality properly added.

maharshi-gor:


fury/window.py
```diff
@@ -1143,9 +1167,101 @@ def remove_animation(self, animation):
...
+        if duration > 0:
+            timestamps = np.arange(0, duration, speed / fps).tolist()
+
+        frames = []
+        writer = None
+        try:
+            for timestamp in timestamps:
+                if isinstance(animation, Timeline):
+                    animation.seek(timestamp)
+                    animation.update(force=True)
```
medha-14: seek() already does an update, so having this here might be redundant?

fury/window.py
```diff
@@ -1113,6 +1136,7 @@ def add_animation(self, animation, *, update_rate=1 / 60):
             return
 
         self._animations.append(animation)
+        animation._record_callback = self.record_animation
```
medha-14: `_record_callback` is only being set on the top-level animation/timeline. It is not being propagated to child animations. it will not work if we only want to record a  specific child animation.

fury/window.py
```diff
@@ -1143,9 +1167,101 @@ def remove_animation(self, animation):
...
+        writer = None
+        try:
+            for timestamp in timestamps:
+                if isinstance(animation, Timeline):
+                    animation.seek(timestamp)
+                    animation.update(force=True)
+                else:
+                    animation.update_animation(time=timestamp)
+                render_screens(show_m.renderer, show_m.screens)
+                frames.append(np.asarray(show_m.renderer.snapshot()))
```
medha-14: It looks like all frames are being stored in memory before the video is written. Won't this cause  an out of memory error for longer animations?

medha-14 (changes requested):

skoudoro (approved):
LGTM, merging

skoudoro:
Thank you for reviewing @medha-14

PR #1276, opened 2026-06-22
Extracts shared navbar/footer styles into fury_theme.css and scope homepage.css to home.html

medha-14:
image

This PR solves the problem of unintentional css effect to non homepages.


maharshi-gor:
LGTM Merging!

PR #1278, opened 2026-06-22
Implements global dark mode theme for internal pages (same as homepage)

medha-14:
To be merged after #1276 

image
image
I have also made the red gradient to be visible even when its scrolled (in dark mode).


maharshi-gor:
LGTM Merging!

PR #1279, opened 2026-06-22
Week 4: UI Components and Website Homepage Revamp

medha-14:


medha-14:
Some checks are failing that are unrelated to this PR.

maharshi-gor:
CIs are unrealted. I have fixed in the record animation PR should be gtg after that.

PR #1281, opened 2026-06-24
Implements global light mode theme for internal pages (same as homepage)

medha-14:
image


maharshi-gor:
Thank you @medha-14 LGTM, merging!

PR #1283, opened 2026-06-24
Introduces `Tree UI`

medha-14:
image

image


medha-14:
I have made the changes PTAL, and let me know if anything else needs fixing. Thanks!

maharshi-gor:
Hello @medha-14

I have not deep dived in the PR but I see some issue from running the tutorials.

1. The `viz_tree_actor_modifier.py` that it does not show the actors.
2. The cylinder and cone option does not even open up.
3. There is no way to scroll if the list opens up bit more.
4. The square on the left hand side are misleading as they feel like check box. If want to keep them set something different when they are active.

Once these are done I can take a deep look in the code.

ganimtron-10:
@medha-14 let us know once this is ready for review

medha-14:
I have added all the things that you pointed out earlier, PTAL. Thanks!

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+
+        Returns
+        -------
+        dict
+            Dict with label, node as key, value.
+        """
+        return self._nodes_dict
+
+
+class TreeNode2D(UI):
```
ganimtron-10: Move this above to be before `Tree2D` as it depends on this

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+        selected_color=(0.8, 0.3, 0.3),
+        auto_resize=True,
+        multiselect=True,
+        z_order=0,
+    ):
+        """
+        Initialize the UI element.
+        """
+        self._initial_children = children or []
+        self.node_icons = {
```
ganimtron-10: Whats the significance of having this? How does it help user? How is it different from the arrow button on the right hand side which helps collapse or expand?

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+        parent=None,
+        children=None,
+        icon=None,
+        selected_icon=None,
+        position=(0, 0),
+        size=(200, 200),
+        indent=20,
+        child_indent=10,
+        child_height=25,
+        color=(0.3, 0.3, 0.3),
```
ganimtron-10: This default value collides with the collapse/expand button and its not visible until and unless we activate the `TreeNode`

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+        self.label_image = ImageContainer2D(
+            img_path=self.icon, size=(self.child_height, self.child_height)
+        )
+
+        self.button_icons = {
+            "expand": read_viz_icons(fname="circle-up.png"),
+            "collapse": read_viz_icons(fname="circle-down.png"),
+        }
+
+        self.button = ImageContainer2D(
```
ganimtron-10: Why are we using `ImageContainer2D` instead of `TexturedButton2D`?

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+            img_path=self.button_icons["collapse"],
```
ganimtron-10: Lets make this button's position cutomizable ie. it can be position on the start or end

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+            color=self.scroll_bar_inactive_color,
+            opacity=1.0,
+        )
+        self.scroll_bar.z_order = self.z_order + 1
+        self._children.append(self.scroll_bar)
+
+        self.scroll_bar.on_left_mouse_button_pressed = self.scroll_click_callback
+        self.scroll_bar.on_left_mouse_button_released = self.scroll_release_callback
+        self.scroll_bar.on_left_mouse_button_dragged = self.scroll_drag_callback
+
```
ganimtron-10: Currently only scrollbar is scrolling the list, we need to also allow the `TreeNodes` or the `TreeBackgroundPanel` to also listen to wheel event and do the scroll ie. something similar to what we have in `ListBox` we can scroll using the scrollbar and also using the wheel on the `ListBoxItems` or the `ListBox` itself.

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+        if actual_height > visible_height:
+            fraction = self.view_offset / (actual_height - visible_height)
+            handle_y = int(
+                track_y + fraction * (visible_height - self.scroll_bar.size[1])
+            )
+            self.scroll_bar.set_position((scroll_x, handle_y))
+        else:
+            self.scroll_bar.set_position((scroll_x, track_y))
+
+    def update_scrollbar(self):
```
ganimtron-10: Do you think this Scrollbar can be combined and abstracted out?
Its used in `ListBox`, `ComboBox` and now `TreeUI` too. Maybe a common abstraction could help all three. Check the feasibility and if needed please do the needyful

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+            text=self.label, size=(self.content_size[0], self.child_height)
+        )
+        self.label_image = ImageContainer2D(
+            img_path=self.icon, size=(self.child_height, self.child_height)
+        )
+
+        self.button_icons = {
+            "expand": read_viz_icons(fname="circle-up.png"),
+            "collapse": read_viz_icons(fname="circle-down.png"),
+        }
```
ganimtron-10: You can use plus and minus png icon instead if its on the left side ie. the start of the node, that will resemble the filetree or usually tree UI.

Lets do this: default these buttons to be on the start/left of the node and have default icons as the plus and minus that will give this UI the `TreeUI` feeling as the expand/collapse arrow is already present in the `ComboBox`.

fury/ui/elements.py
```diff
@@ -3597,6 +3597,1229 @@ def resize(self, size):
...
+                                    )
+                                ),
+                                int(coords[1] * self.child_height) + 35,
+                            )
+                            self.content_panel.update_element(child, pixel_coords)
+                            break
+
+        self._update_actors_position()
+
+    def toggle_view(self, event):
```
ganimtron-10: Can't this be done with Toggle Mode in our Buttons?

docs/examples/viz_tree_ui.py
```diff
@@ -0,0 +1,101 @@
...
+# We also create a LineSlider for the LineSlider node.
+
+lineslider = LineSlider2D(length=200, orientation="vertical")
+
+###############################################################################
+# Now, we add the UI elements to their respective nodes.
+
+tree.add_content("ListBox", listbox)
+tree.add_content("LineSlider", lineslider, (0.5, 0.5))
+
```
ganimtron-10: You can add all the sliders present

docs/examples/viz_tree_ui.py
```diff
@@ -0,0 +1,101 @@
...
+tree.add_content("Panels", panel_first)
+tree.add_content("Panels", panel_second, (0.5, 0.5))
+tree.add_content("ImageContainers", img, (0.5, 0.5))
+
+###############################################################################
+# Now, let's create UI elements for the Elements node.
+# First, we create a ListBox for the ListBox node.
+
+listbox = ListBox2D(values=["First", "Second", "Third", "Fourth"])
```
ganimtron-10: You can also add `ComboBox`

docs/examples/viz_tree_ui.py
```diff
@@ -0,0 +1,101 @@
...
+###############################################################################
+# Next, we create a Disk2D for the Disk node.
+
+disk = Disk2D(outer_radius=50, color=(0.6, 0.2, 0.8))
+
+###############################################################################
+# Now, we add the UI elements to their respective nodes.
+
+tree.add_content("Rectangle", rect)
+tree.add_content("Disk", disk, (0.5, 0.5))
```
ganimtron-10: You can also add Buttons to this, start with core then elements and then containers and include all UI elements we have so this would become a common point to see all the breakdown of UI present in FURY

ganimtron-10 (approved):
Hi @medha-14 , PTAL at the below comments for the initial review.

docs/examples/viz_tree_ui.py
ganimtron-10: I also noticed some issue with the ListBox, PTAL
image

```
Error during handling pointer_up event
  File "/home/ganimtron_10/miniforge3/envs/furyenv/lib/python3.12/site-packages/rendercanvas/_coreutils.py", line 51, in log_exception
    yield
  File "/home/ganimtron_10/miniforge3/envs/furyenv/lib/python3.12/site-packages/pygfx/objects/_events.py", line 397, in handle_event
    callback(event)
  File "/mnt/d/FuryWorkspace/fury/fury/ui/core.py", line 404, in mouse_button_up_callback
    self.left_button_release_callback(event)
  File "/mnt/d/FuryWorkspace/fury/fury/ui/core.py", line 443, in left_button_release_callback
    self.on_left_mouse_button_clicked(event)
  File "/mnt/d/FuryWorkspace/fury/fury/ui/elements.py", line 3586, in left_button_clicked
    self._on_select(item=self, multiselect=multiselect, range_select=range_select)
  File "/mnt/d/FuryWorkspace/fury/fury/ui/elements.py", line 3399, in select
    selection_idx = self.values.index(item.element)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: None is not in list
```

ganimtron-10 (changes requested):

fury/ui/elements.py
medha-14: It was like this before in the code I based this PR on, though it would be better if it was `TextredButton2D`.

fury/ui/elements.py
medha-14: It can we used to provide an icon for the node, like a small icon of a file if we are using it as a file menu showing what type of file this is.

fury/ui/elements.py
medha-14: I can take this up in a separate PR, as it would require changing multiple components.

medha-14:
I have addressed the reviews on this PTAL. Thanks

maharshi-gor:
Hello @medha-14 ,

I tried the PR again, I am not satisfied with the looks of Tree UI.

The circular ring doesn't make any sense along with the + and - icons. The icons do not have any padding.

The plus icon just looks like a giant connected line. 

Also, the cube is not visible in the tree actor modifier test which is very weird.

medha-14:
I have modified the icons, made the left symbol icons optional, and added padding to the + and - symbols, please take a look, and let me know if anything else needs fixing.

maharshi-gor:
Hey @medha-14 good work!

I just have one small bug around the scroll wheel and size. I think if you open two three sections on `actor_modifier` example it shows a scroll bar which scrolls everything up and down and the elements are not even in overflow state. Please check rest LGTM.

Also please do not forget to edit migration guide and state of Tree UI so that we can update it.

Here is the reference for you to understand.
Screenshot 2026-08-10 at 8 14 45 AM

Thanks!

PR #1286, opened 2026-06-26
BF: Dev doc builds were not generating the examples.

maharshi-gor:


fury/tests/test_window.py
```diff
@@ -17,12 +16,16 @@
...
     PerspectiveCamera,
     PointerEvent,
+    QtWidgets,
     Renderer,
     Scene as GfxScene,
     ScreenCoordsCamera,
     Texture,
     TrackballController,
     have_imgui_bundle,
+    have_py_qt5,
```
skoudoro: why qt5 ? I believe we should avoid that

ganimtron-10:
LGTM apart from Serge's comment.

fury/lib.py
```diff
@@ -50,7 +50,8 @@
 PyQt6, have_py_qt6, _ = optional_package("PyQt6", trip_msg=qt_pckg_msg)
 PyQt5, have_py_qt5, _ = optional_package("PyQt5", trip_msg=qt_pckg_msg)
```
skoudoro: why qt5?

fury/lib.py
```diff
@@ -50,7 +50,8 @@
...
 
-if have_py_side6 or have_py_qt6 or have_py_qt5:
+have_qt = have_py_side6 or have_py_qt6 or have_py_qt5
```
skoudoro: why qt5 ?

fury/lib.py
```diff
@@ -59,14 +60,15 @@ def get_app():
...
 
-if have_py_qt6:
-    from PyQt6 import QtWidgets
+    if have_py_qt6:
+        from PyQt6 import QtGui, QtWidgets
 
-if have_py_qt5:
-    from PyQt5 import QtWidgets
+    if have_py_qt5:
+        from PyQt5 import QtGui, QtWidgets
```
skoudoro: why qt5 ?

skoudoro (changes requested):
Hi @maharshi-gor ,

See some comments below. 

Maybe pyproject to update for QT. To check.

maharshi-gor:
Hey @skoudoro Thank you for the feedback.

We can support pyqt5 and pyqt6. So this is correct.

We recommend pyside6 if you want to use qt. That is why pyside6 is present in `pyproject.toml`

There is no change in ourside of code between pyqt5 and pyqt6. It is just for the user convenience,

skoudoro:
Conflict to fix.

I do not recommend to support pyqt5.  
- bad license : GPL
- last release in 2024.  I do not think it respect our minimal python version. Also, dependency might be a nightmare.

so we should force pyside 6 as minimum and avoid pyqt6

maharshi-gor:
Okay I will remove it and update the PR.

skoudoro (approved):
LGTM, thanks @maharshi-gor ,

merging

PR #1297, opened 2026-07-09
Demo: Adding Flight Simulator

ganimtron-10:
Adding flight sim demo


maharshi-gor:
Hey @ganimtron-10 Thanks for the work.

It looks much better than the v1.

There are few things to improve.

1. The w, a, s, d feels little off. 
2. The mountains can not be on the runway.
3. The runway never ends.
4. The plane flips and then you cannot correct it.

Let's see how we can tackle these.

maharshi-gor:
There are still some improvements but that will be for later. It is good for now.

PR #1303, opened 2026-07-17
UI: Combobox expands on `TextBlock` clicks

ganimtron-10:
Combobox previously needed to be clicked specifically on the arrow icon to expand and collapse which wasnt very convenient as we usually click on the textblock too to expand. So added that accessibility.


maharshi-gor:
LGTM Merging!

PR #1305, opened 2026-07-22
NF: Porting Radio button and check box.

maharshi-gor:
- Incorporated new icons from @ganimtron-10 for selected state of the radio button.
- Test cases included.
- Tutorial for the same is included.


fury/ui/elements.py
```diff
@@ -2255,331 +2252,519 @@ def range_slider_handle_move_callback(self, ui):
...
+        Spacing in pixels between two adjacent options.
+    font_size : int, optional
+        Font size of the labels in pixels.
+    font_family : str, optional
+        Font family of the labels. Currently only supports "Arial".
+    text_color : str, tuple, list or ndarray, optional
+        Color of the label text. Accepts a hex string ("#FF0000"), RGB(A) in
+        [0, 1], or RGB(A) in [0, 255].
+    position : (float, float), optional
+        Absolute coordinates (x, y) of the top-left corner of this component.
```
skoudoro, line 2320: are you sure top-left corner ? look like vtk, isn't it lower-left now ?

fury/ui/elements.py
```diff
@@ -2255,331 +2252,519 @@ def range_slider_handle_move_callback(self, ui):
...
```
skoudoro, line 2648: are you sure top-left corner ? look like vtk, isn't it lower-left now ?

fury/ui/elements.py
```diff
@@ -2255,331 +2252,519 @@ def range_slider_handle_move_callback(self, ui):
...
```
skoudoro, line 2707: are you sure top-left corner ? look like vtk, isn't it lower-left now ?

fury/ui/elements.py
```diff
@@ -2255,331 +2252,519 @@ def range_slider_handle_move_callback(self, ui):
...
+        ----------
+        text : TextBlock2D
+            The text block whose layout should be computed.
+        """
+        update = getattr(text.actor, "_update_object", None)
+        if callable(update):
+            try:
+                update()
+            except Exception:
+                pass
```
skoudoro: this is weird. ignore all exception ? to remove or put a warning, something

fury/ui/elements.py
```diff
@@ -2255,331 +2252,519 @@ def range_slider_handle_move_callback(self, ui):
...
+        orientation="vertical",
+        padding=10,
+        font_size=18,
+        font_family="Arial",
+        text_color=(1, 1, 1),
+        position=(0, 0),
+        z_order=0,
+    ):
+        """Init class instance."""
+        if len(checked_labels) > 1:
```
skoudoro, line 2732: not compatible with generator. so use `len(list(checked_labels))`

skoudoro (changes requested):
Hi @maharshi-gor,

Overall looks good.

It seems there is a bug with radio button since selection should be unique and I succeed to have multiple selected. 

Something to dig. (`ButtonGroup.select() ` to look)

skoudoro:
> It seems there is a bug with radio button since selection should be unique and I succeed to have multiple selected.

Strange, I can not reproduce it....

maharshi-gor:
I confirm the top left is the new convention.


I also could not reproduce double checked radio button.

But this was a good catch as many places write lower left instead of top left while porting.

Working on that.

skoudoro (approved):
it seems to be ok. 

Thank you @maharshi-gor. merging

PR #1306, opened 2026-07-22
BF: Fix phantom sphere in the case of imposter (billboard) Sphere.

maharshi-gor:
- Updated the test case for the same.

Before:
image

After
image


skoudoro (approved):
Thank you for this, merging

PR #1310, opened 2026-07-27
BF: macOS glfw warning fix

maharshi-gor:
On macOS the `glfw` package and `imgui_bundle` each ship their own `libglfw.3.dylib`; loading both triggers "Class GLFW... is implemented in both ..." objc warnings. Point `glfw` at imgui_bundle's copy so a single library is loaded. Must run before `glfw` is first imported.

Also, If imgui_bundle is not present it will not pick and fallback to glfw.

### DIFF
--- a/fury/__init__.py
+++ b/fury/__init__.py
@@ -1,6 +1,8 @@
 """Init file for visualization package."""
 
-from os.path import dirname
+from importlib.util import find_spec
+import os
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Nice ! Thanks @maharshi-gor for this, 

Work as expected, merging

PR #1312, opened 2026-07-27
Revamp FURY website

medha-14:


skoudoro (approved):
all good, Thank you for this work @medha-14 and @maharshi-gor !

merging

PR #1313, opened 2026-07-27
RF: Update the code to set unset env variables without needing to set…

maharshi-gor:
Doc generation does not need variables to set.


skoudoro (approved):
All good, thanks @maharshi-gor!

merging

PR #1314, opened 2026-07-31
Seo optimization cherrypick

maharshi-gor:


skoudoro (approved):
it seems still few issues but will come back to improve SEO letter.

We will have to focus on this and ASEO if we want LLM to recommend us.

merging

PR #1325, opened 2026-08-24
Add per corner radius to rounded rectangle and merge rounded rectangle into rectangle

medha-14:


medha-14:
@maharshi-gor @ganimtron-10 could you please review this. Thanks

PR #1326, opened 2026-08-25
PBR support

maharshi-gor:
This PR introduces by default PBR support to the primitive actors.

Also fixes some other smaller issues in text buttons.
image

### DIFF
--- a/fury/window.py
+++ b/fury/window.py
@@ -31,6 +31,7 @@ from fury.lib import (
     EventType,
     GfxGroup,
     JupyterCanvas,
+    MeshStandardMaterial,
     OffscreenCanvas,
     PerspectiveCamera,
     PointerEvent,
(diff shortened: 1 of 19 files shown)


Garyfallidis:
image

Garyfallidis:
Let's save these options as preset.
Please also add transparency.
Do both on a new PR.

Garyfallidis:
We also need API to change PBR properties in an actor with many objects independently.
For example, one actor that contains two spheres each having different metallicity. 
That will require passing PBR properties as vectors.

fury/window.py
```diff
@@ -215,6 +265,7 @@ def add(self, *objects):
                 super().add(obj)
             else:
                 self.main_scene.add(obj)
+                self._apply_env_map(obj)
```
ganimtron-10, line 268: we do this on add but never do `_clear_auto_env_maps` on removing it. A reference still stays even if actor is removed.

PR #1330, opened 2026-09-03
Implement font_size property in elements.py

ShivrajSS-git:
fixes #1194  Add font_size property with getter and setter to manage font size and update text size accordingly.

PR #1331, opened 2026-09-04
Add test for TextBox2D font size background update

ShivrajSS-git:
Added a regression test to verify that changing the font size of TextBox2D updates the associated text and background size.

The test verifies that:
- TextBox2D font_size is updated correctly.
- The internal TextBlock2D font_size is updated.
- The background width increases when the font size increases.
- The background height increases when the font size increases.

If GitHub asks for a short commit message only, just use:


maharshi-gor:
Closing as per the comment in #1330

PR #1332, opened 2026-09-04
Add support for saving frames as GIF

ShivrajSS-git:
Add tests for saving multiple frames as animated GIFs.

The tests verify GIF creation, frame count, color preservation, and handling of invalid file extensions.

### DIFF
--- a/fury/io.py
+++ b/fury/io.py
@@ -263,6 +263,54 @@ def save_image(
     im.save(filename, **save_kwargs)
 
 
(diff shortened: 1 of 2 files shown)


maharshi-gor:
Hey @ShivrajSS-git Thank you for your contribution. 


this is already done in the show manager. I do not understand the implementation motivation.

If you are replacing the function, please update what we have in the show manager.

Otherwise I do not find any added value of this addition.

PR #1333, opened 2026-09-04
Add support for saving frames as GIF

ShivrajSS-git:
Add a save_as_gif function to save multiple image frames as animated GIF files with configurable duration and looping.


ShivrajSS-git:
Closing this PR because the implementation and tests have been combined into #1332.

PR #1336, opened 2026-09-14
NF: GLTF support

maharshi-gor:


maharshi-gor:
Not required as glTF got implemented on polyxios. Will open a new PR for adaption using that.

Issue #856, opened 2023-12-21
TypeError: vector_text() got an unexpected keyword argument 'direction'


chiuhoward:
## Description
When trying to provide the direction argument in actor.vector_text following https://fury.gl/latest/reference/fury.actor.html#vector-text, I get an unexpected keyword argument.

I am running fury from within a miniconda environment with jupyter notebook. 

Code below (which is the example provided in the documentation) works perfectly fine without the direction argument.

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [ ] Code example
```
import os
import os.path as op
import nibabel as nib
import numpy as np

from dipy.io.streamline import load_trk
from dipy.tracking.streamline import transform_streamlines

from fury import actor, window
from fury.colormap import create_colormap

from PIL import Image

scene = window.Scene()
l = actor.vector_text(text='Hello', direction=(0,1,1))
scene.add(l)
window.show(scene)

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-41aa93e6da7e> in <module>
      1 scene = window.Scene()
----> 2 l = actor.vector_text(text='Hello', direction=(0,1,1))
      3 scene.add(l)
      4 window.show(scene)

TypeError: vector_text() got an unexpected keyword argument 'direction'
```

- [ ] Relevant images (if any)
<img width="670" alt="image" src="https://github.com/fury-gl/fury/assets/137316255/481857b8-1e82-46ab-8109-233c60fe10e1”>

- [ ] Operating system and versions (run `python -c "from fury import get_info; print(get_info())”`)
{'fury_version': '0.8.0', 'pkg_path': '/Users/howardchiu/miniconda3/envs/pyAFQ/lib/python3.9/site-packages/fury', 'commit_hash': 'b937ff3418e7c64b0fa89921a54c083ff78dc055', 'sys_version': '3.9.18 (main, Sep 11 2023, 08:25:10) \n[Clang 14.0.6 ]', 'sys_executable': '/Users/howardchiu/miniconda3/envs/pyAFQ/bin/python', 'sys_platform': 'darwin', 'numpy_version': '1.24.3', 'scipy_version': '1.9.1', 'vtk_version': '9.3.0', 'matplotlib_version': '3.8.2', 'dipy_version': '1.7.0'}

chiuhoward:
Screenshot didn’t get uploaded properly but it’s the same as the code and traceback

[...]

PR #1337, opened 2026-09-17
Fix "vector_text" direction argument support

AaryanRS:
This PR address The Issue #856, link - https://github.com/fury-gl/fury/issues/856

Problem : "vector_text()" got an unexpected keyword argument "direction"

Summary : The project documentation and examples still advertise the "vector_text" API with a "direction" argument, but the current actor implementation no longer exposed this function in the public API. As a result, users following the documented example encounter a runtime error even though the API contract is documented.

What I changed :- 

- Restored support for "actor.vector_text(...)" in the planar actor API.
- Added validation for the `direction` argument so zero-length/invalid vectors are rejected cleanly.
- Implemented orientation logic that rotates the default text direction (+Z axis) to the user-provided direction.
- Preserved the expected behavior for "position" and "scale".
- Exported the function in the actor stub so it is available through the public module API.
- Added regression tests covering:
            -valid direction rotation
            -invalid zero-direction rejection

This fix restores compatibility with the documented behavior without making unrelated changes to the rest of the rendering pipeline

Validation :-

I added assertions to verify:

- the actor accepts a valid non-zero direction vector
- the direction rotates the text correctly
- invalid zero vectors raise a `ValueError`
- the actor retains the correct position and scale configuration

This should make the documentation example work as expected again.

Thank you for reviewing this PR. I would appreciate any feedback on the fix and the test coverage.

Best,
Aaryan Shelar


fury/actor/planar.py
```diff
@@ -715,6 +715,62 @@ def text(
...
+    ------
+    ValueError
+        If ``direction`` is not a non-zero three-dimensional vector.
+    """
+    direction = np.asarray(direction, dtype=np.float32)
+    if direction.shape != (3,) or np.linalg.norm(direction) == 0:
+        raise ValueError("direction must be a non-zero vector of length 3.")
+
+    obj = globals()["text"](text=text, colors=color, position=pos)
+    direction /= np.linalg.norm(direction)
```
maharshi-gor, line 756: I do not find this pattern correct.

Please use existing API to provide better reusing of code.

Thank you!

maharshi-gor:
Hey @AaryanRS 

I appreciate you working on this. But, a great way to demonstrate the changes in FURY PRs is to attach a screenshot of the visuals. As we are a visualization heavy library.

Please do that same for this.

fury/actor/planar.py
AaryanRS: Ohk, I will work on it ASAP...

AaryanRS:
Hello @maharshi-gor 

I am sorry for not adding screenshots of visuals, Currently I will work on this issue and will provide visuals of the output.
thank you

Issue #652, opened 2022-08-05
Kernel dies while using window


Paddy-Maya:
## Using Dipy package
I am working with diffusion MRI data and I have been using Dipy package for my studies. Recently in-order to visualize the tractography of whitematter I used fury, even time I run the following command.

`from dipy.viz import window, actor, has_fury
if has_fury:
    scene = window.Scene()
    scene.add(actor.peak_slicer(csa_peaks.peak_dirs,
                                csa_peaks.peak_values,
                                colors=None))

    window.record(scene, out_path='csa_direction_field.png', size=(900, 900))

    if interactive:
        window.show(scene, size=(800, 800))`

the kernel dies and it says that it will automatically restart, I tried almost all the solutions that I was able to find in the previous issues.
I thought maybe this might be issue with space and also tried with a new GPU 
and I still get this error.

I even tried to run only a small explample that was given in
[https://fury.gl/dev/auto_tutorials/01_introductory/viz_texture.html](url)

and I get the same issue 
<img width="1420" alt="Screenshot 2022-08-05 at 3 44 28 PM" src="https://user-images.githubusercontent.com/85098718/183018015-5af43f8d-8226-4727-bc06-91fe7f64a67d.png">
I tried it with even the pre-release of jupyter notebook 
but nothing works.

It would be great if anyone could help me 

Thanks

AaryanRS:
Hello @WassCodeur ,

I Have faced This similar type of issue Before and i can solve it

@Paddy-Maya The reason the kernel dies and it says that it will automatically restart is because....

Problem: "ShowManager.start()" called the global "rendercanvas" event loop through "run()" for every window type. In Jupyter, this loop conflicts with the notebook kernel and can terminate it. The Jupyter widget was displayed only after "run()" returned, which normally never happens during interactive use.

To Fix this I have Simply Added the following function:

<img width="787" height="442" alt="Image" src="https://github.com/user-attachments/assets/88844856-f308-490f-bd47-821e64ad9bb2" />

<img width="447" height="127" alt="Image" src="https://github.com/user-attachments/assets/3ebc3eff-d60d-46d4-ac28-602fd5a26b05" />

here's how they will fix the issue :

In "fury/window.py", "ShowManager.start()" now:

1. Performs the initial render.
2. Preserves the existing "FURY_OFFSCREEN" behavior.
3. Displays the Jupyter canvas immediately when "window_type="jupyter"".
4. Returns without calling the desktop event loop.

[...]

PR #1338, opened 2026-09-18
Fix: resolved The Issue when Kernel dies while using window from fury, (trackerID : 652)

AaryanRS:
This PR address The Issue https://github.com/fury-gl/fury/issues/652

Problem : Kernel dies while using window from fury

Issue-In : The crash path is in "ShowManager.start": a Jupyter canvas is treated like a desktop canvas and enters the global render loop before the widget is displayed. I’m changing that branch to render/display the widget without starting the desktop event loop, with a regression test that proves "run()" is not called.

FIX : In "fury/window.py", "ShowManager.start()" :

1. Performs the initial render.
2. Preserves the existing "FURY_OFFSCREEN" behavior.
3. Displays the Jupyter canvas immediately when [window_type="jupyter"].
4. Returns without calling the desktop event loop.

I Added a focused regression test in "fury/tests/test_window.py" around (line 643). It verifies that Jupyter startup displays the widget and never calls "run()".

Screenshot 2026-09-18 133118
Screenshot 2026-09-18 133235

Thank you for reviewing this PR. I would appreciate any feedback on the fix and the test coverage.

Best,
Aaryan Shelar
Email - aaryan.r.shelar5418@gmail.com


fury/tests/test_window.py
```diff
@@ -639,6 +639,22 @@ def test_display_accepts_iterable_actors(sample_actor):
         assert second_actor in kwargs["scene"].main_scene.children
```
WassCodeur, line 639: Thank you @AaryanRS  for the contribution.

I really appreciate the work you’ve put into this. I’ve left a few comments below for you to take a look at.

fury/window.py
```diff
@@ -1971,15 +1971,17 @@ def start(self):
             self.window.close()
             return
 
+        if self._window_type == "jupyter":
```
WassCodeur, line 1974: I’m curious to know why you decided to move this logic up here. Was there a specific reason behind this change?

fury/tests/test_window.py
```diff
@@ -639,6 +639,22 @@ def test_display_accepts_iterable_actors(sample_actor):
...
         mock_show_manager.return_value.start.assert_called_once_with()
 
+def test_jupyter_start_does_not_enter_desktop_event_loop():
```
WassCodeur, line 642: Image

I noticed that this test is currently failing on my side. I’ve attached a screenshot of the failure.

Could you also run the test on your side to see if you can reproduce it? You can run it with:

```bash
pip install -r requirements/test.txt
pytest -svv fury
```

Let me know what you get.

WassCodeur (commented):
I really appreciate the work you’ve put into this. I’ve left a few comments below for you to take a look at.

AaryanRS:
Thank you for Pointing Out my Error in Code I will work right into and solve it...

PR #1339, opened 2026-09-21
fixed: per-streamline linewidth, (trackerID : 860)

### LINKED ISSUE #860: In need of per-streamline linewidth
I need to visualize tractography streamlines that have different widths (representing axonal diameters), but I can only provide the `actor.streamtube` function with a single linewidth.

**Describe the solution you'd like**
I'd like for actor functions that accept a linewidth argument (Ex: `actor.streamtube`, `actor.line`) to also accept an arraylike version for per-streamline width.

**Describe alternatives you've considered**
> I have tried looping through each streamline and adding them separately to the scene, with no success.
> ```
> 
> s = window.Scene()
>     for i, line in enumerate(tractogram.streamlines[:5]):
>         line_actor = actor.streamtube(line, colors=[1., 0., 0.], opacity=0.5, linewidth=diameters[i])
>         s.add(line_actor)
> ```
> When using this alternative:
> - `actor.line` gets me a memory error:
>  ```
> malloc(): invalid size (unsorted)
> Abandon (core dumped)
>  ```
> - `actor.streamtube` gets me an empty scene, no matter what other actor I add to the scene. It kills the thing 😅

Edit:
This solution actually works by changing it to `... actor.streamtube([line], ...`. The errors were simply related to that.
It's a suboptimal solution, but my problem is fixed for now. I'll keep the issue open as the feature still seems relevant and rather quick to add.

**Additional context**
An example of what I'm trying to achieve, but with variable streamline widths:

image

AaryanRS:
Fixed #860 

Problem: Before These Changes `actor.line()` and `actor.streamtube()` Were having only one thickness Setting, If I have added five streamlines then all five had to use the same width, That's why we were only able to provide the "actor.streamtube()" function with just a single linewidth.

Solution : I added a feature where, when an array of widths is provided, the code creates a group containing one small actor for each streamline. Each actor receives its own width. This is similar to drawing each road separately with a different-sized pen. 

For `actor.line()`, the value controls the line material’s thickness. 
actor.line(lines, linewidth=[1.0, 2.0, 4.0])

For `actor.streamtube()`, the value controls the tube radius.
actor.streamtube(lines, linewidth=[0.1, 0.2, 0.4])

Thus With these changes we can have lines with different Thickness and Radius. It also avoids the problem where manually calling `actor.line()` or `actor.streamtube()` with one raw streamline could cause crashes or an empty scene.

I also made Sure that our old feature runs normaly If we provide only one number for all lines( thickness and Radius )

Also Added Tests, these tests create two streamlines with different widths and check that two child actors are created and that their widths are actually different. Therefore this PR is mainly my new feature added to fury and not a repair to the original scalar-width behavior.

Thank YOU for reading This PR....
Aaryan Shelar


skoudoro:
@maharshi-gor, can you have a look at this ?

Thanks @AaryanRS for looking into this

maharshi-gor:
Hello @AaryanRS 

Thank for the work. But, I do not see any added value of this PR. If you notice that the code changes entire implementation without reusing any of the existing code.

on top of that we have streamlines API which does accept the thickness.

I am closing this PR, please feel free to open it if you still find it useful.

